PR #14, opened 2020-06-25
Improve testing pipeline

pierDipi:
## Proposed Changes

- Skip tests (unit and checkstyle) when building components images
- Remove duplicated Dockerfile and use only one with build arguments


slinkydeveloper:
Ok so this dockerfile is just meant for release and e2e testing right? Now we need another one where we run unit tests in java (using `mvn verify`)

pierDipi:
We have it here: https://github.com/knative-sandbox/eventing-kafka-broker/blob/master/data-plane/docker/test/Dockerfile

pierDipi:
> Ok so this dockerfile is just meant for release and e2e testing right?

Correct.

PR #20, opened 2020-07-01
Add maven wrapper

### LINKED ISSUE #19: Use Maven wrapper
**Problem**
The Maven Wrapper is an excellent choice for projects that need a specific version of Maven (or for users that don't want to install Maven at all). Instead of installing many versions of it in the operating system, we can just use the project-specific wrapper script.

Contributor

**Exit Criteria**
- Use `./mvnw` for running tests.

**Time Estimate (optional):** 
1

- wrapper: https://github.com/takari/maven-wrapper

/good-first-issue

pierDipi:

Fixes #19 

## Proposed Changes

- Add maven wrapper - maven 3.6.3 (`mvn -N io.takari:maven:0.7.7:wrapper`)
- Update development instructions (`DEVELOPMENT.md`)
- Use Java only base images for building and testing


data-plane/docker/Dockerfile
```diff
@@ -27,17 +27,20 @@ COPY dispatcher/pom.xml dispatcher/pom.xml
 COPY generated/pom.xml generated/pom.xml
 COPY tests/pom.xml tests/pom.xml
 COPY checkstyle/suppression.xml checkstyle/suppression.xml
+COPY ./mvnw .
```
slinkydeveloper, line 30: I think it's ok to keep using the maven from the docker image

pierDipi: I think with this change we have a more homogeneous environment because before we ran tests (unit and build) with a maven-based image (I don't know which `JDK` build uses underneath), but for e2e tests, we used `adoptopenjdk`-based image. Now, we use only one image across all tests, and we can maintain only one Java version (in one place), besides regularly updating the wrapper that we should do anyway regardless of how we run tests. 

Is it ok for you to keep this change? Am I missing something?

slinkydeveloper: ok :wink:

PR #29, opened 2020-07-03
Update docs to include installation from scratch

### LINKED ISSUE #12: Retry inside the kafka_setup script
In a followup, Could you have a retry inside the script more than asking to the user to do so?

_Originally posted by @slinkydeveloper in https://github.com/knative-sandbox/eventing-kafka-broker/pull/11_

/good-first-issue

### LINKED ISSUE #26: update developer guide to include tailored k8s cluster installation
**Problem**

The current developer guide documentation references the main `knative install docs` for instructions on how to create a  k8s cluster for this project. However, the referenced knative docs does not include the creation of k8s clusters and assumes that user's already have one created.

Contributor

pierDipi:
Fixes #26 
Fixes #12 

## Proposed Changes

- Add documentation to install everything needed from a clean K8s cluster.
- Wait for Strimzi to register resources on installation.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -76,7 +76,7 @@ follows.
 
 - This guide assumes you have a Kubernetes cluster up and running. - https://kubernetes.io/docs/setup/
 
-1. Execute `source test/e2e-common.sh `
+1. Execute `source test/e2e-common.sh`
 1. Execute `knative_setup`
     - This command deploys Knative Eventing.
 1. Execute `kubectl wait -n knative-eventing pods --all=true --for=condition=Ready`
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -24,18 +24,12 @@ You need to install:
 
 Requirements signaled as "optional" are not required, but it's highly recommended having them installed. If a specific version of a requirement is not explcitly defined above, any version will work during development.
 
-### Create a cluster and a repo
+### Set up a Linux Container Registry
 
-1. [Set up a kubernetes cluster](https://www.knative.dev/docs/install/)
-   - Follow an installation guide up through "Creating a Kubernetes Cluster"
-   - You do _not_ need to install Istio or Knative using the instructions in the
-     guide. Simply create the cluster and come back here.
-   - If you _did_ install Istio/Knative following those instructions, that's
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -78,15 +72,17 @@ _Adding the `upstream` remote sets you up nicely for regularly
...
 
-```bash
-# re-execute the script in case some errors appear. 
-# (this can happen when the CRDs hasn't been registered and we try to create a Kafka cluster)
-./test/kafka/kafka_setup.sh 
+- This guide assumes you have a Kubernetes cluster up and running. - https://kubernetes.io/docs/setup/
 
-kubectl apply -f config
-```
+1. Execute `source test/e2e-common.sh `
```
slinkydeveloper, line 79: there is a space before the end of the code block

pierDipi: Fixed.

What do you think about re-using test script, for development set up? I'm using it daily and it works really well; 
it removes the need for every developer to create its own 5min script.

DEVELOPMENT.md
```diff
@@ -78,15 +72,17 @@ _Adding the `upstream` remote sets you up nicely for regularly
...
+1. Execute `knative_setup`
+    - This command deploys Knative Eventing.
```
Alfusainey, line 81: @pierDipi under the `Checking out your fork` section,  you already mention that the knative eventing repo should be cloned. so at this step, maybe it is either right to assume that the repository exist locally or add a check so that you only clone if there is no existing repo. i get a fatal error indicating that the repo already exist, but the script  continues to execute. WDYT?

pierDipi: I think this is "fine", given that it doesn't break correctness. 
I didn't add a check for existence to keep the script simpler. 
Perhaps we can mention that a fatal error can occur in case you have already eventing cloned.

Alfusainey: sure, i think mentioning that as `NOTE` in the docs should help.

also, are there any minimum requirements for the Kubernetes cluster e.g `cpus`, `memory`, or a specific `vm-driver` to use so that the installation works seamlessly? or any cluster config will work?

pierDipi: We don't have any requirements yet.

PR #37, opened 2020-07-10
Receiver responds with 404 when it receives a request to a non-existing Broker.

### LINKED ISSUE #33: [Data-plane] Receiver should reply with 404 in case it receives a request to a non existing broker
**Problem**
Now the receiver just tries to push the events to the topic, without checking if that broker exists.

**Exit Criteria**
An integration test that replies with 404 when you try to push a valid event to a non existing broker URL

**Time Estimate (optional):** 
2

The problem with receiver is that today it doesn't have knowledge of the existing brokers, aka it doesn't watch the famous data plane config map.

While in dispatcher side you solved the file watcher thread -> vertx communication deploying and undeploying verticles, here I have some ideas that might give an implementation hint:

0. Without revolutionize the code, since you just need a read view and you recreate every time that map, an `AtomicReference` shared between the watcher thread and the http verticle should suffice.
1. Without revolutionize the code, you might have a shared map using https://vertx.io/docs/vertx-core/java/#_using_the_shareddata_api. On file watcher thread, you write in that map and on the http verticle (request handler) you read that map
2. A more vert.x-ish approach could be to use the event bus to perform updates: on http verticle receiver you start an event bus receiver that, every time receives a message at address `updatebrokers.myapp`, updates its internal set of brokers. On file watcher side, every time you receive an update from the file, you broadcast the new set of brokers to `updatebrokers.myapp`.

Choose the simplest approach that fits better both for this issue and eventual next changes to the receiver code that requires knowledge of the available brokers

pierDipi:
Fixes #33 

## Proposed Changes

- The receiver responds with 404 when it receives a request to a non-existing Broker.


data-plane/receiver/src/test/java/dev/knative/eventing/kafka/broker/receiver/EnvTest.java
```diff
@@ -26,30 +26,26 @@
...
-            case Env.LIVENESS_PROBE_PATH:
-              return LIVENESS_PATH;
-            case Env.READINESS_PROBE_PATH:
-              return READINESS_PATH;
-            case Env.PRODUCER_CONFIG_FILE_PATH:
-              return PRODUCER_CONFIG_PATH;
-            default:
-              throw new IllegalArgumentException();
-          }
+        key -> switch (key) {
```
slinkydeveloper, line 34: This is the new switch? cool!

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/RequestHandler.java
```diff
@@ -91,4 +111,16 @@ public void handle(final HttpServerRequest request) {
     producer.send(record, promise);
     return promise.future();
   }
+
+  @Override
+  public Future<Void> reconcile(Map<Broker, Set<Trigger<CloudEvent>>> objects) {
+
+    final var brokers = objects.keySet().stream()
+        .map(b -> "/" + b.namespace() + "/" + b.name())
```
slinkydeveloper, line 119: Any utils method to format the broker path? (You don't need to do it in this pr)

slinkydeveloper: I see this cloned some times and it might be useful to have it

PR #38, opened 2020-07-11
Update receiver pods annotation to refresh mounted config map

pierDipi:
Part of #33, #37 

## Proposed Changes

- Update receiver pods annotation to refresh mounted config map

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -130,6 +130,12 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, broker *eventing.Broker)
 
 	logger.Debug("Brokers and triggers config map updated")
 
+	// After #37 we reject events to a non-existing Broker, which means that we cannot consider a Broker Ready if all
+	// receivers haven't got the Broker, so update failures to receiver pods is a hard failure.
+	// On the other side, dispatcher pods care about Triggers, and the Broker object is used as a configuration
+	// prototype for all associated Triggers, so we consider that it's fine on the dispatcher side to receive eventually
+	// the update even if here eventually means seconds or minutes after the actual update.
+
 	// Update volume generation annotation of receiver pods
 	if err := r.UpdateReceiverPodsAnnotation(logger, brokersTriggers.VolumeGeneration); err != nil {
 		return statusConditionManager.failedToUpdateReceiverPodsAnnotation(err)
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -130,20 +130,27 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, broker *eventing.Broker)
 
 	logger.Debug("Brokers and triggers config map updated")
 
+	// Update volume generation annotation of receiver pods
+	if err := r.UpdateReceiverPodsAnnotation(logger, brokersTriggers.VolumeGeneration); err != nil {
+		return statusConditionManager.failedToUpdateReceiverPodsAnnotation(err)
+	}
+
 	// Update volume generation annotation of dispatcher pods
 	if err := r.UpdateDispatcherPodsAnnotation(logger, brokersTriggers.VolumeGeneration); err != nil {
 		// Failing to update dispatcher pods annotation leads to config map refresh delayed by several seconds.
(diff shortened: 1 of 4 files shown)


control-plane/pkg/reconciler/broker/broker_test.go
```diff
@@ -136,24 +144,19 @@ func brokerReconciliation(t *testing.T, format string, configs Configs) {
 		{
 			Name: "Reconciled normal - with DLS",
 			Objects: []runtime.Object{
-				NewDeletedBroker(
+				NewBroker(
```
pierDipi, line 147: This test was actually wrong because here we're testing `ReconcileKind` and not `FinalizeKind`, so I fixed it.

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -130,21 +130,28 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, broker *eventing.Broker)
...
+	if err := r.UpdateReceiverPodsAnnotation(logger, brokersTriggers.VolumeGeneration); err != nil {
+		return statusConditionManager.failedToUpdateReceiverPodsAnnotation(err)
+	}
+
 	// Update volume generation annotation of dispatcher pods
 	if err := r.UpdateDispatcherPodsAnnotation(logger, brokersTriggers.VolumeGeneration); err != nil {
 		// Failing to update dispatcher pods annotation leads to config map refresh delayed by several seconds.
-		// For now, we use the configuration map only on the dispatcher side, which means we don't lose availability,
-		// since the receiver will accept events. So, log out the error and move on to the next step.
+		// Since the dispatcher side is the consumer side, we don't lose availability, and we can consider the Broker
```
slinkydeveloper, line 141: Why an update failure on receiver side is considered an hard error, while an update failure on dispatcher side is a soft error?

control-plane/pkg/reconciler/broker/broker_lifecycle.go
```diff
@@ -140,20 +140,25 @@ func (manager *statusConditionManager) reconciled() reconciler.Event {
...
 		corev1.EventTypeWarning,
 		"failed to update dispatcher pods annotation",
 		"%v",
 		err,
 	)
 }
 
+func (manager *statusConditionManager) failedToUpdateReceiverPodsAnnotation(err error) reconciler.Event {
+
+	return fmt.Errorf("failed to update receiver pods annotation: %w", err)
```
slinkydeveloper, line 159: Why you don't record the event here?

control-plane/pkg/reconciler/broker/broker.go
pierDipi: After #37 we reject events to a non-existing Broker, which means that we cannot consider a Broker Ready if all receivers haven't got the Broker, so update failures to receiver pods is a hard failure.

On the other side, dispatcher pods care about Triggers, and the Broker object is used as a configuration prototype for all associated Triggers, so I thought that it's fine on the dispatcher side to receive eventually the update even if here eventually means seconds or minutes after the actual update. 

I'm open to suggestions and discussions about this point because there are implications with both approaches.

control-plane/pkg/reconciler/broker/broker_lifecycle.go
pierDipi: The generated reconciler records errors returned by `ReconcileKind` or `FinalizeKind` with reason "InternalError" and the message is the error message.

control-plane/pkg/reconciler/broker/broker.go
slinkydeveloper: That makes sense for now, i would prefer if you can dump this "difference" in the comments. I see you already pointed out why it's not a problem the failure of dispatcher annotation update, but you didn't explained this difference

control-plane/pkg/reconciler/broker/broker_lifecycle.go
slinkydeveloper: Ok

control-plane/pkg/reconciler/broker/broker.go
pierDipi: Sure.

pierDipi: https://github.com/knative-sandbox/eventing-kafka-broker/pull/38/commits/f962da910445114c3d2be2e08a0b62046889bfa9

PR #39, opened 2020-07-13
Register custom ConditionSet for Broker reconciler

pierDipi:
Related: https://github.com/knative/eventing/pull/3545

## Proposed Changes

- Register custom `ConditionSet` for Broker reconciler
- Test status updates


control-plane/pkg/reconciler/broker/broker_lifecycle.go
```diff
@@ -34,7 +34,7 @@ const (
 	ConditionConfigMapUpdated apis.ConditionType = "ConfigMapUpdated"
 )
 
-var conditionSet = apis.NewLivingConditionSet(
+var ConditionSet = apis.NewLivingConditionSet(
 	ConditionReady,
```
dprotaso, line 38: NewLivingConditionSet - will include a top level 'Ready' - what you're supplying here are dependent conditions

pierDipi: Ok, thank you.

PR #40, opened 2020-07-14
Create dependabot.yml

pierDipi:
## Proposed Changes

- Add `Dependabot` configuration

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


pierDipi:
We need to figure out if we can use the bot or how to enable it.

/hold

pierDipi:
https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates

> You enable GitHub Dependabot version updates by checking a configuration file in to your repository. When GitHub detects the dependabot.yml file, it installs the GitHub Dependabot App, if it isn't already installed (for example, if security updates are enabled). The app then raises pull requests to keep the dependencies you configure up-to-date. For each package manager's dependencies that you want to update, you must specify the location of the package manifest files and how often to check for updates to the dependencies listed in those files.

/unhold

PR #42, opened 2020-07-16
Add Trigger reconciler

pierDipi:
## Proposed Changes

- Add Trigger reconciler


pierDipi:
I'll do the Finalizer part in another PR.

PR #50, opened 2020-07-21
Add partition key value as Producer record key

### LINKED ISSUE #49: Set partitionkey value as ProducerRecord key
**Problem**
`partitionkey` is the cloudevents way of specifying events partitioning, we should set ProducerRecord key to the `partitionkey` value.
If there is no `partitionkey` extension the ProducerRecord key defaults to the empty string.

- [Event producer](https://github.com/knative/eventing/blob/master/docs/personas.md#event-producer)

**Exit Criteria**

**Time Estimate (optional):** 
1

- cloudevevents spec: https://github.com/cloudevents/spec/blob/master/extensions/partitioning.md
- `sdk-go` implementation: https://github.com/cloudevents/sdk-go/blob/cf0fe070fc0babac26cd4d608a0298523aa05442/protocol/kafka_sarama/v2/write_producer_message.go#L24-L57

pierDipi:
Fixes #49 

## Proposed Changes

- Add partition key value as Producer record key

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/cloudevents/PartitionKey.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/cloudevents/PartitionKey.java
@@ -1,0 +1,27 @@
+package dev.knative.eventing.kafka.broker.core.cloudevents;
+
+import io.cloudevents.CloudEvent;
+
+public final class PartitionKey {
+
+  public static final String PARTITION_KEY_KEY = "partitionkey";
+
+  public static final String PARTITION_KEY_DEFAULT = "";
+
+  /**
+   * Extract the partitionkey extension from the given event.
+   *
+   * @param event event from which extracting partition key.
+   * @return partitionkey extension value or PARTITION_KEY_DEFAULT if not set
+   * @link https://github.com/cloudevents/spec/blob/master/extensions/partitioning.md
+   */
+  public static String extract(final CloudEvent event) {
+
+    final var partitionKey = event.getExtension(PARTITION_KEY_KEY);
+    if (partitionKey == null) {
+      return PARTITION_KEY_DEFAULT;
+    }
+
+    return partitionKey.toString();
+  }
+}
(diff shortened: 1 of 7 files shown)


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/cloudevents/PartitionKey.java
```diff
@@ -0,0 +1,27 @@
...
+  public static final String PARTITION_KEY_DEFAULT = "";
+
+  /**
+   * Extract the partitionkey extension from the given event.
+   *
+   * @param event event from which extracting partition key.
+   * @return partitionkey extension value or PARTITION_KEY_DEFAULT if not set
+   * @link https://github.com/cloudevents/spec/blob/master/extensions/partitioning.md
+   */
+  public static String extract(final CloudEvent event) {
```
slinkydeveloper, line 18: Why don't you just return null?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/http/HttpSinkResponseHandler.java
```diff
@@ -70,8 +71,9 @@ public HttpSinkResponseHandler(
     if (event == null) {
       return Future.failedFuture(new IllegalArgumentException("event cannot be null"));
     }
-    // TODO set the correct producer record key
-    return Future.succeededFuture(KafkaProducerRecord.create(topic, "", event));
+
+    final var recordKey = PartitionKey.extract(event);
+    return Future.succeededFuture(KafkaProducerRecord.create(topic, recordKey, event));
```
slinkydeveloper, line 76: the `KafkaProducerRecord` can handle null keys. In fact with null key it implements a partitioning strategy, while with `""` key it ends up putting everything in the same partition

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/cloudevents/PartitionKey.java
pierDipi: Done: https://github.com/knative-sandbox/eventing-kafka-broker/pull/50/commits/fb22a92ee9e995cf973a83be383f1b8678861e0b

PR #53, opened 2020-07-22
Add JSON Structured Logging to data plane components

pierDipi:
## Proposed Changes

- Add JSON Structured Logging to data plane components

### DIFF
--- a/data-plane/config/100-config-logging.yaml
+++ b/data-plane/config/100-config-logging.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kafka-broker-config-logging
+  namespace: knative-eventing
+data:
+  config.xml: |
+    <configuration>
+      <appender name="jsonConsoleAppender" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
+      </appender>
+      <root level="INFO">
+        <appender-ref ref="jsonConsoleAppender"/>
+      </root>
+    </configuration>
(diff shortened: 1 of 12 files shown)


data-plane/config/100-config-logging.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kafka-broker-config-logging
+  namespace: knative-eventing
+data:
+  config.xml: |
```
slinkydeveloper, line 7: an xml inside a yaml FTW

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/Main.java
```diff
@@ -52,6 +55,13 @@
    */
   public static void main(final String[] args) {
 
+    // HACK HACK HACK
+    // maven-shade-plugin doesn't include the LogstashEncoder class, neither by specifying the
+    // dependency with scope `provided` nor `runtime`, and adding include rules to
+    // maven-shade-plugin.
+    // Instantiating an Encoder here we force it to include the class.
```
slinkydeveloper, line 62: What about using include directives? http://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html
So you force to include that class

pierDipi: I tried never got it working :cry:

slinkydeveloper: Ok let's keep it this way and we'll figure out later

PR #60, opened 2020-07-23
Export logs from knative-eventing namespace

pierDipi:
Exporting logs from `knative-eventing` namespace at the end of `e2e` and `conformance` test suites can help to debug failed test runs.

## Proposed Changes

- Export logs from `knative-eventing` namespace at the end of `e2e` and `conformance` test suites.


pierDipi:
I need to check if it works
/hold

PR #63, opened 2020-07-27
Create a new ClusterAdmin at each loop

pierDipi:
Fix `sarama` issue: https://github.com/Shopify/sarama/issues/1565 (in our case is `ClusterAdmin`, we have the same error though)

## Proposed Changes

- Create a new ClusterAdmin at each loop


control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -300,26 +305,23 @@ func (r *Reconciler) ConfigMapUpdated(ctx context.Context) func(configMap *corev
 func (r *Reconciler) SetBootstrapServers(servers string) error {
```
slinkydeveloper, line 305: This function doesn't need to return error anymore

pierDipi: Thank you, fixed!

PR #71, opened 2020-07-31
Add bootstrap servers as configuration

pierDipi:
Bootstrap servers are a list of comma-separated host/port pairs to
use for establishing the initial connection to the Kafka cluster.
To change such a list, a user needs to change multiple ConfigMaps.
With these changes, a user can specify bootstrap servers in only one
ConfigMap, `kafka-broker-config`, and automatically propagate such
list from the control plane to the data plane through the usual
mechanism.

## Proposed Changes

- Add bootstrap servers as configuration


proto/def/triggers.proto
```diff
@@ -39,6 +39,9 @@ message Broker {
 
   // broker name
   string name = 6;
+
+ // A comma separated list of host/port pairs to use for establishing the initial connection to the Kafka cluster.
+  string bootstrapServers = 7;
```
slinkydeveloper, line 44: More than a comma separated list, can you do directly a repeated string field here?

pierDipi: On the data plane, Kafka configuration needs a comma-separated list so I thought it's more efficient to keep that as we need on the data plane avoiding allocating the string from the array for each consumer we instantiate.

PR #72, opened 2020-07-31
Use only one ConfigMap for producer and consumer config

pierDipi:
Two separate ConfigMap to configure consumers and producers
seem to be unnecessary, and they can also produce duplication
and inconsistency, so use only one ConfigMap for producer and
consumer configurations.

## Proposed Changes

- Use only one ConfigMap for producer and consumer config


slinkydeveloper:
Don't you need changes in the controller too?

pierDipi:
No, we don't.

slinkydeveloper:
You're right, sorry :smile: 

/lgtm
/approve

PR #79, opened 2020-08-06
Read Broker configuration from Broker.Spec.Config

### LINKED ISSUE #86: Unnecessary object logs
The generated reconciler adds `<obj-namespace>/<obj-name>`, so we don't need to do it. Instead, we can add the object ID that can be useful for debugging.

## Additional Info

https://github.com/knative-sandbox/eventing-kafka-broker/blob/9f12c4ec49246249a2b9f83de1b71c843b6fb33d/control-plane/pkg/log/logger.go#L28-L34

pierDipi:
Each Broker object now can be individually configured with:
- number of partitions
- replication factor
- bootstrap servers

If there is no `Broker.Spec.Config`, fallback to some defaults.

Fixes #86 

## Proposed Changes

- Read Broker configuration from Broker.Spec.Config


control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -220,7 +227,24 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker)
...
-	topic, err := r.deleteTopic(broker)
+	config, err := r.resolveBrokerConfig(logger, broker)
+	var bootstrapServers []string
+	if err != nil {
+		// An error occurred, try to use default bootstrap servers.
+		r.bootstrapServersLock.RLock()
+		bootstrapServers = r.bootstrapServers
+		r.bootstrapServersLock.RUnlock()
+
+		logger.Error(
```
slinkydeveloper, line 238: I don't think any error logging is needed here, it's a pretty normal use case not overriding the bootstrap servers right?

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -234,28 +258,64 @@ func incrementVolumeGeneration(generation uint64) uint64 {
 	return (generation + 1) % (math.MaxUint64 - 1)
 }
 
-func (r *Reconciler) topicDetailFromBrokerConfig(broker *eventing.Broker) *sarama.TopicDetail {
-	// TODO use broker configurations - see https://github.com/knative-sandbox/eventing-kafka-broker/issues/9
+func (r *Reconciler) resolveBrokerConfig(logger *zap.Logger, broker *eventing.Broker) (Config, error) {
```
slinkydeveloper, line 261: I think this function in case a "specific" broker config exists, it should directly return the merged config between the default one and the specific one, so you avoid scattering the code to merge the 2 configs around

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -234,28 +258,64 @@ func incrementVolumeGeneration(generation uint64) uint64 {
...
 
 	// copy the default topic details
 	topicDetail := r.KafkaDefaultTopicDetails
-	return &topicDetail
+	return topicDetail
 }
 
-func (r *Reconciler) getBrokerConfig(topic string, broker *eventing.Broker) (*coreconfig.Broker, error) {
-	bootstrapServers, err := r.getBootstrapServersOrFail()
+func (r *Reconciler) defaultConfig() (Config, error) {
```
slinkydeveloper, line 300: should return a pointer to `Config`?

control-plane/pkg/reconciler/broker/broker_config.go
```diff
@@ -0,0 +1,72 @@
...
+	err := configmap.Parse(cm.Data,
+		configmap.AsInt32(DefaultTopicNumPartitionConfigMapKey, &topicDetail.NumPartitions),
+		configmap.AsInt32(DefaultTopicReplicationFactorConfigMapKey, &replicationFactor),
+		configmap.AsString(BootstrapServersConfigMapKey, &bootstrapServers),
+	)
+	if err != nil {
+		return Config{}, fmt.Errorf("failed to parse config map %s/%s: %w", cm.Namespace, cm.Name, err)
+	}
+
+	if topicDetail.NumPartitions <= 0 || replicationFactor <= 0 || bootstrapServers == "" {
```
slinkydeveloper, line 50: Are you sure these validations can't be done directly from `configmap.Parse`?

control-plane/pkg/reconciler/broker/broker.go
pierDipi: Sure.

control-plane/pkg/reconciler/broker/broker.go
pierDipi: Yeah, make sense

control-plane/pkg/reconciler/broker/broker_config.go
pierDipi: I don't think so, but I'm thinking to write a `ParseFunc` for each config field, which actually encapsulates the validation, so that while we add new configurations this if doesn't become unreadable, WDYT?

slinkydeveloper: I'm fine with that, but let's do it later, because it requires changes to pkg

pierDipi:
Worked :smiling_face_with_three_hearts:, retesting

pierDipi:
Worked :smiling_face_with_three_hearts:, retesting

PR #89, opened 2020-08-10
Install Eventing release if it's a release branch

pierDipi:
If it's a release branch, we need to install the respective eventing release as we do in `eventing-contrib` here:
https://github.com/knative/eventing-contrib/blob/48505ce5b1e7f370b0b4afe526b739be8481967f/test/e2e-tests.sh#L81-L82.

## Proposed Changes

- Install Eventing release if it's a release branch


slinkydeveloper:
good catch!
/lgtm
/approve

PR #96, opened 2020-08-12
Pin pkg and test-infra to release-0.17 branch

pierDipi:
## Proposed Changes

- Pin and knative/test-infra to release-0.17.
  - In preparation for release-0.17 of knative/eventing in a week.

### DIFF
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -32,7 +32,7 @@ cd "${ROOT_DIR}"
 # The list of dependencies that we track at HEAD and periodically
 # float forward in this repository.
 FLOATING_DEPS=(
-  "knative.dev/test-infra@${VERSION}"
+  "knative.dev/test-infra@release-0.17"
   "knative.dev/eventing@${VERSION}"
 )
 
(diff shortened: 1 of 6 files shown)


hack/update-deps.sh
```diff
@@ -32,7 +32,7 @@ cd "${ROOT_DIR}"
 # The list of dependencies that we track at HEAD and periodically
 # float forward in this repository.
 FLOATING_DEPS=(
-  "knative.dev/test-infra@${VERSION}"
+  "knative.dev/test-infra@release-0.17"
```
slinkydeveloper, line 35: That's how they do in eventing too?

pierDipi: Yes https://github.com/knative/eventing/pull/3818/files#diff-58d4b0ee04af7710eefa27cf91255798R34

hack/update-deps.sh
```diff
@@ -32,7 +32,7 @@ cd "${ROOT_DIR}"
...
   "knative.dev/eventing@${VERSION}"
```
slinkydeveloper, line 36: I suppose we need to align this version too?

pierDipi: Yeah, but not now, a day before our release day, `eventing` will be released and we will pin it here.

slinkydeveloper: SGTM

PR #143, opened 2020-09-03
Update dependencies

pierDipi:
## Proposed Changes

- Update deps (`./hack/update-deps --upgrade`)


pierDipi:
@slinkydeveloper, what should we do here? Implement by ourselves conformance tests?

slinkydeveloper:
> @slinkydeveloper, what should we do here? Implement by ourselves conformance tests?

I don't think it's worth it. For now i would say disable that test/comment it to unblock this one. For me we have two options:

* Or we put more pressure for reverting this one: https://github.com/knative/eventing/pull/3967 (maybe just open the pr to revert it to trigger the discussion)
* Or we wait for the test framework (that we don't really know when it will happen)

IMO rewriting the test from scratch doesn't make sense at this point

pierDipi:
> Or we put more pressure for reverting this one: knative/eventing#3967 (maybe just open the pr to revert it to trigger the discussion)

I get this response from GitHub:

>  Sorry, this pull request couldn’t be reverted automatically. It may have already been reverted, or the content may have changed since it was merged.

which means we cannot revert automatically, and since there is some work to adjust those tests, it would be bad if I work on it, and then it gets rejected.

> Or we wait for the test framework (that we don't really know when it will happen)

Yeah, this isn't an option.

> IMO rewriting the test from scratch doesn't make sense at this point

I see a value in it since we remove a dependency that at this point seems to be too unreliable.

pierDipi:
> For now i would say disable that test/comment it to unblock this one.

TBH I'm not confident in disabling those tests.

slinkydeveloper:
>      I see a value in it since we remove a dependency that at this point seems to be too unreliable.

If you write the test by yourself, it's not a conformance test anymore :smile: And again, then we start having different test frameworks inside eventing and this could make just a whole more complex push this project forward. 

I would say, we're not in a rush for this update so let's wait a bunch of more days and let's talk about this at next eventing delivery wg/eventing wg to raise the problem, if nothing happens for the end of the next week, we just copy paste the tests and we place a big todo for it

slinkydeveloper:
@pierDipi can we copy paste the tests?

pierDipi:
Yes, we can!

slinkydeveloper:
@pierDipi so let's do it and let's put a big todo there

PR #151, opened 2020-09-07
Configure Vert.x HTTP server through a config map

### LINKED ISSUE #150: Http server configuration from config map
**Problem**
We need a way to tweak the vertx http server options

Which persona is this feature for? user

**Exit Criteria**
Like https://github.com/knative-sandbox/eventing-kafka-broker/pull/149, but for http server in receiver side

**Time Estimate (optional):** 
1

slinkydeveloper:
Fixes #150

## Proposed Changes

- 🧽 Some reorg of various env retrievers between dispatcher and receiver
- 🎁 Fix #150 adding configuration for http server

**Release Note**

```release-note
Now you can configure the web client options https://vertx-web-site.github.io/docs/apidocs/io/vertx/core/http/HttpServerOptions.html modifying the config map `config-kafka-broker-data-plane`
```

### DIFF
--- a/data-plane/config/100-config-kafka-broker-data-plane.yaml
+++ b/data-plane/config/100-config-kafka-broker-data-plane.yaml
@@ -113,3 +113,4 @@ data:
     # ssl.trustmanager.algorithm
   config-kafka-broker-webclient.properties: |
     idleTimeout=10000
+  config-kafka-broker-httpserver.properties: |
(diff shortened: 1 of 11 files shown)


data-plane/config/100-config-kafka-broker-data-plane.yaml
```diff
@@ -113,3 +113,4 @@ data:
     # ssl.trustmanager.algorithm
   config-kafka-broker-webclient.properties: |
     idleTimeout=10000
+  config-kafka-broker-httpserver.properties: |
```
pierDipi, line 116: Can you add `idleTimeout` or some default configurations? For example, `idleTimeout` is set to `0` by default which means there is no timeout.

slinkydeveloper: TBH I'm in no position to say if it's a good config or not the default idle timeout, so for "demonstration purpose" I've added the idleTimeout conf as default (to check if the properties works)

pierDipi (approved):
/lgtm

PR #159, opened 2020-09-10
Reorganized data-plane poms

### LINKED ISSUE #6: Enable maven enforcer plugin
**Problem**
Enable maven enforcer plugin to check transitive dependencies convergence: 
- http://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html

Contributor

**Exit Criteria**

**Time Estimate (optional):** 
1

### LINKED ISSUE #7: Use <dependencyManagement> in parent data-plane pom
**Problem**
Now the parent data plane pom contains a lot of dependencies that should be specified in child poms.

These should be removed but the parent pom should have the `<dependencyManagement>` definition to import the Vert.x BOM (https://github.com/vert-x3/vertx-stack#bills-of-materials-maven-bom) and the other dependencies versions (slf4j, junit5, ...).
Then the child modules just need to import the dep without explicating the version (this will keep the version aligned across all modules)

Contributor

**Exit Criteria**

**Time Estimate (optional):** 
1

slinkydeveloper:

Fixes #7, Fixed #6 

## Proposed Changes

- Enabled Maven enforcer plugin
- Reorganized dependencies with dependency management and aligned all the different deps using the maven enforcer plugin
- Renamed the module `generated` to `contract` (There is still the package name to change from `dev.knative.eventing.kafka.broker.core.config` to `dev.knative.eventing.kafka.broker.contract`, but this will be done in another PR)
- Fix jacoco report generation
- Some version bumps


pierDipi (approved):
Great! Thank you!

/lgtm

PR #177, opened 2020-09-15
Add Kafka sink finalizer

### LINKED ISSUE #158: Implement KafkaSink finalizer
**Problem**
We haven't implemented the finalizer in this PR: https://github.com/knative-sandbox/eventing-kafka-broker/pull/134

`*`

**Exit Criteria**
`finalizer + unit tests`

**Time Estimate (optional):** 
1

- The underlying topic might be managed by an external entity (eg it might be created before the CO is created), we need to make sure that we don't delete the topic if it isn't managed by us. To do that we can use the annotation `eventing.knative.dev/topic.owner` in the _**status**_ of the KafkaSink. Possible values of this annotation might be:

  - `kafkasink-controller` if it's managed by us
  - `external` if it's not

cc @matzew @slinkydeveloper if you would like to share your opinion on the additional context design/discussion.

pierDipi:
Fixes #158

## Proposed Changes

- Add Kafka sink finalizer

**Release Note**

```release-note
- 🎁 Add new feature
You can stream Knative HTTP CloudEvents into Apache Kafka using KafkaSink objects
  - See also KafkaSink CRD, types, and reconciler - https://github.com/knative-sandbox/eventing-kafka-broker/pull/134
```

**Docs**

- https://github.com/knative/docs/issues/2814


mattmoor (commented):
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -path './third_party' -prune -o -name '*.pb.go' -prune -o -type f -name '*.go' -print)`
  `goimports -w $(find -name '*.go' | grep -v vendor | grep -v third_party | grep -v .pb.go | grep -v wire_gen.go)`

control-plane/pkg/reconciler/testing/objects_sink.go
```diff
@@ -32,6 +33,7 @@ import (
 	eventing "knative.dev/eventing-kafka-broker/control-plane/pkg/apis/eventing/v1alpha1"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/config"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/base"
+	sinkreconciler "knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/sink"
```
mattmoor, line 36: Format Go code:
```suggestion

```

control-plane/pkg/reconciler/testing/objects_sink.go
```diff
@@ -32,6 +33,7 @@ import (
...
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/kafka"
```
mattmoor, line 37: Format Go code:
```suggestion
	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/kafka"
	sinkreconciler "knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/sink"
```

slinkydeveloper:
/hold

I prefer to keep this on hold until we merge #162

slinkydeveloper (commented):
minor nit

matzew:
@pierDipi  should I PR some doc, like:

https://gist.github.com/matzew/e2c2fcd2696a346f25b8bc9e64bfd0fa

pierDipi:
@matzew Sure, feel free to assign yourself to the issue: https://github.com/knative/docs/issues/2814, I appreciate your help!

pierDipi:
```
my-cluster-zookeeper-0                     1/1   Running             0     5m5s
my-cluster-zookeeper-1                     1/1   Running             0     5m5s
my-cluster-zookeeper-2                     0/1   ContainerCreating   0     5m5s
strimzi-cluster-operator-9968fd8c9-vwq4m   1/1   Running             0     5m31s
```
Zookeeper ...
/retest

pierDipi:
```
kafka.kafka.strimzi.io/my-cluster created
Waiting until all pods in namespace kafka are up......................................................................................................................................................
strimzi-cluster-operator-9968fd8c9-cs2kz   0/1   ImagePullBackOff   0     5m30s
Failed Pod (data in YAML format) - strimzi-cluster-operator-9968fd8c9-cs2kz
```
/retest

pierDipi:
```
my-cluster-zookeeper-0                     1/1   Running             0     5m8s
my-cluster-zookeeper-1                     1/1   Running             0     5m8s
my-cluster-zookeeper-2                     0/1   ContainerCreating   0     5m8s
strimzi-cluster-operator-9968fd8c9-92vh6   1/1   Running             0     5m32s
```
Zookeeper,
/retest

pierDipi:
```
 operation.go:75: Failed to get {Pod v1}-first-filtered-event-sender ready: <nil>
```
https://github.com/knative-sandbox/eventing-kafka-broker/pull/183 try to remove flakes,
/retest

control-plane/pkg/reconciler/sink/kafka_sink.go
```diff
@@ -71,19 +80,34 @@ func (r *Reconciler) reconcileKind(ctx context.Context, ks *eventing.KafkaSink)
...
+		isPresentAndValid, err := r.ClusterAdmin.IsTopicPresentAndValid(ks.Spec.Topic, ks.Spec.BootstrapServers)
+		if err != nil {
+			return statusConditionManager.TopicNotPresentOrInvalidErr(err)
+		}
+		if !isPresentAndValid {
+			// The topic might be invalid.
+			return statusConditionManager.TopicNotPresentOrInvalid()
+		}
 	}
 	statusConditionManager.TopicCreated(ks.Spec.Topic)
```
slinkydeveloper, line 112: More than using always the same status, can we have a status for "topic created owned by controller" and "topic ready owned by external"? It provides more verbosity of what the controller does and maybe we can use it even to distinguish owned/not owned topics, avoiding the annotation

control-plane/pkg/reconciler/sink/kafka_sink.go
```diff
@@ -161,12 +185,62 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ks *eventing.KafkaSink) r
 
 func (r *Reconciler) finalizeKind(ctx context.Context, ks *eventing.KafkaSink) error {
 
-	// logger := log.Logger(ctx, "finalize", ks)
+	logger := log.Logger(ctx, "finalize", ks)
+
+	// Get sinks config map.
+	sinksConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx)
```
slinkydeveloper, line 191: `contractConfigMap`

control-plane/pkg/reconciler/sink/kafka_sink.go
```diff
@@ -161,12 +185,62 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ks *eventing.KafkaSink) r
...
+	if err != nil {
+		return fmt.Errorf("failed to get sinks config map %s: %w", r.Configs.DataPlaneConfigMapAsString(), err)
```
slinkydeveloper, line 193: contract config map

control-plane/pkg/reconciler/sink/kafka_sink.go
```diff
@@ -161,12 +185,62 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ks *eventing.KafkaSink) r
...
+	}
+
+	logger.Debug("Got sinks config map")
+
+	// Get sinks data.
+	sinks, err := r.GetDataPlaneConfigMapData(logger, sinksConfigMap)
```
slinkydeveloper, line 199: contract

control-plane/pkg/reconciler/sink/kafka_sink.go
```diff
@@ -161,12 +185,62 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ks *eventing.KafkaSink) r
...
-	// logger := log.Logger(ctx, "finalize", ks)
+	logger := log.Logger(ctx, "finalize", ks)
+
+	// Get sinks config map.
+	sinksConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get sinks config map %s: %w", r.Configs.DataPlaneConfigMapAsString(), err)
+	}
+
+	logger.Debug("Got sinks config map")
```
slinkydeveloper, line 196: contract

control-plane/pkg/reconciler/sink/kafka_sink.go
```diff
@@ -161,12 +185,62 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, ks *eventing.KafkaSink) r
...
+			return err
+		}
+		logger.Debug("Topic deleted", zap.String("topic", topic))
+	}
 
-	// TODO implement finalizer
 	return nil
 }
 
 func incrementGeneration(generation uint64) uint64 {
```
slinkydeveloper, line 234: This function is cloned in several places, can you move it to core? (in another pr is fine too)

slinkydeveloper (commented):
Thanks for fixing some of my leftovers of the previous pr too!

control-plane/pkg/reconciler/sink/kafka_sink.go
pierDipi: The annotation isn't designed for us to distinguish the two cases since `numPartitions` and `replicationFactor` are immutable we can always infer who owns the topic.
The reason for using the annotation is to allow users or machines (since it's structured) using the status regardless of the internal implementation that might vary, and it's an information that is already in the status, so I'm not sure adding verbosity in the status is a pattern I've seen in k8s, it seems more appropriate for a log message. WDYT?

control-plane/pkg/reconciler/sink/kafka_sink.go
pierDipi: I'm not sure it's better it just vaguer :thinking:

pierDipi:
@slinkydeveloper I'm not sure decontextualizing variable names and log messages is better, why do you suggest to use contract everywhere instead of specific names? 
It seems too vague to me and the names shouldn't be the same as the type of the variable. 
We know that in the contract there are sinks, why do we need to call the variable contract and not sinks? If it is for consistency I'll change names everywhere to be more specific.

control-plane/pkg/reconciler/sink/kafka_sink.go
pierDipi: Sure in a follow-up.

control-plane/pkg/reconciler/sink/kafka_sink.go
slinkydeveloper: It's more to keep it consistent

control-plane/pkg/apis/eventing/v1alpha1/kafka_sink_lifecycle.go
```diff
@@ -19,16 +19,14 @@ package v1alpha1
 import "knative.dev/pkg/apis"
 
 const (
-	ConditionAddressable      apis.ConditionType = "Addressable"
-	ConditionTopicReady       apis.ConditionType = "TopicReady"
-	ConditionConfigMapUpdated apis.ConditionType = "ConfigMapUpdated"
+	ConditionAddressable        apis.ConditionType = "Addressable"
```
mattmoor, line 22: Format Go code:
```suggestion
	ConditionAddressable apis.ConditionType = "Addressable"
```

mattmoor (commented):
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -path './third_party' -prune -o -name '*.pb.go' -prune -o -type f -name '*.go' -print)`
  `goimports -w $(find -name '*.go' | grep -v vendor | grep -v third_party | grep -v .pb.go | grep -v wire_gen.go)`

slinkydeveloper:
@pierDipi the reason to keep those names is:

* Name the variable as the entity it contains
* The config map contains a json with `contract` entity, to me logging `sinks config map` is far more confusing because i don't really see any sinks entity anywhere. It's the business logic that gives the sinks meaning to the contract entity
* One of the things I understood is that, in the long term, we might use the data plane indipendently from what is actually implementing: we could have a single data plane deployment and you could have in the same cm both sinks and brokers, the controller operates on all these. in that case, contract is contract :smile:

control-plane/pkg/reconciler/sink/kafka_sink.go
slinkydeveloper: On the contrary, to me this really sounds like a status thing: `TopicAlreadyPresent` vs `NewTopicCreated`. @matzew we need a third opinion on this one :smile:

pierDipi:
Yes, I updated it.

pierDipi:
> One of the things I understood is that, in the long term, we might use the data plane indipendently from what is actually implementing: we could have a single data plane deployment and you could have in the same cm both sinks and brokers, the controller operates on all these. in that case, contract is contract 

I would keep them separately at least in the default setup, and you can do that already today by changing the env variables of the controller. :wink:

control-plane/pkg/reconciler/sink/kafka_sink.go
pierDipi: Conditions should be fixed before the reconciliation happens, see KRShaped thing in pkg.

pierDipi: We can change that but we lose some useful features.

pierDipi: We can emit an event in the CO context, but the general recommendation in the no-op case is to not emit events: https://github.com/knative/pkg/issues/1520, and today we don't really differentiate if we actually create or not the topic in Kafka since we suppress the `TopicAlreadyExistsError`.

pierDipi: > On the contrary, to me this really sounds like a status thing: TopicAlreadyPresent vs NewTopicCreated

If this is the case the distinction should be in the Kafka channel as well, but there isn't such distinction: https://github.com/knative/eventing-contrib/blob/9547d5559586095acfbec6ffa39e8e52d6aee004/kafka/channel/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go#L27

pierDipi: Let me know what do you think, I just wanted to point out the implication of such a change.

slinkydeveloper:
> "contract" isn't a thing from a user perspective since they don't know anything about our terminology, anyway I updated it. :)

But that's something they actually see in the protobuf schema and in the json inside the config map, that's what i mean. And end up naming with 2 different names the same thing in the logs is not a good idea too

control-plane/pkg/reconciler/sink/kafka_sink.go
slinkydeveloper: @pierDipi in the case of the channel, the distinction doesn't exist because the controller always owns the topic

matzew: right, the topic of the channel is a different beast: it's more for backing the messages flying in

Source (and Sink), is totally different.  In most cases the topic will be created by an "external" entiy - most production clusters even disallow the creation of a topic if it does not exist.

slinkydeveloper: Agree with @matzew . I think you can log who owns the topic in the condition of topic ready and that should be enough for the user

pierDipi: I misunderstood your comment, you meant the reason and not a different condition, I was confused by NewTopicCreated vs ...., my bad :), and the channel example was a really bad example.

pierDipi:
Ready for another pass!

slinkydeveloper:
/lgtm
/approve

Thanks for another huge contribution!

PR #210, opened 2020-09-24
Log error on malformed event in response

### LINKED ISSUE #194: [Tracking - eventing#4071] Log warning on malformed event in response
## Expected Behavior

 - https://github.com/knative/eventing/issues/4071

## Actual Behavior

- We don't log anything.

## Steps to Reproduce the Problem

- https://github.com/knative/eventing/issues/4071

## Additional Info

cc @slinkydeveloper 

/kind enhancement

pierDipi:
Fixes #194 

When a sink returns a non-event and a non-empty response there is no signal that
the event is malformed we just discard it, now we log the error.

## Proposed Changes

- Return a failed future when the response is a non-event and the response is non-empty.


aliok (approved):
/lgtm
/approve

PR #199, opened 2020-09-22
Check if codecov likes 2 unit test tasks

slinkydeveloper:
Replace #198

## Proposed Changes

- Java unit tests in a separate gh action

PR #213, opened 2020-09-24
Upload Codecov Java report

pierDipi:
It's different from https://github.com/knative-sandbox/eventing-kafka-broker/pull/199 since in this one we upload reports in two different actions independently with two different flags.

## Proposed Changes

- Upload Codecov Java report


pierDipi:
It seems to be working: https://codecov.io/gh/knative-sandbox/eventing-kafka-broker/pull/213?src=pr&el=continue, also see the table of the report message above.

slinkydeveloper:
Ah, two different flags does the trick, thanks!

/lgtm

PR #265, opened 2020-10-07
Jlink to reduce container image size

### LINKED ISSUE #10: Use Jlink to reduce container image size
**Problem**
A short explanation of the problem, including relevant restrictions.

- System Integrator

**Exit Criteria**
Reduce container image size ( < 300MB )

**Time Estimate (optional):** 
1-4

- https://docs.oracle.com/en/java/javase/11/tools/jlink.html

slinkydeveloper:
Fixes #10 

Now image size (receiver) is 184Mb, compared to previous 478Mb

This is a first pass on the issue, there are some mods picked by jdeps that we could eventually remove (like java.sql, that we don't use but we bring in b/c of jackson-databind). Also, we're on ubuntu base, we could drill down the image even further porting to alpine (but that requires musl libc and we need to figure out how to make that working with jlink).

## Proposed Changes

- Use jlink to create a smaller jdk, in order to reduce image size of our deliverables


matzew:
nice one, Slinky!

data-plane/docker/Dockerfile
```diff
@@ -13,13 +13,17 @@
 # limitations under the License.
 
 ARG JAVA_IMAGE
+ARG BASE_IMAGE
+FROM ${JAVA_IMAGE} as builder
```
pierDipi, line 17: nit: add an empty line above for visual separation.

data-plane/docker/generate_jdk.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
```
pierDipi, line 1: Since we use bash everywhere can we use it here as well? It's for our own sanity :smile:

pierDipi (commented):
Cool!

data-plane/docker/generate_jdk.sh
slinkydeveloper: I used sh because i was initially trying with alpine, that has only sh and not bash. I prefer to keep as is, so on the next iteration i can work on using alpine more than ubuntu

pierDipi: Ok, thanks!

pierDipi (approved):
Can you open an issue to keep track of the next iteration?

/lgtm

slinkydeveloper:
Ok it seems like the vert.x dns works, but it requires the module of the jdk dns, funky

slinkydeveloper:
This PR looks good to me, @pierDipi wanna merge?

pierDipi (approved):
/lgtm
:tada:

PR #268, opened 2020-10-08
The Broker retries sending events 

### LINKED ISSUE #116: Support full Delivery Spec
**Problem**
`broker.spec.delivery` has parameters to specify `retries`, `backoffDelay`, and `backoffPolicy`, we need to support them.

- [x] Update the control plane to add those configs to the ConfigMap
- [x] Update the data plane to act on those configs

Event Consumer

**Exit Criteria**
Pass `TestBrokerRedelivery` test (https://github.com/knative/eventing/blob/master/test/e2e/broker_redelivery_test.go)

**Time Estimate (optional):** 
1

- https://pkg.go.dev/knative.dev/eventing/pkg/apis/duck/v1?tab=doc#DeliverySpec
- We already support `deadLetterSink`

/kind feature-request
/area data-plane
/area control-plane

pierDipi:
Fixes #116 

## Proposed Changes

- Retry sending events through the circuit breaker
- Add E2E tests for the redelivery feature 

**Release Note**

```release-note
- 🎁 Add new feature
The Broker retries sending events.
```


pierDipi:
2/2 Worked, retesting

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/http/HttpConsumerVerticleFactory.java
```diff
@@ -181,11 +193,42 @@ protected CircuitBreakerOptions createCircuitBreakerOptions(
...
 
     return new HttpConsumerRecordSender(
       client,
-      target
+      target,
+      circuitBreaker
     );
   }
+
+  private void setRetryPolicy(EgressConfig egress, CircuitBreaker circuitBreaker) {
```
slinkydeveloper, line 210: More than passing the `circuitBreaker` to configure it, can you return the `Function`?

slinkydeveloper: Also maybe name the method as `computeRetryPolicy`

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/http/HttpConsumerVerticleFactory.java
```diff
@@ -181,11 +193,42 @@ protected CircuitBreakerOptions createCircuitBreakerOptions(
...
+        final var delay = Duration.parse(egress.getBackoffDelay()).toMillis();
+
+        final Function<Integer, Long> retryPolicy = switch (egress.getBackoffPolicy()) {
+          case Linear -> retryCount -> linearRetryPolicy(retryCount, delay);
+          case Exponential, UNRECOGNIZED -> retryCount -> exponentialRetryPolicy(retryCount, delay);
+        };
+
+        circuitBreaker.retryPolicy(retryPolicy);
+      } catch (final Exception ex) {
+        logger.error("failed to set retry policy", ex);
```
slinkydeveloper, line 222: Should this exception be propagated?

pierDipi: The only method that can throw the expectation is `parse` and if it fails we have two possible cases:
- It's an empty string (so it's ok to not propagate it)
- It's not a valid ISO 8601 string (so it's a control plane bug)

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/http/HttpConsumerVerticleFactory.java
pierDipi: I like it!

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/http/HttpConsumerVerticleFactory.java
pierDipi: I'm thinking to change the backoff delay's type in our proto in a follow-up PR so we won't have this problem anymore.

slinkydeveloper: > It's an empty string (so it's ok to not propagate it)

Why don't you check if the string is empty in the if and throw an IllegalStateException for that?

pierDipi: I'll change the type to contain directly milliseconds so we don't have this problem at all, also because we should keep in mind the upstream issue: https://github.com/knative/eventing/issues/4231, can I do this in a separate PR?

slinkydeveloper: Ok I'm fine with it

pierDipi:
Rebased.

Issue #267, opened 2020-10-07
Reduce container image size


slinkydeveloper:
**Problem**
#265 started reducing receiver and dispatcher container image size, but we can improve it further

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for? User

**Exit Criteria**
Smaller container image for receiver and dispatcher

**Time Estimate (optional):** 
2

**Additional context (optional)**
These are the points we should investigate on:

* There are some mods picked by jdeps that we could eventually remove. E.g. We don't use `java.sql` but we bring in b/c of jackson-databind. We should investigate if this can be removed
* We're on ubuntu base, we should check if it's feasible to run on smaller base images like alpine, or even better from scratch. In the case of alpine, it requires musl libc and we need to figure out how to make that working with jlink
* Why do we bring in `java.compiler`?
* Why do we bring in `jdk.management`?

To remove the modules from the generated jdk, you just need to modify the `generate_jdk.sh` script:

```
MODS=$(echo "$MODS" | sed '/^java.compiler/d' | sed '/^java.sql/d')
```

PR #270, opened 2020-10-08
Removed `java.compiler` from used jmods

slinkydeveloper:

Followup of #265, part of #267

## Proposed Changes

- Removed jmods we probably don't need
  * `java.compiler` contains the compiler types. It's brought in by Guava and Vert.x CLI feature. we don't use the Vert.x Cli feature, so we don't need this

### DIFF
--- a/data-plane/docker/generate_jdk.sh
+++ b/data-plane/docker/generate_jdk.sh
@@ -10,7 +10,7 @@ MODS=$(jdeps -q --print-module-deps --ignore-missing-deps "$1")
 echo "Computed mods = '$MODS'"
 
 # Remove compiler, sql, management modules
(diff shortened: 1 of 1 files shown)


pierDipi:
worked, retesting

pierDipi:
Can you share some numbers?

slinkydeveloper:
Removing just compiler i think won't make any difference, but let me try

slinkydeveloper:
Yep it's just 1Mb saved... Well, better than nothing :smile: @pierDipi do you think we should merge it or discard it and leave as is?

pierDipi (commented):
I'm ok with merging it.

One thing: can you retitle the PR and update the body?

/lgtm

PR #284, opened 2020-10-12
Use milliseconds for backoff delay in the contract

### LINKED ISSUE #281: Use milliseconds in contract for backoff delay
**Problem**

> I'll change the type to contain directly milliseconds so we don't have this problem at all, also because we should keep in mind the upstream issue: [knative/eventing#4231](https://github.com/knative/eventing/issues/4231), can I do this in a separate PR?

Originally posted: https://github.com/knative-sandbox/eventing-kafka-broker/pull/268#discussion_r501629198

- Developer

**Exit Criteria**
Use milliseconds in the contract for backoff delay

**Time Estimate (optional):** 
1

/assign
/area control-plane
/area data-plane

pierDipi:
Using the raw string for backoff delay in our contract can lead
to an illegal state on the data plane and such a state isn't reflected
in the status of our resources. Also, it's not clear how to deal
with those exceptions since propagating them leads to a data plane
crash loop and the data plane won't be able to recover, or logging 
the error and move forward leads to a silent error suppression 
and consequently unwanted behavior.

Using milliseconds of type uint64 removes uncertainty and the
following behavior applies:
- `backoffDelay = 0 => no retries`
- `backoffDelay > 0 => milliseconds to wait before retrying`

Fixes #281

## Proposed Changes

- Use milliseconds for backoff delay in the contract


control-plane/pkg/core/config/utils_test.go
```diff
@@ -104,25 +104,48 @@ func TestBackoffDelayFromString(t *testing.T) {
 	tests := []struct {
 		name         string
 		backoffDelay *string
-		defaultDelay string
-		want         string
+		defaultDelay uint64
+		want         uint64
+		wantError    bool
 	}{
 		{
```
slinkydeveloper, line 111: Another test case with microseconds? To check if something breaks or the result is properly 0 milliseconds :smile:

pierDipi: Done https://github.com/knative-sandbox/eventing-kafka-broker/pull/284/commits/5b51863d2049ab020608df9a31cae32e5ac9f140

slinkydeveloper:
@pierDipi can you fix the code style lint?

pierDipi:
```
  Error: [trailing whitespace] reported by reviewdog 🐶
          
  
  Raw Output:
  ./data-plane/contract/src/main/java/dev/knative/eventing/kafka/broker/contract/DataPlaneContract.java:1989:
```
It's generated code :sob:

slinkydeveloper:
Ah, sorry i didn't looked carefully... Can you open an issue in https://github.com/knative-sandbox/.github for that?

/lgtm
/approve

PR #288, opened 2020-10-12
Remove config-logging volume and add server config for the sink

### LINKED ISSUE #287: Remove config-logging volume from the controller
The config-logging volume isn't necessary and we should remove it, see:

https://github.com/knative-sandbox/eventing-kafka-broker/blob/5c63b35d2e539e0995d5cd44dcaa53944c569455/control-plane/config/500-controller.yaml#L114-L117
and
https://github.com/knative-sandbox/eventing-kafka-broker/blob/5c63b35d2e539e0995d5cd44dcaa53944c569455/control-plane/config/500-controller.yaml#L56-L59

## Additional information
/area control-plane
/kind cleanup
/good-first-issue

pierDipi:
The `config-logging` volume mount isn't necessary.

Fixes #287 

## Proposed Changes

- Remove config-logging volume and add server config for the sink

**Release Note**

```release-note
- 🧽 Update or clean up current behavior
Remove config-logging volume from the controller
```

### DIFF
--- a/control-plane/config/500-controller.yaml
+++ b/control-plane/config/500-controller.yaml
@@ -53,10 +53,6 @@ spec:
         - name: controller
           image: ko://knative.dev/eventing-kafka-broker/control-plane/cmd/kafka-controller
           imagePullPolicy: IfNotPresent
-          volumeMounts:
-            - name: config-logging
-              mountPath: /etc/config-logging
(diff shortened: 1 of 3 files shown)


slinkydeveloper:
Good catch!

PR #306, opened 2020-10-16
Reduce container images size

### LINKED ISSUE #267: Reduce container image size

pierDipi:
This PR reduces the container image's sizes from `187MB` to `90MB +- 1MB`.

I worked with the [distroless community](https://github.com/GoogleContainerTools/distroless) to build non-root Java images,
the base image contains only system libraries (`glibc`, `zlib`, etc) and 
it runs as non-root, uid: `65532`.

- Base image size: ~30MB.
- Jar and the custom Java environment size: ~60MB.

Layers:
```
sha256:3041eb8028850a81494bb10b0843ae7a0822578bba3901b90dc63557f9437e9d   2 minutes ago       /bin/sh -c #(nop)  ENV PATH=/app/jdk/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin   0B                  
sha256:4ff88c0b75a38046f0c99c15333b31dd35de4b69328c981e368d2a92ded5cc20   2 minutes ago       /bin/sh -c #(nop) WORKDIR /app                                                                          0B                  
sha256:fa682947dcbf8b3ed768fe42c6f7e5660423385a69002145ac287d6e486c0fe8   2 minutes ago       /bin/sh -c #(nop) COPY dir:e92f937cea101df7dc902550b122be947360f8bec413e3b503f595dbedd5912e in /app     59.9MB              
sha256:a92e78c22ba509a04f3eb070f0b6978559d75fee91ad79b6f5e6fc5ab75bca6f   50 years ago        bazel build ...                                                                                         8.64MB              
<missing>                                                                 50 years ago        bazel build ...                                                                                         1.97MB              
<missing>                                                                 50 years ago        bazel build ...                                                                                         17.4MB              
<missing>                                                                 50 years ago        bazel build ...                                                                                         1.8MB               
```

Fixes #267 

## Proposed Changes

- Reduce container image sizes

**Release Note**

```release-note
🧽 Improvements
The container image's sizes are ~90MB instead of ~287MB.
```

### DIFF
--- a/data-plane/config/sink/template/500-receiver.yaml
+++ b/data-plane/config/sink/template/500-receiver.yaml
@@ -36,7 +36,7 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
-        runAsUser: 999
+        runAsUser: 65532
       containers:
         - name: kafka-sink-receiver
           image: ${KNATIVE_KAFKA_SINK_RECEIVER_IMAGE}
--- a/data-plane/config/template/500-dispatcher.yaml
+++ b/data-plane/config/template/500-dispatcher.yaml
@@ -36,7 +36,7 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
-        runAsUser: 999
+        runAsUser: 65532
       containers:
         - name: kafka-broker-dispatcher
           image: ${KNATIVE_KAFKA_BROKER_DISPATCHER_IMAGE}
--- a/data-plane/config/template/500-receiver.yaml
+++ b/data-plane/config/template/500-receiver.yaml
@@ -36,7 +36,7 @@ spec:
     spec:
       securityContext:
         runAsNonRoot: true
-        runAsUser: 999
+        runAsUser: 65532
       containers:
         - name: kafka-broker-receiver
           image: ${KNATIVE_KAFKA_BROKER_RECEIVER_IMAGE}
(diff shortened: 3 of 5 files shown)


slinkydeveloper:
Now I see, so you're still using the jlink thingy to generate the jdk but you're changing the base image to one that contains only the shared libs required for the jvm to run, right? but that `java-debian10:base-nonroot` base image does not already contain the jdk?

pierDipi:
No, there is no JDK, see https://github.com/GoogleContainerTools/distroless/blob/a71e663938c986f0bfd79938b93fdd3d03599b8a/java/BUILD#L22-L52.

slinkydeveloper:
I see, it's just shared libs, i like it

slinkydeveloper (approved):
/lgtm
/approve

PR #308, opened 2020-10-19
Improve error messages in kafka-consumer test image 

pierDipi:
If the job fails, it's hard to diagnose what went wrong, this
PR adds some more logging to help diagnose errors.

## Proposed Changes

- Improve error messages in the `kafka-consumer` test image


pierDipi:
Flaky
/retest

PR #334, opened 2020-10-23
Gracefully clean up resources on shutdown

pierDipi:
## Proposed Changes

- Gracefully clean up resources on shutdown

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/Main.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/Main.java
@@ -126,8 +126,8 @@ public class Main {
 
     } catch (final Exception ex) {
       logger.error("Failed during filesystem watch", ex);
-    }
 
-    Shutdown.closeSync(vertx).run();
+      Shutdown.closeSync(vertx).run();
+    }
   }
 }
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/Main.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/Main.java
@@ -33,7 +33,6 @@ import io.vertx.ext.web.client.WebClient;
 import io.vertx.ext.web.client.WebClientOptions;
 import io.vertx.micrometer.backends.BackendRegistries;
 import java.io.File;
-import java.io.IOException;
 import java.nio.file.FileSystems;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
@@ -60,7 +59,7 @@ public class Main {
    *
    * @param args command line arguments.
    */
-  public static void main(final String[] args) throws IOException, InterruptedException {
+  public static void main(final String[] args) {
     // HACK HACK HACK
     // maven-shade-plugin doesn't include the LogstashEncoder class, neither by specifying the
     // dependency with scope `provided` nor `runtime`, and adding include rules to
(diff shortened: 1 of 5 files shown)


slinkydeveloper (approved):
/lgtm

pierDipi:
I'm fixing another thing.
/hold

pierDipi:
Ready for a review.
/cc @slinkydeveloper

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/Main.java
```diff
@@ -108,26 +113,46 @@ public static void main(final String[] args) {
...
+        vertx.close(v -> System.exit(1));
+      });
+    waitVerticle.await(5, TimeUnit.SECONDS);
+
+    final var publisher = new ContractPublisher(vertx.eventBus(), ResourcesReconcilerMessageHandler.ADDRESS);
+    final var fs = FileSystems.getDefault().newWatchService();
+    var fw = new FileWatcher(fs, publisher, new File(env.getDataPlaneConfigFilePath()));
+
+    // Gracefully clean up resources.
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
```
slinkydeveloper, line 134: Without makind complex abstractions, can we wrap part of this shutdown code somewhere (like a static method) and place it in core? I see some cloned code here

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/Main.java
```diff
@@ -72,13 +75,13 @@ public static void main(final String[] args) {
     final var vertx = Vertx.vertx(
       new VertxOptions().setMetricsOptions(MetricsOptionsProvider.get(env, METRICS_REGISTRY_NAME))
     );
+    Runtime.getRuntime().addShutdownHook(new Thread(Shutdown.closeSync(vertx)));
```
slinkydeveloper, line 78: Why this? Just add one shutdown hook?

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/Main.java
```diff
@@ -76,6 +79,8 @@ public static void main(final String[] args) {
     final var vertx = Vertx.vertx(
       new VertxOptions().setMetricsOptions(MetricsOptionsProvider.get(env, METRICS_REGISTRY_NAME))
     );
+    Runtime.getRuntime().addShutdownHook(new Thread(Shutdown.closeSync(vertx)));
```
slinkydeveloper, line 82: Same as above

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/Main.java
pierDipi: If it fails before registering the other hook and after the verticle deploy from a readiness/liveness probe perspective the pod is ready, so closing Vertx makes sure we report not ready as soon as possible. WDYT?

slinkydeveloper: If something fails during the main, i think it doesn't make any difference having this hook or not.

pierDipi: What happens if the application receives a SIGTERM with no hook registered?

slinkydeveloper: well, give it a try :smile: run locally the jar and check what happens

slinkydeveloper: I was looking here https://vertx-web-site.github.io/docs/vertx-core/java/#_causing_vert_x_to_exit so we need anyway to close vertx to close the application, i was wrong in the previous comment

pierDipi: I've already tried and that is the reason for this PR. 
What happens is that the main thread finishes but the pod continues to be ready because the verticle server continues to respond.

slinkydeveloper: uhm, wrapping in the try catch all the "potentially failing" things after the creation of vertx makes sense?

My feeling is that adding 2 shutdown hooks that close vertx might be problematic

slinkydeveloper: When you add the second shutdown hook, you still have the first one mounted. So one will close vertx, the other won't do anything i think, because it's trying to close something already closed

pierDipi: Calling close twice isn't problematic. I've checked the implementation.

> When you add the second shutdown hook, you still have the first one mounted. So one will close vertx, the other won't do anything i think, because it's trying to close something already closed

But the point is making sure we close Vertx in middle as well.

slinkydeveloper: Ok what happens if the first one you register is invoked before the second one? You won't have the shutdown as expected right?

slinkydeveloper: That's why I'm saying: try catch to wrap them all sounds reasonable and simpler to understand

pierDipi: Ok, please check the last commit.

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/Main.java
```diff
@@ -75,61 +74,60 @@ public static void main(final String[] args) throws IOException, InterruptedExce
...
+
       fw.watch(); // block forever
 
     } catch (final Exception ex) {
       logger.error("Failed during filesystem watch", ex);
-      Shutdown.closeSync(vertx).run();
-      throw ex;
     }
+
+    Shutdown.closeSync(vertx).run();
```
slinkydeveloper, line 131: I admit that I'm still lost, why you added this outside the catch?

pierDipi: Yep, I already fixed it!

PR #364, opened 2020-10-30
Test on KinD

### LINKED ISSUE #290: Test on KinD 1.17-1.19
**Problem**
Run e2e tests on Github Actions with KinD to get coverage of supported version range: 1.17 - 1.19

- Developer

**Exit Criteria**
e2e tests run on KinD

**Time Estimate (optional):** 
1

We should be able to enable just by copy-pasting the action from eventing and change installation steps:
- Action: https://github.com/knative/eventing/blob/master/.github/workflows/kind-e2e.yaml
- PR: https://github.com/knative/eventing/pull/4273/files

/help

pierDipi:
Fixes #290

## Proposed Changes

-  Test on KinD

### DIFF
--- a/hack/update_strimzi.sh
+++ b/hack/update_strimzi.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+# After running this script you need to add the env variable KUBERNETES_SERVICE_DNS_DOMAIN to .cluster.local in order to use 
+# a different DNS suffix to the strimzi-cluster-operator Deployment.
+# `.cluster.local` will be replaced to the generated one on GitHub Actions.
+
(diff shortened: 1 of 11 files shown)


hack/update_strimzi.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
slinkydeveloper, line 1: TBH I prefer to avoid committing the strimzi setup and always download from GH like on master, is that possible or is it causing issues?

pierDipi: We need to do the dance to change the cluster suffix manually.

pierDipi: Also this makes our builds reproducible.

slinkydeveloper: Ok I just checked eventing-kafka, they have the strimzi setup file committed, so ok let's keep it here. Can you add a comment explaining that this script updates the strimzi file we use in builds?

pierDipi:
Updated.

Waiting for GitHub Actions to succeed.
/hold

PR #372, opened 2020-10-31
Explicit parallel test run

pierDipi:
Let's see if this speeds things up.

## Proposed Changes

- Explicit parallel test run

/hold


pierDipi:
No evident benefits :disappointed:

PR #397, opened 2020-11-11
[WIP] Point to SNAPSHOT SDK

pierDipi:
Let's see how this goes.
Ref: https://github.com/knative-sandbox/eventing-kafka-broker/pull/393


pierDipi:
@slinkydeveloper this one passes, so that's the bug.

slinkydeveloper:
Ok I think we can just skip milestone4 of the sdk, no big loss for us

PR #423, opened 2020-11-23
Add data plane licenses check

pierDipi:
This PR adds a check for license headers and creates a file with all third-party licenses.

The plugin adds the inception year instead of the current year. (I think it's not a big deal)
 
## Proposed Changes

- Add data plane licenses check
- Add third-party licenses file.


slinkydeveloper (approved):
/lgtm
/approve

Thanks for tackling this one

PR #455, opened 2020-12-03
Add data plane profiling action

pierDipi:
This PR adds a GitHub Action to profile our data plane, and upload
the resulting artifacts so that reviewer can inspect results, and
take actions accordingly.

## Proposed Changes

- Add data plane profiling action

### DIFF
--- a/data-plane/profiler/run.sh
+++ b/data-plane/profiler/run.sh
@@ -1,0 +1,158 @@
+#!/usr/bin/env bash
+#
+# Copyright © 2018 Knative Authors (knative-dev@googlegroups.com)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Set runtime variables
+# echo 1 > /proc/sys/kernel/perf_event_paranoid
+# echo 0 > /proc/sys/kernel/kptr_restrict
+
+EVENT=${EVENT:-alloc} # which event to trace (cpu, alloc, lock, cache-misses etc.)
+SUFFIX=${SUFFIX:-""}  # file output suffix.
+
+perf_event_paranoid_value=$(cat /proc/sys/kernel/perf_event_paranoid)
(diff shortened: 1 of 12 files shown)


pierDipi:
There we go, profiling data https://github.com/knative-sandbox/eventing-kafka-broker/suites/1603086052/artifacts/29261836 on every PR!

slinkydeveloper (commented):
I just cannot wait to see this in action! Just a bunch of questions:

* Can you reorganize a bit the script to use it locally (setup + trigger sacura and connecting async profiler)?
* Would it makes sense to have this task periodically, more than on each PR?

data-plane/profiler/run.sh
```diff
@@ -0,0 +1,158 @@
...
+export INGRESS_PORT="8080"
+export METRICS_PORT="9089"
+export INSTANCE_ID="dispatcher"
+
+# Run dispatcher.
+java \
+  -XX:+UnlockDiagnosticVMOptions \
+  -XX:+DebugNonSafepoints \
+  -Dlogback.configurationFile="${RESOURCES_DIR}"/config-logging.xml \
+  -jar "${PROJECT_ROOT_DIR}"/dispatcher/target/dispatcher-1.0-SNAPSHOT.jar &
```
slinkydeveloper, line 122: In case you're still wondering the advantages of keeping the data plane independent from the control plane :smile:

pierDipi: Never!

pierDipi:
> Would it makes sense to have this task periodically, more than on each PR?

I think it sometimes makes sense to run it on a PR before merging it to see differences, so I think the ultimate goal is having something to trigger it on a PR, maybe using something like https://github.com/peter-evans/slash-command-dispatch.
The downside is that we need to find it in the Actions tabs because it's not associated with the CI checks.

slinkydeveloper:
> I think the ultimate goal is having something to trigger it on a PR, maybe using something like https://github.com/peter-evans/slash-command-dispatch.

Yeah I would definitely prefer it. We can do it in a followup

> The downside is that we need to find it in the Actions tabs because it's not associated with the CI checks.

Seems like you can add comments with completion? https://github.com/peter-evans/slash-command-dispatch#responding-to-the-comment-on-command-completion

pierDipi:
> Seems like you can add comments with completion? https://github.com/peter-evans/slash-command-dispatch#responding-to-the-comment-on-command-completion

To add a comment, we need a personal access token with specific permissions. 
I tried in this PR to comment on completion (as you can see there are 30 force-pushes), but the default PAT of GitHub is read-only so we cannot comment on a PR. 
That is due to security since a PR is potentially created by an untrusted actor.

slinkydeveloper:
Ok then no problems, we can figure out these details later.

So does the script run locally?

pierDipi:
Yes, it does.

PR #486, opened 2020-12-16
Consistent data plane code style 

pierDipi:
## Proposed Changes

- Add Checkstyle config
- Format (using https://github.com/google/google-java-format) and fix style errors

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/eventbus/ContractMessageCodec.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/eventbus/ContractMessageCodec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018 Knative Authors (knative-dev@googlegroups.com)
+ * Copyright 2020 The Knative Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,26 +22,24 @@ import io.vertx.core.eventbus.MessageCodec;
 
 /**
  * This is a noop codec to send the contract on the event bus (works only in local environments)
- * <p>
- * https://github.com/eclipse-vertx/vert.x/issues/3375
+ *
+ * <p>https://github.com/eclipse-vertx/vert.x/issues/3375
  */
 public final class ContractMessageCodec
-  implements MessageCodec<DataPlaneContract.Contract, DataPlaneContract.Contract> {
+    implements MessageCodec<DataPlaneContract.Contract, DataPlaneContract.Contract> {
 
   @Override
-  public void encodeToWire(Buffer buffer,
-                           DataPlaneContract.Contract contract) {
+  public void encodeToWire(final Buffer buffer, final DataPlaneContract.Contract contract) {
     throw new UnsupportedOperationException();
   }
 
   @Override
-  public DataPlaneContract.Contract decodeFromWire(int i, Buffer buffer) {
+  public DataPlaneContract.Contract decodeFromWire(final int i, final Buffer buffer) {
     throw new UnsupportedOperationException();
   }
 
   @Override
-  public DataPlaneContract.Contract transform(
-    DataPlaneContract.Contract contract) {
+  public DataPlaneContract.Contract transform(final DataPlaneContract.Contract contract) {
     return contract;
   }
 
@@ -55,7 +53,7 @@ public final class ContractMessageCodec
     return -1;
   }
 
-  public static void register(EventBus eventBus) {
+  public static void register(final EventBus eventBus) {
     eventBus.registerDefaultCodec(DataPlaneContract.Contract.class, new ContractMessageCodec());
   }
 }
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2018 Knative Authors (knative-dev@googlegroups.com)
+ * Copyright 2020 The Knative Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,13 +36,14 @@ public class ResourcesReconcilerImpl implements ResourcesReconciler {
   private final IngressReconcilerListener ingressReconcilerListener;
   private final EgressReconcilerListener egressReconcilerListener;
 
-  // Every resource ingress is identified by its resource, so we don't need to store ingress separately
+  // Every resource ingress is identified by its resource, so we don't need to store ingress
+  // separately
   private final Map<String, DataPlaneContract.Resource> cachedResources;
   private final Map<String, DataPlaneContract.Egress> cachedEgresses;
 
   private ResourcesReconcilerImpl(
-    IngressReconcilerListener ingressReconcilerListener,
-    EgressReconcilerListener egressReconcilerListener) {
+      final IngressReconcilerListener ingressReconcilerListener,
+      final EgressReconcilerListener egressReconcilerListener) {
     if (ingressReconcilerListener == null && egressReconcilerListener == null) {
       throw new NullPointerException("You need to specify at least one listener");
     }
@@ -54,13 +55,12 @@ public class ResourcesReconcilerImpl implements ResourcesReconciler {
   }
 
   @Override
-  public Future<Void> reconcile(
-    Collection<DataPlaneContract.Resource> newResources) {
-    final Map<String, DataPlaneContract.Resource> newResourcesMap = new HashMap<>(
-      newResources
-        .stream()
-        .collect(Collectors.toMap(DataPlaneContract.Resource::getUid, Function.identity()))
-    );
+  public Future<Void> reconcile(final Collection<DataPlaneContract.Resource> newResources) {
+    final Map<String, DataPlaneContract.Resource> newResourcesMap =
+        new HashMap<>(
+            newResources.stream()
+                .collect(
+                    Collectors.toMap(DataPlaneContract.Resource::getUid, Function.identity())));
     final List<Future> futures = new ArrayList<>(newResources.size() * 2);
 
     final var resourcesIterator = this.cachedResources.entrySet().iterator();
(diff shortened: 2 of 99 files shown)


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
```diff
@@ -54,13 +55,12 @@ private ResourcesReconcilerImpl(
...
-      newResources
-        .stream()
-        .collect(Collectors.toMap(DataPlaneContract.Resource::getUid, Function.identity()))
-    );
+  public Future<Void> reconcile(final Collection<DataPlaneContract.Resource> newResources) {
+    final Map<String, DataPlaneContract.Resource> newResourcesMap =
+        new HashMap<>(
+            newResources.stream()
+                .collect(
+                    Collectors.toMap(DataPlaneContract.Resource::getUid, Function.identity())));
```
slinkydeveloper, line 63: That style is so bad, with all the parenthesis at the end. I want lines with one close parens indented.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
```diff
@@ -75,67 +75,11 @@ private ResourcesReconcilerImpl(
...
-            futures.add(
-              this.ingressReconcilerListener.onDeleteIngress(oldResource, oldResource.getIngress())
-            );
-          } else if (newResource.hasIngress()) {
-            futures.add(
-              this.ingressReconcilerListener.onUpdateIngress(newResource, newResource.getIngress())
-            );
-          }
-        }
+        reconcileIngress(futures, oldResource, newResource, resourceAreEquals);
```
slinkydeveloper, line 80: Why that change? The linter doesn't like big functions? I don't like that, big functions might have a performance reasoning behind it, i would prefer to remove that lint

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/eventbus/ContractMessageCodec.java
```diff
@@ -22,26 +22,24 @@
 
 /**
  * This is a noop codec to send the contract on the event bus (works only in local environments)
- * <p>
- * https://github.com/eclipse-vertx/vert.x/issues/3375
+ *
+ * <p>https://github.com/eclipse-vertx/vert.x/issues/3375
```
slinkydeveloper, line 26: p should have its own line here

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/ReceiverVerticle.java
```diff
@@ -37,20 +37,22 @@
   private HttpServer server;
   private MessageConsumer<Object> messageConsumer;
   private final Function<Vertx, RequestMapper<String, CloudEvent>> requestHandlerFactory;
-  private final Function<Handler<HttpServerRequest>, Handler<HttpServerRequest>>[] handlerDecoratorFactories;
+  private final Function<Handler<HttpServerRequest>, Handler<HttpServerRequest>>[]
+      handlerDecoratorFactories;
```
slinkydeveloper, line 41: The forced wrap here doesn't make any sense, why should i have the field name on a newline?

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/ReceiverVerticle.java
```diff
@@ -63,34 +65,29 @@ public ReceiverVerticle(
...
   @Override
-  public void stop(Promise<Void> stopPromise) {
-    CompositeFuture.all(
-      server.close().mapEmpty(),
-      messageConsumer.unregister()
-    )
-      .<Void>mapEmpty()
-      .onComplete(stopPromise);
+  public void stop(final Promise<Void> stopPromise) {
+    CompositeFuture.all(server.close().mapEmpty(), messageConsumer.unregister())
```
slinkydeveloper, line 89: Why it didn't kept the `all` indentation?

data-plane/tests/src/test/java/dev/knative/eventing/kafka/broker/tests/DataPlaneTest.java
```diff
@@ -317,18 +335,18 @@ private static ConsumerDeployerVerticle setUpDispatcher(final Vertx vertx, final
...
-
-    final Function<Vertx, RequestMapper<String, CloudEvent>> handlerFactory = v -> new RequestMapper<>(
-      vertx,
-      producerConfigs(),
-      new CloudEventRequestToRecordMapper(v),
-      properties -> KafkaProducer.create(v, properties),
-      mock(Counter.class),
-      mock(Counter.class)
-    );
+  private static ReceiverVerticle setUpReceiver(final Vertx vertx, final VertxTestContext context)
```
slinkydeveloper, line 338: It wrapped the `throws`?! that's bad

slinkydeveloper (commented):
Ok My biggest concern with this code style goes around 2 points:

* The forced wrapping of class, fields and method signatures. Can we remove it and figure out by ourselves when it makes sense or not to wrap?
* The forced closed parenthesis on the same line. This just makes the code more unreadable to me

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
pierDipi: google-java-style format code in this way, no way to change it by design.

pierDipi: There are no other maintained tools like that out there afaik.

data-plane/tests/src/test/java/dev/knative/eventing/kafka/broker/tests/DataPlaneTest.java
pierDipi: google-java-format is design to give a way to change style, bad or good, it's consistent, like in Go

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
pierDipi: Cyclomatic complexity was really high, should I remove it? It seems much more readable now.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/eventbus/ContractMessageCodec.java
pierDipi: google-java-style format code in this way, no way to change it by design.

pierDipi:
> - The forced wrapping of class, fields, and method signatures. Can we remove it and figure out by ourselves when it makes sense or not to wrap?
> - The forced closed parenthesis on the same line. This just makes the code more unreadable to me

This is what we do and it hasn't worked since we have a codebase completely inconsistent.
I prefer to have a style that is consistent.
Also, I don't want to have discussions on a PR around what is a good or a bad style (too much opinionated, that leads to what we have today), I want to have a tool that does this for me, like `gofmt`.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
slinkydeveloper: TBH I personally don't like cyclomatic complexity as a measurement of good code. In this case, I rather prefer to have the full algorithm in a single place, more than cutted in several pieces. This way I personally think it increases the brain overload to grasp the full algorithm, because I must go through all the different indirections. Deciding whether we should create a function or not is a human job, not a tooling job.

Also, as a side note, from the JIT point of view, this code might be optimized in a different way than the original one.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
slinkydeveloper: Oh really? This is really a no go for me.... Can't you just disable the parenthesis formatting?

pierDipi: Nope, we can't.

slinkydeveloper:
> Also, I'd like to avoid having discussions on PRs around what is a good or a bad style (too much opinionated, that leads to what we have today), I want to have a tool that does this for me, like gofmt.

Java is by design more verbose and less opinionated than Golang in terms of syntax and semantics, hence the hope to use a very opinionated tool like `gofmt` doesn't bring any good results. As you see these tool you tried made some very "opinionated choices" which doesn't make sense in practice: the forced wraps makes sense for the sake of consistency, but practically they make the code more unreadable. 

I'm ok with trying to make the codebase a bit more consistent, but this is not Golang, I think there is a certain degree of liberty that has to be left to the developer to make the code more reasonable to him. For me indenting the code this way:

```java
someOutput = list.stream()
  .map(someLambda).filter(Objects::nonNull)
  .collect(Collectors.toList())
```

Or this way:

```java
someOutput = list.stream()
  .map(someLambda)
  .filter(Objects::nonNull)
  .collect(Collectors.toList())
```

Makes me reason about the code in different ways. In the first case, my brain sees filter as part of the same logical step of map, it doesn't give to it much importance, while in the other case I see filter as a separate step, which deserves more attention. The tooling doesn't care about that and wraps anyway. Same exact discussion goes for wraps of fields (which is absolutely unreasonable IMO), signatures and, very important IMO, parens.

And, last but not least, the reviewers should be the people who ultimately care about the inconsistencies in code style. 

Long story short: I'm ok with adding a tool to format the codebase, but that tool should not decide the way I want to show the code to a reader.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
matzew:
```
                    Collectors.toMap(DataPlaneContract.Resource::getUid, Function.identity())));
```
is ugly to read - no?

pierDipi:
No way to make everybody happy, I tried clang-format, it has similar problems, closing.

PR #511, opened 2020-12-28
Skip ConfigMap update when there are no changes

pierDipi:
During E2E tests there are a lot of no-op reconciliation
since chaosduck kills the leader pod and every time
a new replica starts we trigger a global resync that 
most of the time doesn't actually change the ConfigMap.

For example, the ConfigMap diff looks like this:

```
  	"""
  	{
- 	  "generation": "1",
+ 	  "generation": "2",
  	  "resources": [
  	    {
  	... // 15 identical lines
  	"""
  )
```

This leads to an unnecessary reconciliation loop on the data plane
and increase update conflicts of the ConfigMap.

## Proposed Changes

- Skip ConfigMap update when there are no changes

**Release Note**

```release-note
:broom: Update or clean up current behavior
Skip ConfigMap update when there are no changes
```

/kind cleanup

### DIFF
--- a/control-plane/pkg/core/config/egress.go
+++ b/control-plane/pkg/core/config/egress.go
@@ -17,7 +17,9 @@
 package config
 
 import (
+	"google.golang.org/protobuf/proto"
 	"k8s.io/apimachinery/pkg/types"
+
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/contract"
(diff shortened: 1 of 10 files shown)


slinkydeveloper:
/lgtm
/approve

Sounds like a nice improvement!

Issue #596, opened 2021-02-04
Speed up integration tests pipeline


pierDipi:
We should speed up our integration tests pipeline since it takes ~1h to complete.

Some ideas could be:
- Parallel test images upload and data plane or control plane build (or both)
- Parallel Kafka cluster deploy and data plane or control plane build (or both)

## Expected Behavior

integration tests pipeline tasks << 1h

## Actual Behavior

integration tests pipeline takes ~1h

## Steps to Reproduce the Problem

1. Just run integration tests

## Additional Info

- Increasing parallelism had been tried in other repositories and there are no advantages.
- The main pipeline flow is in https://github.com/knative-sandbox/eventing-kafka-broker/blob/master/test/e2e-common.sh
- Ideally, we don't introduce too many bash scripts or functions

/area test
/kind performance

pierDipi:
We're at 45 min.

---

> Parallel test images upload and data plane or control plane build (or both)
> Parallel Kafka cluster deploy and data plane or control plane build (or both)

Those don't help that much and they add complexity. (verified in https://github.com/knative-sandbox/eventing-kafka-broker/pull/607)

slinkydeveloper:
Can this be considered done? Any other ideas you wanna try? :)

pierDipi:
I see serving is using a very high parallel number:
https://github.com/knative/serving/blob/0751aed142e00726647a2042fcc1f587a9a7827a/test/e2e-tests.sh#L136

let's try this.

pierDipi:
Let's close this, we've got 15m improvements, which is good for now.

PR #607, opened 2021-02-08
Execute deploy and image publishing in parallel

pierDipi:
Part of #596

## Proposed Changes

- Execute parallel deploy and image publishing

/cc pierDipi


pierDipi:
The setup is 12m faster.

pierDipi:
No real gain here, closing.

PR #608, opened 2021-02-08
Avoid building receiver twice

pierDipi:
We're building the receiver image twice.

Part of #596

## Proposed Changes

- Avoid building receiver twice


slinkydeveloper:
/lgtm
/approve

too much shell :(

PR #609, opened 2021-02-08
Execute sacura job in parallel

pierDipi:
Part of #596

## Proposed Changes

- Execute sacura job in parallel

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -39,7 +39,11 @@ export_logs_continuously "kafka-broker-dispatcher" "kafka-broker-receiver" "kafk
 
 failed=false
 
-go_test_e2e -timeout=50m ./test/... || failed=true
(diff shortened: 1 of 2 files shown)


test/e2e-tests.sh
```diff
@@ -40,15 +39,10 @@ export_logs_continuously "kafka-broker-dispatcher" "kafka-broker-receiver" "kafk
 
 failed=false
 
-go_test_e2e -timeout=30m ./test/... || failed=true
+go_test_e2e -timeout=50m ./test/... || failed=true
```
slinkydeveloper, line 42: Uhhhhh are you sure of that?

pierDipi: I changed again to use a 2 step test since GH actions don't run the sacura job and reverted this timeout change.

PR #613, opened 2021-02-09
Reduce sleep time in GH actions test setup

pierDipi:
As per title.

Part of #596


slinkydeveloper:
/lgtm
/approve

Sounds like it's working :smile:

PR #631, opened 2021-02-12
Improve SSL support

### LINKED ISSUE #605: SASL/PLAIN does not work with public TLS certificate
**Describe the bug**
- https://github.com/knative-sandbox/eventing-kafka/issues/356
- https://github.com/knative-sandbox/eventing-kafka/pull/359

pierDipi:
When `protocol=SSL` or `protocol=SASL_SSL` CA certificates aren't
required since users might want to use system's root CA set.

When `protocol=SSL` client auth might or might not be required
on the broker side. 
Users can use `user.skip: true` to disable client authentication
checks.

Note: I thought to use `user.skip` since we use `user.crt` and
`user.key` for providing client key pair.

Fixes #605

## Proposed Changes

- Improve SSL support

**Docs**

- https://github.com/knative/docs/issues/3202

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/security/Credentials.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/security/Credentials.java
@@ -30,7 +30,7 @@ public interface Credentials {
    *
    * @return true if client auth should be skipped otherwise false
    */
-  boolean skipUser();
+  boolean skipClientAuth();
 
   /**
    * @return user certificate.
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/security/Credentials.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/security/Credentials.java
@@ -26,6 +26,13 @@ public interface Credentials {
   String caCertificates();
 
   /**
+   * Skip client auth.
+   *
+   * @return true if client auth should be skipped otherwise false
+   */
+  boolean skipUser();
+
+  /**
    * @return user certificate.
    */
   String userCertificate();
--- a/control-plane/pkg/security/secret.go
+++ b/control-plane/pkg/security/secret.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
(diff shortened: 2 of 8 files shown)


slinkydeveloper:
cc @matzew

matzew:
So, with `protocol=SASL_SSL` - I now will use `user.skip`, to skip the CA certs etc, since they are available via the base image?

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/security/Credentials.java
```diff
@@ -25,6 +25,13 @@
    */
   String caCertificates();
 
+  /**
+   * Skip client auth.
+   *
+   * @return true if client auth should be skipped otherwise false
+   */
+  boolean skipUser();
```
matzew, line 33: Not sure the name is really correct ... :thinking:

pierDipi:
> So, with protocol=SASL_SSL - I now will use user.skip, to skip the CA certs etc, since they are available via the base image?

`user.skip` is only used when `protocol=SSL` to disable/enable client authentication by means of client's key pair (`user.crt` and `user.key`)

---

For `protocol=SASL_SSL` or `protocol=SSL` to use CA cert of the base image you can omit `ca.crt` in the secret.

---

There is documentation on the website: https://knative.dev/development/eventing/broker/kafka-broker/#security

pierDipi:
I guess I should haven't mixed 2 problems in one PR, `user.skip` is not related to #605

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/security/Credentials.java
pierDipi: `skipClientAuth`?

pierDipi:
https://github.com/knative/pkg/issues/1509
```
Failed to get trigger "trigger-with-broker": conversion webhook for eventing.knative.dev/v1, Kind=Trigger failed: Post https://eventing-webhook.knative-eventing.svc:443/resource-conversion?timeout=30s: EOF
```
/retest

slinkydeveloper (approved):

PR #633, opened 2021-02-12
Add tests for previous versions of Eventing

pierDipi:
This PR starts running E2E tests using previous versions of Eventing.

## Proposed Changes

- Add tests for previous versions of eventing

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -26,6 +26,12 @@ jobs:
           - v1.19.4
           - v1.20.0
 
+        eventing-config:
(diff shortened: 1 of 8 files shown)


slinkydeveloper:
I would love to wait for it and merge it after the release, since it might introduce additional CI instability

/hold

pierDipi:
This runs in GH Actions only, the problem we have is only with `sacura`, anyway no strong opinion.

slinkydeveloper:
I know it runs on gh actions, but i don't want to let the repo looks like unstable

pierDipi:
Can we merge this now?

slinkydeveloper (approved):

PR #651, opened 2021-02-17
Make sure we can switch logger whenever we want

### LINKED ISSUE #641: Figure out how to get rid of keyValue for logging
**Describe the bug**
Today we have `keyValue` scattered in different log statements, which are tied to the logging implementation

 **Expected behavior**
We should not be tied to the logging implementation, hence we should remove every statement that couple our code to logstash logback json encoder.

slinkydeveloper:

Fixes #641

## Proposed Changes

- Just some funky classloader hack to make sure keyValue still works without logstash json encoder in classpath


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/utils/Logging.java
```diff
@@ -0,0 +1,24 @@
+package dev.knative.eventing.kafka.broker.core.utils;
```
pierDipi, line 1: Missing header.

data-plane/core/src/test/java/dev/knative/eventing/kafka/broker/core/utils/LoggingTest.java
```diff
@@ -0,0 +1,34 @@
+package dev.knative.eventing.kafka.broker.core.utils;
```
pierDipi, line 1: Missing header.

slinkydeveloper:
@pierDipi do you like the classpath hack? :smile:

pierDipi:
Yes, I like it! 
/lgtm

PR #657, opened 2021-02-18
Control protocol Java implementation

slinkydeveloper:

Draft of Control protocol implementation for Java using Vert.x

Note: I decided on purpose to not let the control protocol package depend on core, in order to allow in future to extract it from eventing-kafka-broker, if necessary


devguyio:
I'd like to take a stab reviewing this
/assign

slinkydeveloper:
@pierDipi this should be ready to rock! I would love to expand it to create a nicer interface to use, but first I want to get this "foundational work" in. Can you check it out?

data-plane/control-protocol/src/main/java/dev/knative/eventing/control/protocol/impl/TCPControlServerVerticle.java
```diff
@@ -0,0 +1,183 @@
...
+      logger.warn("Received an ack for an unknown uuid {}", ackMessage.uuid());
+    } else {
+      message.reply(ackMessage);
+    }
+  }
+
+  private void emitMessageOnEventBus(ControlMessage message) {
+    this.serializedEventBusRequester.request(message)
+      .onFailure(t -> logger
+        .error("Cannot route the incoming control message to {}: {}", this.incomingMessageAddress, message, t))
```
pierDipi, line 134: We should revisit whether logging a message at this level is safe and we don't leak sensitive data, a TODO/issue?

data-plane/control-protocol/src/main/java/dev/knative/eventing/control/protocol/impl/TCPControlServerVerticle.java
```diff
@@ -0,0 +1,183 @@
...
+      return;
+    }
+    this.internalWriteOnConnection(message, message.toBuffer());
+  }
+
+  private void internalWriteOnConnection(ControlMessage message, Buffer buffer) {
+    actualConnection.write(buffer)
+      .onSuccess(v -> logger.debug("Successfully sent {}", message))
+      .onFailure(t -> {
+        logger.error("Cannot dispatch message {}", message, t);
```
pierDipi, line 178: Same here.

data-plane/control-protocol/src/main/java/dev/knative/eventing/control/protocol/impl/TCPControlServerVerticle.java
```diff
@@ -0,0 +1,183 @@
...
+        // TODO should we retry here?
```
pierDipi, line 179: I think we should have "a business level" retry policy, maybe some messages are important others might be less

pierDipi: and at this layer, we don't have enough context. :thinking:

pierDipi: I'm specifically thinking at stuff like:

"I cannot run a producer because xyz" vs "I cannot apply a retry policy update"

data-plane/config/broker/template/500-dispatcher.yaml
```diff
@@ -97,8 +97,6 @@ spec:
             # This should be set according to initial delay seconds
             - name: WAIT_STARTUP_SECONDS
               value: "8"
-          command:
-            - "java"
```
pierDipi, line 101: See the question on the docker file changes.

data-plane/config/broker/template/500-receiver.yaml
```diff
@@ -108,8 +108,6 @@ spec:
             # This should be set according to initial delay seconds
             - name: WAIT_STARTUP_SECONDS
               value: "8"
-          command:
-            - "java"
```
pierDipi, line 112: Same here.

data-plane/config/sink/template/500-receiver.yaml
```diff
@@ -108,8 +108,6 @@ spec:
             # This should be set according to initial delay seconds
             - name: WAIT_STARTUP_SECONDS
               value: "8"
-          command:
-            - "java"
```
pierDipi, line 112: Same here.

pierDipi (commented):
I think this is a nice addition.

Most of my comments are questions and follow-ups.

The 2 big things we need to do/discuss are:
 - TLS and certs rotation
 - Multiple replicas for both control and data plane

data-plane/docker/Dockerfile
```diff
@@ -55,3 +57,6 @@ COPY --from=builder /app /app
 
 # Configure PATH
 ENV PATH="/app/jdk/bin:${PATH}"
+
+ENTRYPOINT ["java"]
+CMD ["-jar", "/app/app.jar"]
```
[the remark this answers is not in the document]
slinkydeveloper: I know, but this is going to be overridden by the deployment yaml, so it won't affect production configs. Check this: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes

Without modifying the dockerfile, the conformance test doesn't run "trivially" :smile:

data-plane/control-protocol/src/main/java/dev/knative/eventing/control/protocol/impl/TCPControlServerVerticle.java
slinkydeveloper: I can address it on this PR. I think the problem is that the `ControlMessageImpl#toString()` should not print the payload of the message, that's the only place where sensitive data can exist.

slinkydeveloper:
Just to make things a bit clearer, this PR was meant just as the beginning, while working on it I've identified the next steps, in particular:

* An interface like the Golang one `Service` to simplify the usage and abstract from the tcp protocol itself. The interface you see now in the conformance test runner is definitely tooo low level.
* Figure out the whole mTLS story
* Understand retries behavior

After we merge this one, I'll try to draw a plan for the next steps, including how to start adopting it in "production", that is, inside receiver and dispatcher.

> Multiple replicas for both control and data plane

This is out of the actual design of the whole control protocol story. I still don't see how and why we should need this TBH, but I definitely look forward to discuss it. Can we move this particular topic on an issue on https://github.com/knative-sandbox/control-protocol?

slinkydeveloper:
FYI https://github.com/knative-sandbox/eventing-kafka-broker/issues/853

pierDipi (commented):
> This is out of the actual design of the whole control protocol story. I still don't see how and why we should need this TBH, but I definitely look forward to discuss it. Can we move this particular topic on an issue on https://github.com/knative-sandbox/control-protocol?

From [config-leader-election](https://github.com/knative/serving/blob/7a9e98b952f6f5d1b270599dd438c2d669cca7ce/config/core/configmaps/leader-election.yaml#L53):
```
    # buckets is the number of buckets used to partition key space of each
    # Reconciler. If this number is M and the replica number of the controller
    # is N, the N replicas will compete for the M buckets. The owner of a
    # bucket will take care of the reconciling for the keys partitioned into
    # that bucket.
    buckets: "1"
```
therefore at any given time, there could be multiple controller pods that are leaders (of different buckets).

I'd like to understand how this works with the control protocol given that I see that we keep only a single connection.

slinkydeveloper:
> therefore at any given time, there could be multiple controller pods that are leaders (of different buckets).

From the sound of it, that seems very broken and racy, and also practically useless. I don't get how this could work, because several controllers will compete to reconcile the resource and write it on the api server... But again, let's open an issue in the control-protocol repo and discuss there.

data-plane/control-protocol/src/main/java/dev/knative/eventing/control/protocol/impl/SerializedEventBusRequester.java
```diff
@@ -0,0 +1,76 @@
...
+import io.vertx.core.Promise;
+import io.vertx.core.eventbus.DeliveryOptions;
+import io.vertx.core.eventbus.EventBus;
+import io.vertx.core.eventbus.Message;
+import java.util.ArrayDeque;
+import java.util.Map;
+import java.util.Queue;
+
+/**
+ * This class sends requests to event bus in order, waiting every time for the previous one to reply before sending the next one.
```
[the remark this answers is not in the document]
slinkydeveloper: The reason to have this logic is to "stay on the safe side" to begin experimenting. Control messages should take very very short amount of time to be handled.

I foresee how, in case we start having scalability problems and this becomes a significant bottleneck, we can have "smarter" logic to serialize requests, e.g. like you sad, reconcile requests for trigger B could run in parallel with reconcile requests for trigger A, but two requests for trigger A should be serialized. But I want this to become a problem before tackling it, and that's the reason why I made a separate class for it, more than including this logic straight in the `TCPControlServerVerticle`.

pierDipi:
Design document https://docs.google.com/document/d/1i_QHjQO2T3SNv49xjZLWlivcc0UvZN1Tbw2NKxThkyM/edit?usp=sharing

slinkydeveloper:
@pierDipi I think you're mistaken here.

> therefore at any given time, there could be multiple controller pods that are leaders (of different buckets).

This means that at the same time, only one controller has the ownership of one bucket (and its contained resources). In other words, only one reconciler owns a specific CR, controlling it. Trigger x.y.z cannot be at the same time handled by reconciler replica 1 and replica 2.

This means that the assumption made by the control protocol is correct.

pierDipi:
We're saying the same thing, but you're assuming that a given data plane pod handles only resources of a single bucket.

Let's say we have10 buckets, 2 controllers and 1 dispatcher. 
Controller 1 handles bucket [0, 4] and controller 2 handles bucket [5, 9]

I create a trigger (trigger1) that ends up being in the bucket 1 and another one (trigger2) that ends up being in the bucket 9, both controllers should schedule trigger1 and trigger2 in the same dispatcher pod, therefore they needs to talk to the same data plane pod.

slinkydeveloper:
> but you're assuming that a given data plane pod handles only resources of a single bucket.

Ah ah then you're right! Sorry, I misunderstood your point completely! Yep that's definitely important, this needs to work.

This might not be an easy fix, I need to reiterate on it in control-protocol repo first, I've opened an issue there: https://github.com/knative-sandbox/control-protocol/issues/66.

/hold

slinkydeveloper:
/unhold

I've added in this PR the broadcast implementation, as explained in https://github.com/knative-sandbox/control-protocol/issues/66

data-plane/docker/Dockerfile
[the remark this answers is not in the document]
pierDipi: TIL

data-plane/control-protocol/src/main/java/dev/knative/eventing/control/protocol/impl/TCPControlServerVerticle.java
pierDipi: Yes, sounds good to me let's thing to this problem later, I was just sharing some random thoughts.

slinkydeveloper:
I've removed the broadcast implementation, for more details https://github.com/knative-sandbox/control-protocol/issues/66

matzew:
I think we can close this?

/cc @pierDipi ?

pierDipi:
Yes, we can close this.

PR #667, opened 2021-02-20
Add verify committed offset test

pierDipi:
Verify that after sacura run, the consumer group lag
`Lag = Sum(offset to produce next - committed offset)`
is 0.

While I was creating the test, I've seen an opportunity to
create a reusable (and testable in isolation) library to
calculate consumer group lags.

## Proposed Changes

- Add verify committed offset test

### DIFF
--- a/control-plane/pkg/kafka/consumer_group_lag.go
+++ b/control-plane/pkg/kafka/consumer_group_lag.go
@@ -218,7 +218,7 @@ func (cgl ConsumerGroupLag) String() string {
 	return sb.String()
 }
 
-// Lag returns LatestOffset - ConsumerOffset
+// Lag returns LatestOffset - ConsumerOffset.
 func (pl PartitionLag) Lag() int64 {
 	return pl.LatestOffset - pl.ConsumerOffset
 }
--- a/control-plane/pkg/kafka/consumer_group_lag.go
+++ b/control-plane/pkg/kafka/consumer_group_lag.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/Shopify/sarama"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 const (
@@ -54,13 +55,25 @@ type PartitionLag struct {
 type adminFunc func(client sarama.Client) (sarama.ClusterAdmin, error)
 
 type consumerGroupLagProvider struct {
-	client    sarama.Client
-	adminFunc adminFunc
+	client         sarama.Client
+	adminFunc      adminFunc
+	offsetStrategy int64
 }
 
+var (
+	offsetStrategies = sets.NewInt64(sarama.OffsetNewest, sarama.OffsetOldest)
+)
+
+const (
+	defaultOffsetStrategy = sarama.OffsetOldest
+)
+
 // NewConsumerGroupLagProvider creates a new ConsumerGroupLagProvider.
-func NewConsumerGroupLagProvider(client sarama.Client, adminFunc adminFunc) ConsumerGroupLagProvider {
-	return &consumerGroupLagProvider{client: client, adminFunc: adminFunc}
+func NewConsumerGroupLagProvider(client sarama.Client, adminFunc adminFunc, saramaOffsetStrategy int64) ConsumerGroupLagProvider {
+	if !offsetStrategies.Has(saramaOffsetStrategy) {
+		saramaOffsetStrategy = defaultOffsetStrategy
+	}
+	return &consumerGroupLagProvider{client: client, adminFunc: adminFunc, offsetStrategy: saramaOffsetStrategy}
 }
 
 // GetLag returns consumer group lag for a given group.
@@ -133,12 +146,19 @@ func (p *consumerGroupLagProvider) getPartitionLag(partition int32, topic string
 
 	if consumerOffset <= invalidOffset {
 		// When we receive an invalid consumer offset, it means no offset has yet been committed.
-		// So, set consumer offset to the offset that will be produced next.
-		//
-		// Note: setting it to 0 isn't a viable option since depending on the initial offset strategy
-		// (sarama.OffsetNewest or sarama.OffsetOldest) the offset that will be consumed next might be
-		// any valid offset and, in this case, it's exactly the offset that will be produced next.
-		consumerOffset = latestOffset
+
+		if p.offsetStrategy == sarama.OffsetOldest {
+			// Set consumer offset to the first offset.
+			consumerOffset = 0
+		} else if p.offsetStrategy == sarama.OffsetNewest {
+			// Set consumer offset to the offset that will be produced next.
+			consumerOffset = latestOffset
+		} else {
+			return PartitionLag{}, fmt.Errorf(
+				"received invalid offset (%d) and unknown offset strategy (%d) configured",
+				consumerOffset, p.offsetStrategy,
+			)
+		}
 	}
 
 	pl := PartitionLag{
@@ -194,7 +214,7 @@ func (cgl ConsumerGroupLag) String() string {
 	writeSeparator(n, sb)
 
 	for p, l := range cgl.ByPartition {
-		sb.WriteString(fmt.Sprintf("%d: %d\n", p, l))
+		sb.WriteString(fmt.Sprintf("%d: %s\n", p, l))
 	}
 	sb.WriteString(fmt.Sprintf("Total lag: %d\n", cgl.Total()))
 	return sb.String()
--- a/control-plane/pkg/kafka/consumer_group_lag.go
+++ b/control-plane/pkg/kafka/consumer_group_lag.go
@@ -17,6 +17,7 @@
 package kafka
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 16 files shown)


pierDipi:
Maven
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.2.4:shade (default) on project receiver: Execution default of goal org.apache.maven.plugins:maven-shade-plugin:3.2.4:shade failed: Plugin org.apache.maven.plugins:maven-shade-plugin:3.2.4 or one of its dependencies could not be resolved: Could not transfer artifact org.ow2.asm:asm:jar:8.0 from/to central (https://repo.maven.apache.org/maven2): Transfer failed for https://repo.maven.apache.org/maven2/org/ow2/asm/asm/8.0/asm-8.0.jar: Connection reset -> [Help 1]
```
/test pull-knative-sandbox-eventing-kafka-broker-unit-tests

slinkydeveloper:
What's the lag?

pierDipi:
`Lag = Sum(offset to produce next - committed offset)`

control-plane/pkg/kafka/consumer_group_lag.go
```diff
@@ -0,0 +1,225 @@
...
+
+// ConsumerGroupLag contains partition lag of a topic.
+type ConsumerGroupLag struct {
+	Topic         string
+	ConsumerGroup string
+	ByPartition   []PartitionLag
+}
+
+// PartitionLag contains consumer lag information of a partition.
+type PartitionLag struct {
```
slinkydeveloper, line 49: Maybe rename `PartitionOffsetStatus`?

control-plane/pkg/kafka/consumer_group_lag.go
```diff
@@ -0,0 +1,225 @@
...
+	LatestOffset   int64 // Offset that will be produced next.
+	ConsumerOffset int64 // Offset that will be consumed next.
+	Lag            int64 // LatestOffset - ConsumerOffset
```
slinkydeveloper, line 52: Don't store it, it's more clear if it's not store IMO

test/e2e/consumer_group_lag_provider_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/types"
+	testlib "knative.dev/eventing/test/lib"
+
+	kafkatest "knative.dev/eventing-kafka-broker/test/pkg/kafka"
+	pkgtesting "knative.dev/eventing-kafka-broker/test/pkg/testing"
+)
+
+func TestConsumerGroupLagProvider(t *testing.T) {
```
slinkydeveloper, line 33: Can you use the new test framework for this e2e test? I would love to avoid adding new tests using the old test code, if possible

slinkydeveloper (commented):
TBH the lag name just doesn't sound right to me, except that I'm ok with this PR and thanks for tackling it!

control-plane/pkg/kafka/consumer_group_lag.go
pierDipi: It's pretty common in the Kafka world `Lag`, in fact, this command:
```./bin/kafka-consumer-groups.sh --bootstrap-server localhost:9092 --group mygroup  --describe```
produces an output like this:
```
TOPIC                          PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG        CONSUMER-ID                                       HOST
```

File: control-plane/pkg/kafka/consumer_group_lag.go (hunk 2 above)
Comment (pierDipi) [reply control-plane/pkg/kafka/consumer_group_lag.go:52]: Sure!

File: control-plane/pkg/kafka/consumer_group_lag.go (hunk 1 above)
Comment (pierDipi) [reply control-plane/pkg/kafka/consumer_group_lag.go:49]: Lag in the output is exactly `PartitionLag` in our code.

File: test/e2e/consumer_group_lag_provider_test.go (hunk 3 above)
Comment (pierDipi) [reply test/e2e/consumer_group_lag_provider_test.go:33]: Great suggestion! Let me check what I can do with it.

File: control-plane/pkg/kafka/consumer_group_lag.go (hunk 1 above)
Comment (slinkydeveloper) [reply control-plane/pkg/kafka/consumer_group_lag.go:49]: Ok I wasn't aware of that, i'm sorry, then lag it is

Comment (pierDipi) [reply control-plane/pkg/kafka/consumer_group_lag.go:49]: No problem!

File: test/e2e/consumer_group_lag_provider_test.go (hunk 3 above)
Comment (pierDipi) [reply test/e2e/consumer_group_lag_provider_test.go:33]: I tried some stuff, but this requires some other changes to some of our helpers, that don't belong to this PR.
Also, I'm a little worried about the flakiness we've seen with the new test framework in eventing.
With that being said, I'd like to wait for this migration for now.

Comment (slinkydeveloper) [reply test/e2e/consumer_group_lag_provider_test.go:33]: Ok, then we'll do that in another PR

**Comment by slinkydeveloper** [comment]
Can you fix the lint?

**Comment by pierDipi** [comment]
Sure.

/hold

File: control-plane/pkg/kafka/consumer_group_lag.go [hunk 4]
```diff
@@ -0,0 +1,228 @@
...
+	latestOffset, err := p.client.GetOffset(topic, partition, sarama.OffsetNewest)
+	if err != nil {
+		return PartitionLag{}, fmt.Errorf("failed to find latest offset for topic %s and partition %d", topic, partition)
+	}
+	latestOffset = max(0, latestOffset) // latest offset should always be greater or equal to 0.
+
+	if consumerOffset <= invalidOffset {
+		// When we receive an invalid consumer offset, it means no offset has yet been committed.
+
+		if p.offsetStrategy == sarama.OffsetOldest {
```
slinkydeveloper, line 149: This offset strategy thing is weird... I don't see why I should provide it to default some value. If no offset is committed, the lag is just 0 and the consumerOffset is -1, right?

slinkydeveloper:
More than having more test images to run some specific asserts on kafka broker, can we run them in the test directly accessing to the kafka cluster?

Exposing the kafka cluster should be straightforward with strimzi: https://strimzi.io/docs/operators/latest/using.html#assembly-configuring-external-listeners-str just use a nodeport svc?

control-plane/pkg/kafka/consumer_group_lag.go
pierDipi: No committed offset doesn't mean that the lag is 0 because the producer might have produced messages and depending on the strategy it could be 0 (OffsetNewest) or N (OffsetOldest).

slinkydeveloper: Then can we just represent the state "no offset committed"? Otherwise, this information get lost, which might be useful otherwise (e.g. I'm thinking to unordered delivery tests)

pierDipi: I think propagating "no offset committed" to the caller is a good idea but I feel the logic to determine the consumer offset naturally belongs to this function.

I'd add a field called `OffsetCommitted` to determine whether an offset has been committed.

WDYT?

slinkydeveloper: Ok

pierDipi: Done

pierDipi:
> Exposing the kafka cluster should be straightforward with strimzi: https://strimzi.io/docs/operators/latest/using.html#assembly-configuring-external-listeners-str just use a nodeport svc and then get the url retrieving the svc?

NodePort svcs require special network setups in some environments (firewall rules, etc) to be accessed from the outside, I'd prefer to keep these checks in Jobs running inside the cluster.

slinkydeveloper:
> NodePort svcs require special network setups in some environments (firewall rules, etc) to be accessed from the outside, I'd prefer to keep these checks in Jobs running inside the cluster.

IMO I think this is worth the effort: there are other tests I would love to develop that directly connects to kafka, so we should at least give it a try... I'll try to draft something and I'll let you know.

test/test_images/committed-offset/admin.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	consumerGroupLagProvider := kafka.NewConsumerGroupLagProvider(client, sarama.NewClusterAdminFromClient, sarama.OffsetOldest)
+	defer consumerGroupLagProvider.Close()
+
+	lag, err := consumerGroupLagProvider.GetLag(envConfig.Topic, envConfig.Group)
+	if err != nil {
+		log.Fatal("Failed to get lag", err)
+	}
+
+	if lag.Total() != 0 {
```
slinkydeveloper, line 56: can you print the lag data structure as json, parse it in the test runner and then assert there? I prefer to use these test images only as "data collectors", so we can more flexibly perform assertions and debug in the test runner.

test/test_images/consumer-group-lag-provider-test/admin.go
```diff
@@ -0,0 +1,212 @@
...
+	"time"
+
+	"github.com/Shopify/sarama"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"knative.dev/eventing-kafka-broker/control-plane/pkg/kafka"
+	testingpkg "knative.dev/eventing-kafka-broker/test/pkg/testing"
+)
+
+func main() {
```
slinkydeveloper, line 31: Is this thing testing your test code? If that's the case, do we really need it? Your code to check the consumer group lag is quite easy, unit test should be enough i guess?

slinkydeveloper (commented):
I tried to go down the path to expose kafka, it's way too complicated and makes worse the dev experience (which is already quite convoluted today)

So let's go down the test_images road. I've left a comment about the test_images approach, in order to make a bit easier debugging etc

test/test_images/committed-offset/admin.go
pierDipi: Sounds interesting but parsing logs seems less robust.

I'm not sure how we use k8s' events for this kind of communication but maybe we can use a similar pattern?

test/test_images/consumer-group-lag-provider-test/admin.go
pierDipi: I prefer to keep this test, I discovered a bug with this test :upside_down_face: which is a signal that it's useful.
I was debugging the data plane but the bug was actually in the test code.

slinkydeveloper: I see... TBH, an integration test for the test code is weird :smile: Maybe can you just comment that this is useful for testing the test code and maybe disable it in our test pipeline?

test/test_images/committed-offset/admin.go
slinkydeveloper: > I'm not sure how we use k8s' events for this kind of communication but maybe we can use a similar pattern?

Way too complicated, I suggest we avoid going down that road. WDYM by parsing logs is less robust?

test/test_images/consumer-group-lag-provider-test/admin.go
pierDipi: Disabled.

test/test_images/committed-offset/admin.go
pierDipi: Something like
```
	log.Printf(`{"lag": %d}`, lag.Total())
```

and then in the test should I try to parse every log line looking for the above Json? 

Is this what you mean? It seems weird. :thinking:

slinkydeveloper: How about just parse the last log line? Is that reasonable?

slinkydeveloper (approved):
@pierDipi can we merge this one? If yes, remove the hold.

/lgtm

Issue #589, opened 2021-02-04
Ordered consumer


slinkydeveloper:
**Problem**
We need ordered message consumptions to cover a range of use cases

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for? User

**Exit Criteria**
An option to setup ordered consumption of topics

**Steps**

* [x] Refactor the Consumer verticle and friends to make room for ordered consumer #614 #628
* [x] Implement ordered consumer verticle #639
* [x] Add to the contract the ordered/unordered for the egress #689
* [x] Define an API to expose ordered/unordered consumer and implement it #701
* [x] Docs https://github.com/knative/docs/pull/3369
* [x] Add DeliveryOrder status condition or annotation for Trigger #732
* [x] Add a test with more partitions using the partition key to check the ordered delivery across partitions #756

slinkydeveloper:
In order to develop an ordered consumer, we need a FSM that polls the consumer and process the events. More details to come

pierDipi:
_Note: this is not part of this issue, just sharing an idea._

What do you think about having 2 topics?

The reasoning being that when there is no `partitionkey` the order is "random" from a user perspective and they don't care about ordering, so we can have 2 topics:

1. 1 topic for strictly order delivery
1. 1 topic for unordered delivery

The receiver sends events to (1) when there is a `partitionkey` set and to (2) when there is not.
The dispatcher can: 
1. instantiate a consumer for (1) respecting strictly ordered delivery and a consumer for (2) with a parallel consumer. Or
1. instantiate a consumer for (1) and (2) and route events to a component based on the topic.

slinkydeveloper:
TBH I'm not sure the effort is worth and ends up in a reasonable speedup. In any case, I'm going to implement the ordered consumer without removing the unordered consumer, so this will be an option to turn it on (then on next release, we make it default)

pierDipi:
We can consider this one: https://github.com/confluentinc/parallel-consumer

pierDipi:
There is a Vert.x integration too.

pierDipi:
> Why not use the Vert.x library yourself in your processing loop?
> 
> - Vert.x us used in this library to provide a non-blocking IO system in the message processing step. Using Vert.x without using this library with ordered processing requires dealing with the quite complicated, and not straight forward, aspect of handling offset commits with Vert.x asynchronous processing system.
> 
[...]

PR #689, opened 2021-03-02
Added delivery guarantee to the egress contract

slinkydeveloper:

Part of #589

## Proposed Changes

- Added DeliveryGuarantee enum to the egress contract and defaulting to unordered


slinkydeveloper:
Not sure how should i fix the copyright, it seems like the old version doesn't have the copyright header too

slinkydeveloper:
/retest

Pass on kind, fails on prow

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/consumer/DeliveryGuarantee.java
```diff
@@ -23,5 +25,15 @@
...
+
+  public static DeliveryGuarantee fromContract(DataPlaneContract.DeliveryGuarantee deliveryGuarantee) {
+    if (deliveryGuarantee == null) {
+      return UNORDERED;
+    }
+    return switch (deliveryGuarantee) {
+      case ORDERED -> ORDERED;
+      case UNORDERED, UNRECOGNIZED -> UNORDERED;
+    };
+  }
```
pierDipi, line 38: Can we have a UT?

proto/def/contract.proto
```diff
@@ -43,6 +43,12 @@ message EgressConfig {
   uint64 backoffDelay = 4;
 }
 
+// Check dev.knative.eventing.kafka.broker.dispatcher.consumer.DeliveryGuarantee for more details
+enum DeliveryGuarantee {
+  ORDERED = 0;
+  UNORDERED = 1;
+}
```
pierDipi, line 50: Since we will probably add `at least once` / `at most once`, can you rename this enum to something less ambiguous?

slinkydeveloper: Why should we have at most once? Don't we always support at least once and that's it? Do you plan to have at most once?

slinkydeveloper:
@pierDipi any ideas about the failures?

pierDipi:
Do you mind the linter? 
The linter fails when we change generated code.

slinkydeveloper:
Then i guess this one is ready to go in?

proto/def/contract.proto
pierDipi: Yes, see https://github.com/knative-sandbox/eventing-kafka/issues/413

pierDipi: At least, that's the plan / there is a possibility.

Since we will use the annotation suffix `delivery.order`, I'd call this `DeliveryOrder` to keep the language or mapping consistent and straightforward.

slinkydeveloper: Ok I've renamed both enums to DeliveryOrder

pierDipi (approved):

PR #691, opened 2021-03-02
Drop hack and test from CODEOWNERS

pierDipi:
With this setting, we will need approval from
productivity for every new test, which can
easily become a bottleneck.

## Proposed Changes

- Drop hack and test from CODEOWNERS


slinkydeveloper (approved):
Good one

/lgtm

PR #705, opened 2021-03-06
Fix profiler test runner

pierDipi:
The profiler was failing due to missing env var and the profiler release has been deleted.

## Proposed Changes

- Fix profiler test runner


slinkydeveloper (approved):
/lgtm

pierDipi:
Unauthorized error

PR #710, opened 2021-03-07
Fix OpenCircuitException: open circuit

pierDipi:
While analyzing logs I see most of the `sacura` flakiness is due to
errors like:
`io.vertx.circuitbreaker.OpenCircuitException: open circuit`, 
which means that the circuit breaker, after switching to its open state, 
doesn't switch its state to closed. The actual reason is not yet clear.

Log history (they are inverted the first lines come from the dispatcher)

```
 {
    "id": "ffe1198c-7c16-472c-9b14-605aeea761ba",
    "history": [
[...]

PR #717, opened 2021-03-09
[release-0.21] Fix OpenCircuitException: open circuit (#710)

pierDipi:
While analyzing logs I see most of the `sacura` flakiness is due to
errors like:
`io.vertx.circuitbreaker.OpenCircuitException: open circuit`,
which means that the circuit breaker, after switching to its open state,
doesn't switch its state to closed. The actual reason is not yet clear.

- Increase backoff delay of sacura trigger
- Set circuit breaker max failure to a number greater than max retries
- Set circuit breaker reset timeout to backoff dealy

Part of #416


slinkydeveloper (approved):
/lgtm

PR #749, opened 2021-03-20
Refresh development docs and flow 

### LINKED ISSUE #740: Simplify the local development loop
Today we have a huge amount of scripts, very dependent on bash, which makes a whole more complex the local development loop.

We need to clean them up and make sure we have a super easy and documented local development loop.

Some actions I want to investigate:

* [ ] Single command build and deploy to local kind the infra for testing. There is already the sourcing of a script and then the run of `test_setup`, but it's dependent on bash, and we can have a script that just does that, or even have maven to do that. Maybe we can have a pom that does that? I've also found a funny thing I want to try it out :smiling_imp: https://github.com/raydac/mvn-golang 
* [ ] Single command to generate release artifacts for local testing (similar to above)
* [ ] Single command run sacura (similar to above)
* [ ] Single command run profiling (similar to above)
* [ ] proto generation can be moved to maven, where we can config a pom to emit both golang and java code, in the respective positions we want them. Some interesting example https://github.com/raydac/mvn-golang/blob/master/mvn-golang-examples/mvn-golang-example-protobuf/pom.xml
* [ ] Building images and pushing them to local kind doesn't really work (for data plane)

Some of these are already there, we might just need to simpligy their usage and/or document these.

We also need to make sure we don't diverge from the scripts used in CI. Preferably, the local dev scripts should be the same used by CI

/assign

pierDipi:
Part of #740 

## Proposed Changes

- Simplify and document development flow


slinkydeveloper:
Let's keep this as wip, i had another idea for the "catch all local dev script" :smile:

slinkydeveloper (commented):
About the makefile (I still didn't check out the DEVELOPMENT.md and CONTRIBUTING.md, since they depend on the makefile):

* There isn't any kafka setup target (is it inside knative_setup?)
* A target only for `deploy-control-plane` and `deploy-data-plane`?
* Some targets can be composed by others (e.g. generate, unit-test) but we don't do that because we use different scripts (maybe because there are some parts cloned among different scripts?!). It's worth in future iterations of this PR to make these rules composed, it may help to reduce the LOCs of bash :)

My problem with makefiles is that the way you developed it is not very "makefile" idiomatic. You should have target dependencies, naming "compliant" to best make practices (e.g. cleanall, buildall, all) etc. What you're really using here of makefiles is just the parsing of the "target" command, nothing else. I don't see the point of using makefiles only for that, because it sets "expectations" and using it as it should be used is also not fitting our needs (because then, for example, integration-test should rebuild everytime kind the whole project).

My idea to improve the local dev loop was something different: a single script, like `./dev.sh knative-setup`, where we have pretty much the targets we have here, but not a single line of "build logic". If you just type `./dev.sh`, it prints all the different things you can do with the script. How does that sound?

As a side note, I was even thinking to use this stuff https://www.bughunter2k.de/blog/cursor-controlled-selectmenu-in-bash to make the `dev.sh` script nicer to use, but it's definitely not important :)

Makefile
```diff
@@ -0,0 +1,94 @@
...
+# Install dependencies and publish test images
+deps-up:
+	source ${ROOT_DIR}/test/e2e-common.sh && knative_setup
+
+# Delete dependencies
+deps-down:
+	source ${ROOT_DIR}/test/e2e-common.sh && knative_teardown
+
+# Install Eventing Kafka Broker
+up:
```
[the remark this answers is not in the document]
pierDipi: No Kafka is a dependency just like eventing.

Makefile
```diff
@@ -0,0 +1,94 @@
...
+
+# Run code generators
+generate: generate-proto update-codegen
+
+# Run protobuf generator
+generate-proto:
+	${ROOT_DIR}/proto/hack/generate-proto.sh
+
+# Run API generator
+update-codegen:
```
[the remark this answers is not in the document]
pierDipi: This is not for reconcilers only, we generate CM checksums, the contract, and reconcilers.

Makefile
```diff
@@ -0,0 +1,94 @@
...
+# Create a KinD cluster
+kind-up:
+	${ROOT_DIR}/hack/create-kind-cluster.sh
+
+# Delete KinD cluster
+kind-down:
+	kind delete cluster
+
+# Install dependencies and publish test images
+deps-up:
```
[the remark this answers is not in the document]
pierDipi: This is not only eventing, there is eventing, Kafka, test images, and whatever will come next.

hack/create-kind-cluster.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
pierDipi: I pushed this logic to the eventing repository, I'll re-use that one.

hack/update-codegen.sh
```diff
@@ -51,5 +51,5 @@ ${KNATIVE_CODEGEN_PKG}/hack/generate-knative.sh "injection" \
 
 group "Update deps post-codegen"
 
-${REPO_ROOT_DIR}/proto/hack/generate_proto
+${REPO_ROOT_DIR}/proto/hack/generate-proto.sh
```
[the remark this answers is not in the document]
pierDipi: I'd do this in a follow-up PR, not really related to this PR.

pierDipi:
> There isn't any kafka setup target (is it inside knative_setup?)

Yes, Kafka is a dependency just like eventing, in my dev loop, I usually deploy both of them when I start.

> My problem with makefiles is that the way you developed it is not very "makefile" idiomatic. You should have target dependencies, naming "compliant" to best make practices (e.g. cleanall, buildall, all) etc. What you're really using here of makefiles is just the parsing of the "target" command, nothing else. I don't see the point of using makefiles only for that, because it sets "expectations" and using it as it should be used is also not fitting our needs (because then, for example, integration-test should rebuild everytime kind the whole project).
>
> My idea to improve the local dev loop was something different: a single script, like ./dev.sh knative-setup, where we have pretty much the targets we have here, but not a single line of "build logic". If you just type ./dev.sh, it prints all the different things you can do with the script. How does that sound?

This is what I wanted with this PR, I chose makefile because I don't need to build the error-prone logic for parsing arguments.

---

Regarding command names:

I chose 2 terms, one the opposite of the other, and used them throughout: `up` and `down`.
These 2 terms follow a single component: `chaos-(up|down)`, `sacura-(up|down)`, `deps-(up|down)`.
Your suggestion in my opinion make commands less rememberable:
- sometimes `run`
- sometimes `deploy`
- sometimes `clean`
- sometimes `teardown`

IMO a single term should be used, and I really like `up` and `down`, they are short and simple to type.

Makefile
```diff
@@ -0,0 +1,94 @@
...
+# Delete Eventing Kafka Broker
+down:
+	source ${ROOT_DIR}/test/e2e-common.sh && test_teardown
+
+# Run unit tests
+ut:
+	${ROOT_DIR}/test/presubmit-tests.sh --unit-tests
+
+# Run data-plane unit tests
+dp-ut:
```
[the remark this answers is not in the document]
pierDipi: IMO it's too long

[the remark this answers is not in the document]
slinkydeveloper: longer is better than incomprehensible.

Makefile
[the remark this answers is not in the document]
slinkydeveloper: Ok what about `deploy-infra`?

Makefile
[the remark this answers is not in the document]
slinkydeveloper: Ah ok I misunderstood, then `generate` is fine

slinkydeveloper:
> This is what I wanted with this PR, I chose makefile because I don't need to build the error-prone logic for parsing arguments.

I understand your point of view, but the makefile you've developed is not what an usual GNU make user is going to expect. When you use a tool like that, you inherently set expectations like the ones described above.

I know that going to a shell script adds error-prone parsing logic, but here we're merely talking about taking the first positional argument and do a switch case on it...

> Also, If we use integration-test, unit-test, and so on, why the command to run sacura test should be test-sacura?

That was a mistake i think, `sacura-test` is fine :smile:

pierDipi:
Ok, naming updates, I'll use `deploy-<something>`, `teardown-<something>`, and `<something>-test`

pierDipi:
> I understand your point of view, but the makefile you've developed is not what an usual GNU make user is going to expect. When you use a tool like that, you inherently set expectations like the ones described above.
>
> I know that going to a shell script adds error-prone parsing logic, but here we're merely talking about taking the first positional argument and do a switch case on it...

Ok, I'll move this to a bash script.

hack/update-codegen.sh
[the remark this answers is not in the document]
pierDipi: Opened an issue.

slinkydeveloper (commented):
Do you need to rebase? Why the vendor dir is being updated here?

hack/run
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
```
slinkydeveloper, line 1: Add the `.sh` extension

hack/tools.go
```diff
@@ -23,6 +23,9 @@ import (
 	_ "knative.dev/pkg/configmap/hash-gen"
 	_ "knative.dev/pkg/hack"
 
+	// Hack scripts from eventing
+	_ "knative.dev/eventing/hack"
```
slinkydeveloper, line 27: Ahhh now I see why the vendor dir

slinkydeveloper: For what specifically we use the `eventing/hack` except the `create-kind-cluster.sh`?

hack/run
```diff
@@ -0,0 +1,69 @@
...
+  # Run control-plane unit tests
+	go test -race ./...
+elif [[ "${action}" == "build-test" || "${action}" == "build-tests" ]]; then
+  # Run build tests
+	"${ROOT_DIR}"/test/presubmit-tests.sh --build-tests
+elif [[ "${action}" == "integration-test" || "${action}" == "integration-tests" ]]; then
+  # Run integration tests
+	source "${ROOT_DIR}"/test/e2e-common.sh && go_test_e2e -timeout=30m ./test/e2e_new/...
+	source "${ROOT_DIR}"/test/e2e-common.sh && go_test_e2e -timeout=30m ./test/e2e/...
+	source "${ROOT_DIR}"/test/e2e-common.sh && go_test_e2e -tags=deletecm ./test/e2e/...
```
slinkydeveloper, line 37: Same comment i did in the makefile, let's have the actual "test commands" in the proper script. 

The reason is that the knative developers expect those commands in `e2e-tests.sh`, not here.

Also, if I'm not mistake, you're creating a circular ref between scripts. Although this technically doesn't break, it definitely breaks my mind :laughing:. Let's keep this script just an "interface" to run scripts, but then the actual logic where it belongs

hack/run
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+
+set -ex
+
+ROOT_DIR=$(dirname $0)/..
+
+action="$1"
```
slinkydeveloper, line 7: if there's no $1, then print all the available commands

slinkydeveloper (commented):
Can you split this refactoring a bit? It helps me understand where we're going.

For example (but you don't have to follow it):

* The `create-kind-cluster.sh` + the GH action change
* The profiler script changes and doc
* Changing test/config/sacura splitting in 2 files
* Moving generate proto
* Changes to the e2e-* scripts
* The `run.sh` script + its doc

hack/tools.go
pierDipi: I think it's the only script we use here.

slinkydeveloper: Is it worth vendoring it with all that new deps more than just copy-pasting it?

CONTRIBUTING.md
```diff
@@ -1,11 +1,51 @@
...
+- [`Go`](https://golang.org/) - (_required_)
+    - check
+      [go \<version\>](https://github.com/knative-sandbox/eventing-kafka-broker/blob/master/go.mod)
+      for the required Go version used in this project
+- [`Java`](https://www.java.com/en/) (we recommend an `openjdk` build) -
+  (_required_)
+    - check
+      [java.version](https://github.com/knative-sandbox/eventing-kafka-broker/blob/master/data-plane/pom.xml)
+      maven property for the required Java version used in this project
+
```
slinkydeveloper, line 25: You should add as requirements here:

* `protoc`. Specify that the version to use is here https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/pkg/contract/contract.pb.go#L4. As guide to install it, you can point this one https://grpc.io/docs/protoc-installation/#install-pre-compiled-binaries-any-os
* `protoc-gen-go` Specify that the version to use is here https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/pkg/contract/contract.pb.go#L3. To install it:

```
go install google.golang.org/protobuf/cmd/protoc-gen-go@[version]
```

pierDipi:
Fancy output :)
```
Unrecognized action

Usage                                                      for more information read DEVELOPMENT.md
./hack/run.sh <command>

command:
   deploy-infra                                            Deploy eventing, Kafka (Strimzi), publish test images
   teardown-infra                                          Remove eventing, Kafka (Strimzi)
   deploy                                                  Deploy eventing-kafka-broker
   teardown                                                Remove eventing-kafka-broker
   unit-tests, unit-test                                   Run unit tests
   unit-tests-data-plane, unit-test-data-plane             Run data-plane unit tests
   unit-tests-control-plane, unit-test-control-plane       Run control-plane unit tests
   build-tests-control-plane, build-test-control-plane     Run build tests
   integration-tests, integration-test-control-plane       Run integration tests
   deploy-sacura                                           Deploy sacura job
   sacura-test                                             Run sacura tests
   teardown-sacura                                         Remove sacura job
   deploy-chaos                                            Deploy chaosduck
   teardown-chaos                                          Remove chaosduck
   profiler                                                Run profiling tests
   generate                                                Run code generators
```

hack/run
pierDipi: sgtm, done!

test/e2e-tests.sh
```diff
@@ -25,10 +28,8 @@ fi
 save_release_artifacts || fail_test "Failed to save release artifacts"
 
 if ! ${LOCAL_DEVELOPMENT}; then
-  scale_controlplane kafka-controller kafka-webhook-eventing eventing-webhook eventing-controller
-  wait_until_pods_running knative-eventing || fail_test "Pods in knative-eventing didn't come up"
-  apply_chaos || fail_test "Failed to apply chaos"
-  apply_sacura || fail_test "Failed to apply Sacura"
+  "${ROOT_DIR}"/hack/run.sh deploy-sacura || fail_test "Failed to apply Sacura"
+  "${ROOT_DIR}"/hack/run.sh deploy-chaos || fail_test "Failed to apply chaos"
```
slinkydeveloper, line 32: You still have these depending on run.sh

test/e2e-tests.sh
```diff
@@ -39,20 +40,10 @@ header "Running tests"
...
+go_test_e2e -timeout=30m ./test/e2e_new/...
+go_test_e2e -timeout=30m ./test/e2e/...
+go_test_e2e -tags=deletecm ./test/e2e/...
 
-go_test_e2e -tags=deletecm ./test/e2e/... || failed=true
-
-if [ $failed = true ]; then
-  fail_test "Integration tests failed"
-fi
+"${ROOT_DIR}"/hack/run.sh sacura-tests  || fail_test "Sacura test failed"
```
slinkydeveloper, line 47: You still have this depending on run.sh

slinkydeveloper (commented):
I love this PR! I left 2 leftovers from the previous change, there's still this comment I want to address https://github.com/knative-sandbox/eventing-kafka-broker/pull/749/files#r620054753 and then for me we're good to go. This is a great addition to the project :)

test/e2e-tests.sh
pierDipi: Done

hack/tools.go
slinkydeveloper: Ping on this @pierDipi. My worry is that `eventing/hack` is bringing in that mako proto code, which might at some point create problems with our protobuf version.

pierDipi: Good point, I'll look at this.

pierDipi: /hold

control-plane/pkg/contract/contract.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.26.0
-// 	protoc        v3.15.8
+// 	protoc-gen-go v1.25.0-devel
+// 	protoc        v3.14.0
```
pierDipi, line 4: I don't have these versions aligned locally, this is what I wanted to avoid by using a container.

/hold

slinkydeveloper: Then align them :smile:

slinkydeveloper: You can still use a container locally

pierDipi: Done! :smile: 

/unhold

slinkydeveloper (approved):
/lgtm

slinkydeveloper:
I cannot wait to start using this PR!

slinkydeveloper (approved):
/lgtm

slinkydeveloper:

Maybe that's flakyness? It failed only a test or two... https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-sandbox_eventing-kafka-broker/749/pull-knative-sandbox-eventing-kafka-broker-integration-tests/1396736967135203328

PR #847, opened 2021-04-26
Removed protobuf dockerfile and moved generate_proto.sh script

### LINKED ISSUE #838: Move protobuf generator script from proto/hack to hack
https://github.com/knative-sandbox/eventing-kafka-broker/pull/749#discussion_r598594269

slinkydeveloper:
Removed usage of Empty

Fixes #838

## Proposed Changes

- Removed protobuf Dockerfile, now we're relying on the local setup
- Moved generate_proto script to `/hack/generate_proto.sh` 
- Removed usage of `google.protobuf.Empty`. Include paths of the system setup are a **huge** mess, and there's no reason to let contributors lose 1 afternoon of their life for a one line definition empty message.


slinkydeveloper:
cc @pierDipi

slinkydeveloper:
The verify Github Action is not going to work with this change... What if the `generate-proto.sh` installs the proto compiler (and the protoc-gen-go) if missing?

proto/contract.proto
```diff
@@ -1,11 +1,12 @@
 syntax = 'proto3';
 
-import "google/protobuf/empty.proto";
-
 option go_package = "control-plane/pkg/contract";
 option java_package = "dev.knative.eventing.kafka.broker.contract";
 option java_outer_classname = "DataPlaneContract";
 
+// Protobuf include nightmare? No thanks!
+message Empty {}
```
pierDipi, line 8: Is this safe while upgrading?

pierDipi (commented):
Make sense, but how serving is doing this? They're using protoc afaik.

proto/contract.proto
pierDipi: Can we remove this comment? 
it won't make sense without having context.

slinkydeveloper: > Is this safe while upgrading?

I don't think so, even if it's empty, it's still a different message id I guess. But, it's 100% worth it, we might even need to expand this type at some point, so it's more future proof

slinkydeveloper: > Can we remove this comment?

:rofl: I may have written this in a moment of rage

slinkydeveloper: Now the comment is more specific

slinkydeveloper:
After this PR, we need to make sure we specify in the development guide the `protoc-gen-go` and `protoc` version

slinkydeveloper:
@pierDipi prow now has the protobuf tools we need (thanks @albertomilan for the help!).

What we miss is to figure out how to change the github actions... If I change them here, I guess our funky actions bot will stop working... Maybe should we just disable the protobuf check when running on github actions?

slinkydeveloper:
Ok now it's failing just because the codegenerated stuff doesn't have the project style, but the rest should be ok and ready to merge

proto/contract.proto
pierDipi: Thanks! :smile:

pierDipi (approved):

PR #912, opened 2021-05-19
Add more details in the status about ordered and unordered

### LINKED ISSUE #732: Add DeliveryOrder status condition or annotation for Trigger
**Problem**
In https://github.com/knative-sandbox/eventing-kafka-broker/pull/701 we introduced a new annotation to configure the delivery order and during the discussion there, we come up to the conclusion that would like to have a status condition to give users feedback on it since we don't have a way to signal that the annotation has been read correctly.

- Event consumer

**Exit Criteria**

**Time Estimate (optional):** 
1

- Discussion in https://github.com/knative-sandbox/eventing-kafka-broker/pull/701

slinkydeveloper:

Fixes #732

## Proposed Changes

- :gift: Add some details in the existing subscriber resolved condition about the delivery order. I didn't added a new condition for it because it's quite complicated and requires several changes in eventing too

**Release Note**

```release-note
- :gift: Add some details in the existing subscriber resolved condition about the delivery order.
```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/trigger/trigger.go
+++ b/control-plane/pkg/reconciler/trigger/trigger.go
@@ -142,7 +142,9 @@ func (r *Reconciler) reconcileKind(ctx context.Context, trigger *eventing.Trigge
 	if err != nil {
 		return statusConditionManager.failedToResolveTriggerConfig(err)
 	}
-	statusConditionManager.subscriberResolved()
+	statusConditionManager.subscriberResolved(
+		fmt.Sprintf("Subscriber will receive events with the delivery order: %s", triggerConfig.DeliveryOrder.String()),
+	)
 
 	changed := coreconfig.AddOrUpdateEgressConfig(ct, brokerIndex, triggerConfig, triggerIndex)
 
--- a/control-plane/pkg/reconciler/trigger/trigger_lifecycle.go
+++ b/control-plane/pkg/reconciler/trigger/trigger_lifecycle.go
@@ -125,7 +125,7 @@ func (m *statusConditionManager) failedToUpdateDispatcherPodsAnnotation(err erro
 func (m *statusConditionManager) subscriberResolved(details string) {
 	m.Trigger.GetConditionSet().Manage(&m.Trigger.Status).MarkTrueWithReason(
 		eventing.TriggerConditionSubscriberResolved,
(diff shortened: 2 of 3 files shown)


slinkydeveloper:
cc @pierDipi

control-plane/pkg/reconciler/trigger/trigger.go
```diff
@@ -142,7 +142,9 @@ func (r *Reconciler) reconcileKind(ctx context.Context, trigger *eventing.Trigge
 	if err != nil {
 		return statusConditionManager.failedToResolveTriggerConfig(err)
 	}
-	statusConditionManager.subscriberResolved()
+	statusConditionManager.subscriberResolved(
+		fmt.Sprintf("Subscriber will receive events with the delivery order: %s", triggerConfig.DeliveryOrder.String()),
```
pierDipi, line 146: A bit more concise?
```suggestion
		fmt.Sprintf("Delivery order: %s", triggerConfig.DeliveryOrder.String()),
```

control-plane/pkg/reconciler/trigger/trigger_lifecycle.go
```diff
@@ -122,6 +122,10 @@ func (m *statusConditionManager) failedToUpdateDispatcherPodsAnnotation(err erro
 	)
 }
 
-func (m *statusConditionManager) subscriberResolved() {
-	m.Trigger.Status.MarkSubscriberResolvedSucceeded()
+func (m *statusConditionManager) subscriberResolved(details string) {
+	m.Trigger.GetConditionSet().Manage(&m.Trigger.Status).MarkTrueWithReason(
+		eventing.TriggerConditionSubscriberResolved,
+		"SubscriberResolved",
```
pierDipi, line 128: Any public constant?

control-plane/pkg/reconciler/trigger/trigger.go
slinkydeveloper: I prefer this wording, because the name of the reason is still `SubscriberResolved`...

pierDipi: Yeah, make sense, thanks!

pierDipi (approved):

PR #934, opened 2021-05-26
[release-0.21] Do not add a finalizer on triggers without a broker

pierDipi:
This is an cherry-pick of #880

```release-note
Do not add a finalizer on triggers without a broker.
```

/kind bug

Closes #920


slinkydeveloper (approved):
/lgtm

/hold
Merge when you think it's ready

PR #963, opened 2021-06-04
Cleanup the `Tracing` class and move setup as instance method of `TracingConfig`

slinkydeveloper:

Followup of #900

## Proposed Changes

- :broom: Cleanup the `Tracing` class and move setup as instance method of `TracingConfig`
- :wastebasket: Removed `TracingConfig.Builder` as it was used only internally


matzew (approved):
from a refactoring perspective this 
/lgtm

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/tracing/TracingConfig.java
```diff
@@ -39,70 +69,124 @@ private TracingConfig(final Backend backend, final String url, final float sampl
...
-    if (!Files.exists(backendPath)) {
-      return TracingConfig.builder()
-        .setBackend(Backend.UNKNOWN)
-        .createTracingConfig();
+  public OpenTelemetrySdk setup() {
+    logger.info(
+      "Registering tracing configurations {} {} {} {}",
+      keyValue("backend", getBackend()),
+      keyValue("sampleRate", getSamplingRate()),
+      keyValue("url", getUrl()),
```
pierDipi, line 84: `url` might contain sensitive data, keys or passwords of a monitoring provider, even if config-tracing by default is a CM, it's potentially possible to convert and use a secret but we should print it here.

pierDipi (commented):
LGTM

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/tracing/TracingConfig.java
slinkydeveloper: I don't recall this was there before this cleanup or not TBH, but I'll pr it

Issue #864, opened 2021-05-04
KafkaSink reports Ready, actually returns 404 with "resource not found"


antoineco:
**Describe the bug**

KafkaSink reports Ready, but all requests are responded to with a 404, and the receiver contains the following warnings:

`{"@timestamp":"2021-05-03T17:39:48.095Z","@version":"1","message":"resource not found resources=[] path=/myuser/my-kafka-topic","logger_name":"dev.knative.eventing.kafka.broker.receiver.RequestMapper","thread_name":"vert.x-eventloop-thread-0","level":"WARN","level_value":30000,"resources":[],"path":"/myuser/my-kafka-topic"}`

This _seems_ to happen when a KafkaSink is removed, and later re-created.

```yaml
apiVersion: eventing.knative.dev/v1alpha1
kind: KafkaSink
metadata:
  annotations:
    eventing.knative.dev/creator: myuser
    eventing.knative.dev/lastModifier: myuser
    kapp.k14s.io/identity: v1;myuser/eventing.knative.dev/KafkaSink/my-kafka-topic;eventing.knative.dev/v1alpha1
    kapp.k14s.io/original: '{"apiVersion":"eventing.knative.dev/v1alpha1","kind":"KafkaSink","metadata":{"labels":{"kapp.k14s.io/app":"1620120245002031000","kapp.k14s.io/association":"v1.ceceb4609152cf648d1103061d856b85"},"name":"my-kafka-topic","namespace":"myuser"},"spec":{"auth":{"secret":{"ref":{"name":"kafkahackathon"}}},"bootstrapServers":["<redacted>.us-east4.gcp.confluent.cloud:9092"],"topic":"hackathon"}}'
    kapp.k14s.io/original-diff-md5: d03ec57594679a2a193f85b9b0f2388f
  creationTimestamp: "2021-05-04T09:24:12Z"
  finalizers:
  - kafkasinks.eventing.knative.dev
  generation: 1
  labels:
    kapp.k14s.io/app: "1620120245002031000"
    kapp.k14s.io/association: v1.ceceb4609152cf648d1103061d856b85
  name: my-kafka-topic
  namespace: myuser
  resourceVersion: "210533552"
  selfLink: /apis/eventing.knative.dev/v1alpha1/namespaces/myuser/kafkasinks/my-kafka-topic
  uid: 50ba957b-84b2-4158-b213-3f6c836bbfcb
spec:
  auth:
    secret:
      ref:
[...]

PR #974, opened 2021-06-09
[WIP] Probe resource address for ingress readiness

pierDipi:
## Proposed changes

- Probe broker address
- Probe KafkaSink address

Probing is done using the knative/networking data-plane contract
(already implemented in Eventing through knative/pkg).
ref issue https://github.com/knative/eventing/issues/3911

Part of #864 (follow-up consumer readiness)

**Release Note**

```release-note
The control plane probes the data plane for determining resources readiness.
```

/kind bug

### DIFF
--- a/control-plane/pkg/reconciler/base/reconciler.go
+++ b/control-plane/pkg/reconciler/base/reconciler.go
@@ -354,24 +354,24 @@ func (r *Reconciler) probeReceiver(object metav1.Object, namespace, name, ip, po
 	req.Header.Add(network.ProbeHeaderName, "probe")
 	req.Header.Add(network.HashHeaderName, "hash")
 
-	enqueue := func(key string, arg interface{}) { r.Enqueue(arg.(metav1.Object)) }
+	onExpire := func(key string, arg interface{}) { r.Enqueue(arg.(metav1.Object)) }
 
 	go func() {
 		response, err := r.RequestProbeDoer(req)
 		if err != nil {
-			r.ProbeCache.UpdateStatus(target, prober.StatusUnknown, object, enqueue)
+			r.ProbeCache.UpsertStatus(target, prober.StatusUnknown, object, onExpire)
 			// TODO logger
 			// return fmt.Errorf("failed to probe pod %s/%s with ip %s: %w", namespace, name, ip, err)
 			return
 		}
 
 		if response.StatusCode != http.StatusOK {
-			r.ProbeCache.UpdateStatus(target, prober.StatusNotReady, object, enqueue)
+			r.ProbeCache.UpsertStatus(target, prober.StatusNotReady, object, onExpire)
 			// TODO logger
 			// return fmt.Errorf("probe failed response status code %d from pod %s/%s with ip %s", response.StatusCode, namespace, name, ip)
 			return
 		}
-		r.ProbeCache.UpdateStatus(target, prober.StatusReady, object, enqueue)
+		r.ProbeCache.UpsertStatus(target, prober.StatusReady, object, onExpire)
 	}()
 
 	return nil
--- a/control-plane/pkg/reconciler/base/reconciler.go
+++ b/control-plane/pkg/reconciler/base/reconciler.go
@@ -19,6 +19,7 @@ package base
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/encoding/protojson"
@@ -30,6 +31,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/util/retry"
+	"knative.dev/pkg/network"
 	"knative.dev/pkg/tracker"
 
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/contract"
@@ -55,6 +57,10 @@ const (
 
 	Protobuf = "protobuf"
 	Json     = "json"
+
+	ReceiverPort = "8080"
+
+	ProbeProtocol = "http"
 )
 
 // Base reconciler for broker and trigger reconciler.
@@ -74,6 +80,8 @@ type Reconciler struct {
 
 	DispatcherLabel string
 	ReceiverLabel   string
+
+	RequestProbeDoer func(r *http.Request) (*http.Response, error)
 }
 
 func (r *Reconciler) IsReceiverRunning() bool {
@@ -301,4 +309,48 @@ func (r *Reconciler) OnDeleteObserver(obj interface{}) {
 	if r.SecretTracker != nil {
 		r.SecretTracker.OnDeletedObserver(obj)
 	}
+}
+
+func (r *Reconciler) ProbeReceivers(path string) error {
+	namespace := r.SystemNamespace
+
+	pods, err := r.PodLister.Pods(namespace).List(r.receiverSelector())
+	if err != nil {
+		return fmt.Errorf("failed to list pods for probing: %w", err)
+	}
+
+	if len(pods) == 0 {
+		return fmt.Errorf("no receiver pods were found in namespace %s with selector %v", namespace, r.receiverSelector())
+	}
+
+	for _, p := range pods {
+		if podIP := p.Status.PodIP; podIP != "" {
+			if err := r.probeReceiver(p.Namespace, p.Name, podIP, ReceiverPort, path); err != nil {
+				return err // probeReceiver already decorates err.
+			}
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) probeReceiver(namespace, name, ip, port, path string) error {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s://%s:%s%s", ProbeProtocol, ip, port, path), nil)
+	if err != nil {
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Add(network.ProbeHeaderName, "probe")
+	req.Header.Add(network.HashHeaderName, "hash")
+
+	response, err := r.RequestProbeDoer(req)
+	if err != nil {
+		return fmt.Errorf("failed to probe pod %s/%s with ip %s: %w", namespace, name, ip, err)
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf("probe failed response status code %d from pod %s/%s with ip %s", response.StatusCode, namespace, name, ip)
+	}
+
+	return nil
 }
--- a/control-plane/pkg/reconciler/base/reconciler.go
+++ b/control-plane/pkg/reconciler/base/reconciler.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"time"
 
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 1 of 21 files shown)


slinkydeveloper (commented):
Will look in detail later. But the first thing I noticed is that you put the probing in the same http server of the "hot path", can we avoid it and have a separate http server for that (with a separate verticle too?).

My idea to do readyness was to use control-protocol, but it's fine to go in this direction now.

slinkydeveloper:
cc @devguyio can you please check it out?

pierDipi:
> Will look in detail later. But the first thing I noticed is that you put the probing in the same http server of the "hot path", can we avoid it and have a separate http server for that (with a separate verticle too?).

Since I'm implementing the knative/networking contract we must respond to probes on the same address in the status.

I can implement an optimization (in this or a new PR) where we probe the first time and after a configurable delay a pod for a given broker/kafkasink, that reduces the number of requests.

slinkydeveloper:
> Since I'm implementing the knative/networking contract we must respond to probes on the same address in the status.

I rather prefer we modify the code in knative/networking. Having a controller pinging the receiver server it's a bad idea, you're adding unnecessary load to that server...

pierDipi:
But that's a contract, not code.

Eventing does the same thing and with the optimization above we end up doing 2 requests per broker to a pod in a given time span (let's say 1 hour), 2k requests per hour is a really small number that doesn't justify another server with all the synchronization or complexity that that adds.

slinkydeveloper:
> But that's a contract, not code

Where is this contract defined?

pierDipi:
In the issue linked in the body of this PR, I'm not sure if the networking WG has a different document for this.

pierDipi:
https://github.com/knative/pkg/issues/1509
```
Failure", Message:"Internal error occurred: failed calling webhook \"defaulting.webhook.kafka.eventing.knative.dev\": Post \"https://kafka-webhook-eventing.knative-eventing.svc:443/defaulting?timeout=2s\": EOF
```
/retest

control-plane/pkg/reconciler/base/reconciler.go
```diff
@@ -55,6 +57,10 @@ const (
 
 	Protobuf = "protobuf"
 	Json     = "json"
+
+	ReceiverPort = "8080"
```
slinkydeveloper, line 61: Maybe `ProbeReceiverPort`? Even if it's the same server, IMO it's good to clarify that we use it only for that purpose

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -210,6 +213,17 @@ func (r *Reconciler) reconcileKind(ctx context.Context, broker *eventing.Broker)
...
+	ingress := brokerResource.Ingress.IngressType.(*contract.Ingress_Path)
+	if err := r.ProbeReceivers(ingress.Path); err != nil {
+		statusConditionManager.ProbeFailed(err)
+		// When a probe fails, we don't return an error but we requeue our broker so that we can probe it again later.
+		r.EnqueueAfter(broker, time.Duration(r.Configs.ProbeFailureRequeueDelayMs)*time.Millisecond)
+		return nil
+	} else {
+		statusConditionManager.ProbeSucceeded()
+	}
+
```
slinkydeveloper, line 226: I think this is problematic, because it's gonna perform a request to every receiver on every broker reconciliation. I think we should instead apply the same logic we use in eventing-kafka and in networking for performing probing, which reduces that load and applies that periodically

@devguyio is more expert than me on this :smile:

control-plane/pkg/reconciler/broker/controller.go
```diff
@@ -97,6 +100,10 @@ func NewController(ctx context.Context, watcher configmap.Watcher, configs *Conf
 
 	impl := brokerreconciler.NewImpl(ctx, reconciler, kafka.BrokerClass)
 
+	reconciler.EnqueueAfter = func(broker *eventing.Broker, duration time.Duration) {
+		impl.EnqueueAfter(broker, duration)
```
slinkydeveloper, line 104: can you just use the function pointer here? :smile:

control-plane/pkg/reconciler/sink/controller.go
```diff
@@ -73,6 +76,10 @@ func NewController(ctx context.Context, _ configmap.Watcher, configs *config.Env
 
 	impl := sinkreconciler.NewImpl(ctx, reconciler)
 
+	reconciler.EnqueueAfter = func(ks *eventing.KafkaSink, duration time.Duration) {
+		impl.EnqueueAfter(ks, duration)
```
slinkydeveloper, line 80: Same as above

slinkydeveloper (commented):
The thing that you should change in this PR is IMO the way the prober works.

In eventing-kafka the prober is off the reconciliation path, it can be queried to check out if the pod is ready or not and it automatically enqueues resources that has to be updated:

* https://github.com/knative-sandbox/eventing-kafka/blob/main/pkg/channel/consolidated/reconciler/controller/controller.go#L97
* https://github.com/knative-sandbox/eventing-kafka/blob/main/pkg/channel/consolidated/reconciler/controller/controller.go#L170
* https://github.com/knative-sandbox/eventing-kafka/blob/main/pkg/channel/consolidated/reconciler/controller/kafkachannel.go#L256

The same kind of "out of the reconciliation path pattern" is also applied to the way control-protocol notifications work: https://github.com/knative-sandbox/control-protocol/blob/main/docs/async_command_pattern.md

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/RequestMapper.java
```diff
@@ -107,6 +112,19 @@ public RequestMapper(
 
   @Override
   public void handle(final HttpServerRequest request) {
+    // This implements the knative/networking data-plane contract.
```
slinkydeveloper, line 115: Maybe you can implement this as decorator like the `SimpleProbeHandlerDecorator`?

control-plane/pkg/reconciler/broker/broker.go
pierDipi: As I said in a comment somewhere above, I plan to do a follow-up with a caching layer to cache the status.

control-plane/pkg/reconciler/broker/controller.go
pierDipi: I was trying to prevent passing a random object that is not a broker since `impl.EnqueueAfter` takes `interface{}`.

control-plane/pkg/reconciler/base/reconciler.go
pierDipi: SGTM.

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/RequestMapper.java
pierDipi: Sure.

control-plane/pkg/reconciler/broker/broker.go
slinkydeveloper: That's not really an optimization, but instead a different way to reconcile... Is it reasonable to do it here?

pierDipi: Yeah, I understand but the end goal should be the same, it's just (IMO) simpler to reason about in the reconcile flow, the first time even if your just setting up your "out of flow thread" we have to do the request ASAP to set the status, so a synchronous probe makes sense to me here.

Once you have done the first request, we cache that pod p1 has responded for broker b1 with ok, so we don't probe it again for x time.

The "out of the flow thread" would periodically probe broker b1 (even if nothing changes in the system), we instead probe broker b1 once something changes in the system.

slinkydeveloper: > The "out of the flow thread" would periodically probe broker b1 (even if nothing changes in the system), we instead probe broker b1 once something changes in the system that "touches" that particular broker b1.

And that's my problem with this solution, if you have 10 brokers to reconcile one after another, this solution is going to be very expensive, because it's gonna ping all the receiver pods on every reconciliation loop. It will also slow down the reconciliation, maybe even considerably after a certain amount of receiver instances

Caching is also hard, because you need to keep that cache in sync, and there's no easy way to do it. While with probing, you just sync the cache periodically.

> it's just (IMO) simpler to reason about in the reconcile flow,

That might be true, on the other hand, i would love to reuse as much as possible the same pattern we have in eventing-kafka and networking.

This approach might also help us with the "future" problem of dispatcher readyness, similar to the one tackled in eventing-kafka

pierDipi: Sorry, but do you mean that we should probe a receiver pod and assume that all brokers in that pod are good? 
That's not true, see linked issue's comments.

Regarding cache, what do you mean by in sync? In sync with what?

pierDipi: Regarding consumer readiness eventing-kafka has 2 solution, 1 is probing (channel) and the other one (mtsource) is through kafka which is imo more reliable.

pierDipi: Ok, sorry, I missed this. 

I meant that we probe broker b1 when something changes and the time x expires.

pierDipi: Now, it should be clear that it's equivalent

slinkydeveloper:
Since we disagree on the way to do probing and readyness in general, how about we start by merging `SimpleProbeHandlerDecorator` in another PR and then continue here with the probing discussion, rebasing this PR? I would love to loop in some people that have expertise in that sense @lionelvillard @devguyio

control-plane/pkg/reconciler/broker/controller.go
devguyio: prevent who from passing a random object?

devguyio (changes requested):
@pierDipi nice work! I see where you're coming from wrt issuing light weight probe requests to the data plane on reconciliation, but I believe putting async requests in the reconcile loop defies the design principles I understand this broker is trying to adopt, namely following a reactive/async programming model. eventing-kafka consolidated channel has some design flaws technical debt that we should learn from and avoid. 

Basically the recommended reconcile loop pattern is:
1. The check the status of sub objects (pods, db connections, kafka consumer groups etc) from a cache, and initialize watchers if they don't exist that would update the cached status.
2. The status propagation of the sub objects should occur in a reactive, async manner via those watchers and basically :
    1. Cache that status changes.
    2. Trigger an enqueue of the parent object (the `Broker` in this case) on those changes.

That way, you keep sync operations out of the reconcile loop ensuring it executes as fast as possible, while setting up any needed watchers for keeping the status up-to-date and consistent. Otherwise, the reconcile loop would be very problematic. 

That's why the networking wg implemented such functionality for [ingress status](https://github.com/knative/networking/blob/main/pkg/status/status.go) and you can see how they are used in the reconcile loops of kourier ([prober](https://github.com/knative-sandbox/net-kourier/blob/main/pkg/reconciler/ingress/controller.go#L123) and [ReconcileKind](https://github.com/knative-sandbox/net-kourier/blob/main/pkg/reconciler/ingress/ingress.go#L70)) and contour ( [prober](https://github.com/knative-sandbox/net-contour/blob/main/pkg/reconciler/contour/controller.go#L108) and [ReconcileKind]((https://github.com/knative-sandbox/net-contour/blob/main/pkg/reconciler/contour/contour.go#L264) 

So in conclusion and to not shoot ourselves in the foot, how do you feel about thinking of this PR as a PoC of the probing pods idea, and build the probing functionality properly on top of it?

devguyio:
just holding until we agree on best way to move forward
/hold

pierDipi:
Thanks Ahmed for your review!

control-plane/pkg/reconciler/broker/controller.go
pierDipi: We developers of this reconciler, I like having compile-type checks over raw `interface{}`.

Once we had a bug in eventing because we were enqueuing a trigger in place of a broker in the broker reconciler.

PR #979, opened 2021-06-10
Cleanup of the control-plane log package

slinkydeveloper:

## Proposed Changes

- :broom: Renamed log to logging, to be consistent with naming used in eventing and pkg
- :broom: Moved log_marshaller.go to contract package, so it doesn't need to have another "wrapper" type for implementing the ObjectMarshaller interface
- :broom: Renamed log.Logger to CreateReconcileMethodLogger and CreateFinalizeMethodLogger, to make clear what these methods are useful for


control-plane/pkg/logging/logger.go
```diff
@@ -24,16 +24,16 @@ import (
 	"knative.dev/pkg/logging"
 )
 
-func Logger(ctx context.Context, action string, object metav1.Object) *zap.Logger {
+func CreateReconcileMethodLogger(ctx context.Context, object metav1.Object) *zap.Logger {
```
pierDipi, line 27: What if we remove these 2 methods and have only NewLogger that takes ctx, object and action?

slinkydeveloper: Also I opted for `Create` more than `New` because new always recalls me constructors...

pierDipi: I don't like these names they are not really go-ish, NewReconcileLogger?

pierDipi: Or perhaps just remove create and method like  logging.ReconcileLogger?

pierDipi: Anyway, we can iterate on nits later ;)
/lgtm

slinkydeveloper: New is for a constructor, for me it's a no-go here because this is not a constructor of zap.Logger, it's more a wrapper :smile:

slinkydeveloper:
/retest

This doesn't sound related to this PR, let's see if it fails consistently

PR #999, opened 2021-06-15
Improve data-plane pods annotation update logic

pierDipi:
There could be situations where our reconciler failed to update a pod annotation
in a previous reconcile and since there might be multiple pods running we
want to make sure that all pods have the same annotation.

This patch moves the logic to check that we're not performing an
unnecessary update down to when we're trying to update a single pod
annotation.

### DIFF
--- a/control-plane/pkg/reconciler/base/reconciler.go
+++ b/control-plane/pkg/reconciler/base/reconciler.go
@@ -19,6 +19,7 @@ package base
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/encoding/protojson"
@@ -29,7 +30,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
-	"k8s.io/client-go/util/retry"
 	"knative.dev/pkg/tracker"
 
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/contract"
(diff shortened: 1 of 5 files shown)


control-plane/pkg/reconciler/base/reconciler.go
```diff
@@ -243,6 +231,15 @@ func (r *Reconciler) updatePodsAnnotation(ctx context.Context, logger *zap.Logge
 			annotations = make(map[string]string, 1)
 		}
 
+		// Check whether pod's annotation is the expected one.
+		if v, ok := annotations[VolumeGenerationAnnotationKey]; ok {
+			v, err := strconv.ParseUint(v /* base */, 10 /* bitSize */, 64)
+			if err == nil && v == volumeGeneration {
```
slinkydeveloper, line 237: maybe using `v >= volumeGeneration` it's even safer, since this number is incremental?

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -176,38 +174,39 @@ func (r *Reconciler) reconcileKind(ctx context.Context, broker *eventing.Broker)
 	}
 	statusConditionManager.ConfigMapUpdated()
 
-	if changed == coreconfig.ResourceChanged {
+	// We update receiver and dispatcher pods annotation regardless of our contract changed or not due to the fact
+	// that in a previous reconciliation we might have failed to update one of our data plane pod annotation, so we want
+	// to anyway update remaining annotations with the contract generation that was saved in the CM.
```
slinkydeveloper, line 179: s/to anyway update/to update anyway

slinkydeveloper (commented):
Minor comment

control-plane/pkg/reconciler/base/reconciler.go
pierDipi: We can start from 0 again and so `v >= volumeGeneration` is not ok:

https://github.com/knative-sandbox/eventing-kafka-broker/blob/bb90971f62275a019930278f2e72c5361921461d/control-plane/pkg/core/config/utils.go#L125

control-plane/pkg/reconciler/broker/broker.go
pierDipi: Thanks

control-plane/pkg/reconciler/base/reconciler.go
pierDipi: It might happen since we're using the cache but we will retry later on and it will eventually converge, so as long as our CM is consistent we are good.

pierDipi: That might happen even today without this PR.

PR #1022, opened 2021-06-21
Introduced `AsyncCloseable`

slinkydeveloper:
While working on improving the dispatcher code base, I've found out we have several code snippets, all around the `close` problem. This PR is an attempt to solidify, abstract and share code for "closing" components

The `AsyncCloseable` is like the vert.x `Closeable`, but more than accepting a promise, it returns a `Future` instead. It implements vert.x `Closeable` and can be converted to a blocking `java.util.AutoCloseable`.

It has static methods to compose multiple `AsyncCloseable` and to wrap `AutoCloseable`, for better integration with our blocking components.

## Proposed Changes

- :gift: Introduce `AsyncCloseable` interface
- :broom: Cleanup all various usages of close routines and replace them with `AsyncCloseable`

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/AsyncCloseable.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/AsyncCloseable.java
@@ -56,9 +56,11 @@ public interface AsyncCloseable extends Closeable {
    * @return the composed closeables
    */
   static AsyncCloseable compose(AsyncCloseable... closeables) {
-    return () -> CompositeFuture
-      .all(Arrays.stream(closeables).map(AsyncCloseable::close).collect(Collectors.toList()))
-      .mapEmpty();
+    return () -> CompositeFuture.all(
+      Arrays.stream(closeables)
+        .map(AsyncCloseable::close)
+        .collect(Collectors.toList())
+    ).mapEmpty();
   }
 
   /**
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/AsyncCloseable.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/AsyncCloseable.java
@@ -1,0 +1,47 @@
+package dev.knative.eventing.kafka.broker.core;
+
+import io.vertx.core.Closeable;
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+/**
+ * Interface for components that can be closed asynchronously.
+ */
+@FunctionalInterface
+public interface AsyncCloseable extends Closeable {
+
+  /**
+   * Close this object.
+   *
+   * @return a future notifying the completion of the close operation
+   */
+  Future<Void> close();
+
+  @Override
+  default void close(Promise<Void> completion) {
+    this.close().onComplete(completion);
+  }
+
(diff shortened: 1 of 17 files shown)


pierDipi:
Yes, I'm ok with this.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/AsyncCloseable.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright © 2018 Knative Authors (knative-dev@googlegroups.com)
```
matzew, line 2: Do we actually need to provide a date? Or worth to update the templates on the IDE ? :smile:

slinkydeveloper: From what I know, this template is fine and legal friendly, because it is the starting date of the project :smile:

matzew:
Is this something that might be useful for others too? E.g. why is this not something that would make sense on vert.x directly, rather than adding here some kinda library feature

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/AsyncCloseable.java
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Compose several {@link AsyncCloseable} into a single {@link AsyncCloseable}. One close failure will cause the whole close to fail.
+   *
+   * @param closeables the closeables to compose
+   * @return the composed closeables
+   */
+  static AsyncCloseable compose(AsyncCloseable... closeables) {
+    return () -> CompositeFuture
+      .all(Arrays.stream(closeables).map(AsyncCloseable::close).collect(Collectors.toList()))
```
matzew, line 60: would inlining make this nicer to read?

slinkydeveloper:
> Is this something that might be useful for others too? E.g. why is this not something that would make sense on vert.x directly, rather than adding here some kinda library feature

I think the interface might fit inside Vert.x core, although there is already a `Closeable` interface (which is a bit callback-ish style), but we can give it a try in future.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/AsyncCloseable.java
slinkydeveloper: let me format it properly

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/RecordDispatcher.java
```diff
@@ -79,11 +80,7 @@ public RecordDispatcher(
     this.subscriberSender = composeSenderAndSinkHandler(subscriberSender, sinkResponseHandler, "subscriber");
     this.dlsSender = composeSenderAndSinkHandler(deadLetterSinkSender, sinkResponseHandler, "dead letter sink");
     this.offsetManager = offsetManager;
-    this.closer = () -> CompositeFuture.all(
-      sinkResponseHandler.close(),
-      deadLetterSinkSender.close(),
-      subscriberSender.close()
-    );
+    this.closeable = AsyncCloseable.compose(sinkResponseHandler, deadLetterSinkSender, subscriberSender);
```
matzew, line 83: this is actually nice

matzew (approved):
/lgtm
/approve

PR #1028, opened 2021-06-22
Split RecordDispatcher in interface and implementation class

slinkydeveloper:
My plan is to reorganize the code inside the dispatcher package and inside the http package in particular, and decouple a bit more and reorganize the concerns of consumer verticles and record dispatcher. This is one step ahead, next PRs I'll do the actual move of the code. 

## Proposed Changes

- :broom: Split RecordDispatcher in interface and implementation class
- :broom: RecordDispatcher doesn't implement anymore `Function` anymore, nor `Handler`. The `Handler` implementation was a bad idea, because it's hiding a technical detail just for the sake of being easier to use for a single user (the UnorderedConsumerVerticle) 
- :broom: Rename OffsetManager in RecordDispatcherListener, which better resembles the interface purpose.


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/RecordDispatcher.java
```diff
@@ -16,239 +16,15 @@
...
-import org.slf4j.LoggerFactory;
-
-import static dev.knative.eventing.kafka.broker.core.utils.Logging.keyValue;
 
 /**
- * This class implements the core algorithm of the Dispatcher (see {@link
- * RecordDispatcher#handle(KafkaConsumerRecord)}).
- *
- * @see RecordDispatcher#handle(KafkaConsumerRecord)
+ * This class performs the dispatch of consumed records.
```
matzew, line 24: perhaps minor, but this was changed to be an interface, rather than being a clazz

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/consumer/impl/UnorderedConsumerVerticle.java
```diff
@@ -32,7 +32,7 @@ public UnorderedConsumerVerticle(Initializer initializer,
   @Override
   void startConsumer(Promise<Void> startPromise) {
     this.consumer.exceptionHandler(this::exceptionHandler);
-    this.consumer.handler(this.recordDispatcher);
+    this.consumer.handler(record -> this.recordDispatcher.dispatch(record));
```
matzew, line 35: before this was a handler (wrapping the apply), now it's "just" a `RecordDispatcher`.  

There is no more needed to switch from the "framework" implicit `handle` invocation, to the explicit `dispatch` ?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/consumer/impl/OrderedConsumerVerticle.java
```diff
@@ -122,7 +122,7 @@ void recordsHandler(KafkaConsumerRecords<String, CloudEvent> records) {
 
   void enqueueRecord(TopicPartition topicPartition, KafkaConsumerRecord<String, CloudEvent> record) {
     this.recordDispatcherExecutors.computeIfAbsent(topicPartition, (tp) -> new OrderedAsyncExecutor())
-      .offer(() -> this.recordDispatcher.apply(record).onComplete(v -> this.pendingRecords--));
+      .offer(() -> this.recordDispatcher.dispatch(record).onComplete(v -> this.pendingRecords--));
```
matzew, line 125: I actually do like the more specific term `dispatch`, over the generic `Function.apply`

data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/RecordDispatcherTest.java
```diff
@@ -57,7 +56,7 @@ public void shouldNotSendToSubscriberNorToDeadLetterSinkIfValueDoesntMatch() {
     );
 
     final var record = record();
-    dispatcherHandler.handle(record);
```
matzew, line 60: The `handle` invokes the `apply`, which does the actual execution chain.   Now, we will call `dispatch`.

Out of curiousity, could this test have also call directly `apply`, rather than the wrapper?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/consumer/impl/UnorderedConsumerVerticle.java
slinkydeveloper: The reason I removed the `Handler` impl for clarity of the contract: `RecordDispatcher.dispatch` returns a `Future`, if you wanna discard that future is up to you. But having `Handler` implementation of `RecordDispatcher.dispatch` created a kind of "false illusion" that the `RecordDispatcher` is implementing the `Handler` contract "properly"

data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/consumer/impl/OrderedConsumerVerticleTest.java
```diff
@@ -81,8 +81,8 @@ public void consumeOneByOne(final long delay, final int tasks, final int partiti
     // Mock the record dispatcher to count down the latch and save the received records order
     CountDownLatch latch = new CountDownLatch(tasks);
     final Map<TopicPartition, List<Long>> receivedRecords = new HashMap<>(partitions);
-    final var recordDispatcher = mock(RecordDispatcher.class);
-    when(recordDispatcher.apply(any())).then(invocation -> {
+    final var recordDispatcher = mock(RecordDispatcherImpl.class);
+    when(recordDispatcher.dispatch(any())).then(invocation -> {
```
matzew, line 85: here in contrast we used to invoke the `apply`, instead of `handle`

data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/RecordDispatcherTest.java
slinkydeveloper: There is not `apply` method anymore in `RecordDispatcher`, but yes using the previous shape of the file, it could have

data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/consumer/impl/OrderedConsumerVerticleTest.java
slinkydeveloper: because ordered consumption requires to know when the dispatch ends, when unordered consumption does not

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/consumer/impl/UnorderedConsumerVerticle.java
matzew: thanks for the reasoning!

data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/RecordDispatcherTest.java
matzew: yeah, I meant about the past version.

matzew:
LGTM. I like the reorg!

leaving the final approval to others on the team

PR #1041, opened 2021-06-24
Reorganize RequestMapper and friends

slinkydeveloper:
The goal of this PR is to properly split `RequestMapper` into different classes, one for each concern. The final result of this PR looks like this:

* The part of `RequestMapper` doing the reconciliation of the internal maps of producers is moved inside the class called `IngressProducerReconciliableStore`.
* The logic of the request handling is not organized anymore in decorators and then `RequestMapper` as the final step, but it's "orchestrated" by the `ReceiverVerticle.handle` method as follows:
  1. A bunch of "pre handlers" are executed, these includes the method not allowed check, the probe handler and potentially other checks/headers put in future, like the webhoook stuff. These just implement `Handler<HttpServerRequest>` and do not wrap the next handler anymore.
  2. The `ReceiverVerticle` owns the `IngressProducerReconciliableStore`, which it queries to find out the `IngressProducer` associated to that request
  3. If the `IngressProducer` resolved is not null, the `IngressRequestHandler` is invoked, which will produce the record and end the request


slinkydeveloper:
cc @pierDipi how does it looks like? I still need to reorg the tests, but is it a bit more organized now?

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/handler/MethodNotAllowedHandler.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public void handle(final HttpServerRequest request) {
+    if (request.method() != HttpMethod.POST) {
+      request.response().setStatusCode(METHOD_NOT_ALLOWED.code()).end();
+
+      logger.warn("Only POST method is allowed. Method not allowed: {}",
+        keyValue("method", request.method())
+      );
+    }
```
matzew, line 52: I think we should allow `OPTIONS` too, see here: https://github.com/knative/eventing/pull/5542

and perhaps also set the `Allow` header (see PR above too for this)

slinkydeveloper: Another PR and another handler singleton :smile:

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Collections;
+import java.util.Objects;
+import java.util.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static dev.knative.eventing.kafka.broker.core.utils.Logging.keyValue;
+import static io.netty.handler.codec.http.HttpResponseStatus.NOT_FOUND;
+
+public class ReceiverVerticle extends AbstractVerticle {
```
matzew, line 39: wanna add some javadoc here?

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
```diff
@@ -0,0 +1,117 @@
...
+      preHandler.handle(request);
+      // preHandler might end the request prematurely
+      if (request.isEnded()) {
+        return;
+      }
+    }
+
+    // Look up for the ingress producer
+    IngressProducer producer = this.ingressProducerStore.resolve(request.path());
+    if (producer == null) {
```
matzew, line 105: when would we hit this?

matzew:
General I like the notion of `Ingress...` on the class names. We should do for "egress" too (e.g. `Egress...`).

Just wondering should we also move that to dedicated packages?

slinkydeveloper:
> Just wondering should we also move that to dedicated packages?

About package reorg, I think what we have in this PR is already good enough... Do you have something in particular?

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
slinkydeveloper: if you're doing a request to a broker path that does not exists

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
slinkydeveloper: Added some details from the PR description

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/handler/MethodNotAllowedHandler.java
matzew: OK. makes sense. I'd like to send that handler for this - OK ?

data-plane/receiver/src/test/java/dev/knative/eventing/kafka/broker/receiver/handler/IngressRequestHandlerImplTest.java
```diff
@@ -0,0 +1,145 @@
+package dev.knative.eventing.kafka.broker.receiver.handler;
```
pierDipi, line 1: Copyright

data-plane/receiver/src/test/java/dev/knative/eventing/kafka/broker/receiver/handler/PreHandlerTest.java
```diff
@@ -0,0 +1,71 @@
+package dev.knative.eventing.kafka.broker.receiver.handler;
```
pierDipi, line 1: Copyright

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
```diff
@@ -0,0 +1,128 @@
...
+      .<Void>mapEmpty()
+      .onComplete(stopPromise);
+  }
+
+  private void handle(HttpServerRequest request) {
+    // Run preHandlers
+    for (Handler<HttpServerRequest> preHandler : this.preHandlers) {
+      preHandler.handle(request);
+      // preHandler might end the request prematurely
+      if (request.isEnded()) {
```
pierDipi, line 109: What if a preHandler handles a request asynchronously, would this return false?
It seems we can a build a more robust design here :thinking:

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/IngressRequestHandler.java
```diff
@@ -0,0 +1,27 @@
...
+package dev.knative.eventing.kafka.broker.receiver;
+
+import io.vertx.core.http.HttpServerRequest;
+
+/**
+ * This class handles incoming ingress requests.
+ */
+public interface IngressRequestHandler {
+
+  void handle(HttpServerRequest request, IngressProducer producer);
```
pierDipi, line 25: Nice!

pierDipi (commented):
I like this! I left a few comments.

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
slinkydeveloper: Do we have an actual use case for that? I feel like I'm recreating `vertx-web`, and I don't wanna go down that route :smile: How about we introduce that change only if a use case comes out?

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/handler/MethodNotAllowedHandler.java
slinkydeveloper: ok, after this pr is merged feel free to do so.

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
pierDipi: A class shouldn't change because another one changes. 
Why don't we organize these handlers as a chain https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern?

slinkydeveloper: > A class shouldn't change because another one changes.

Of course, what I meant is that because we have no use case for async pre handlers, nor I can think of any future use cases in mind for that, there is no need for `ReceiverVerticle` to support async handlers, because it makes the code more complicated for no reason. 

What I'm implementing here is already a chain of responsability, where handlers are invoked one after the other where each one does its job, It's just not meant to include async handlers. In order to implement it with async handlers, we should have `Function<HttpServerRequest, Future<Void>>`.

The reason why the `IngressRequestHandler` is "special" is because I would like to have the implementation be unaware of the store, the reconciliation and all the complexity brought in with it. I tried to find a nice way to fix that, but then the injection of the `vertx` object would become too much complicated, so I'm out of solutions atm :disappointed: 

The solution in this PR is not be best solution of course, but I guess it's a beginning...

matzew (approved):
/lgtm
/hold


for other to jump in

pierDipi:
There are failures https://github.com/knative-sandbox/eventing-kafka-broker/pull/1041/checks?check_run_id=2931263174
```
Error:  Errors: 
Error:    RequestMapperTest.lambda$testStore$18:298->lambda$testStore$15:298 » NullPointer
Error:    RequestMapperTest.lambda$testStore$18:298->lambda$testStore$15:298 » NullPointer
Error:    RequestMapperTest.lambda$testStore$18:298->lambda$testStore$15:298 » NullPointer
Error:    RequestMapperTest.lambda$testStore$18:298->lambda$testStore$15:298 » NullPointer
Error:    RequestMapperTest.lambda$testStore$18:298->lambda$testStore$15:298 » NullPointer
```

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
pierDipi: I see, let's keep this for now

PR #1042, opened 2021-06-25
Fix build on MacOS

aliok:
Fixes #

`uuid --hdr` doesn't work on MacOs 🕶️ 

## Proposed Changes

- Added an if/else to run command with different param on MacOS
-
-

**Release Note**

```release-note

```

**Docs**


slinkydeveloper (approved):
/lgtm

PR #1052, opened 2021-06-28
Reorganize dispatcher code

slinkydeveloper:

Followup of #1028. Follows the same criteria of https://github.com/knative-sandbox/eventing-kafka-broker/pull/1041. As the other PRs, it's a lot of files changed, but don't be scared, there is a very little number of changes:

* Renamed `ConsumerRecordSender` to `CloudEventSender` and modified its interface to accept directly the event, more than the record. There is no reason for this class to accept `record`, so this seemed like leaking details of the overall architecture of the `RecordDispatcher` logic.
* Renamed `SinkResponseHandler` to `ResponseHandler`. The reason is that the original name is confusing, since sink has a specific meaning for knative eventing, while the class is used also for handling the response of the dead letter.
* Collapsed `consumer` package in the `dispatcher` package, now the two "interface" are inside `dispatcher`, while the impl classes are under `impl.consumer`
* Moved `ConsumerDeployerVerticle` and `ConsumerVerticleFactoryImpl` (previously named `HttpConsumerVerticleFactory`) to `main` package
* Moved `RecordDispatcherImpl` to `impl` subpackage
* Renamed `HttpSinkResponseHandler` to `KafkaResponseHandler`, to identify that the response is sent back to kafka (another impl will be necessary to send back to another channel, when implementing KafkaChannel).
* Renamed `HttpConsumerRecordSender` to `WebClientCloudEventSender`, in order to underline that `WebClient` is used behind the hood.


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
```diff
@@ -168,7 +169,7 @@ public AbstractVerticle get(final DataPlaneContract.Resource resource, final Dat
           filter,
           egressSubscriberSender,
           egressDeadLetterSender,
-          new HttpSinkResponseHandler(vertx, resource.getTopics(0), producer),
+          new KafkaResponseHandler(producer, resource.getTopics(0)),
```
matzew, line 172: why don't we provide the `vertx` instance to the ResponseHandler anymore?

matzew:
/lgtm
/hold
for other to review and wave in

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
slinkydeveloper: Simply, we don't need it anymore (this was probably required before for the tracing old impl)

matzew:
@slinkydeveloper feel free to rebase and unhold?

PR #1056, opened 2021-06-28
Renamed getProperties to readPropertiesSync

slinkydeveloper:

Renamed both `getProperties` and `getPropertiesAsJson` to `read[]Sync` in order to underline that the method is performing a file read and that the method is blocking, thus it shouldn't be called on the event loop

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/utils/Configurations.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/utils/Configurations.java
@@ -32,7 +32,7 @@ public class Configurations {
   /**
    * Retrieve a properties file. Note: this method is blocking
    */
-  public static Properties getProperties(final String path) {
+  public static Properties readPropertiesSync(final String path) {
     if (path == null) {
       return new Properties();
     }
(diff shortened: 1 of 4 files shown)


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/utils/Configurations.java
```diff
@@ -32,7 +32,7 @@
   /**
    * Retrieve a properties file. Note: this method is blocking
```
matzew, line 33: wanna add `thus it shouldn't be called on the event loop` to the JavaDoc?

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/utils/Configurations.java
```diff
@@ -50,8 +50,8 @@ public static Properties getProperties(final String path) {
   /**
    * Retrieve a properties file and translates it to json. Note: this method is blocking
```
matzew, line 51: wanna add `thus it shouldn't be called on the event loop` to the JavaDoc?

matzew (approved):
/lgtm
/approve

PR #1062, opened 2021-06-29
File watcher thread

slinkydeveloper:
This PR started as an attempt to:

* Hide a bit more the details of the file watcher, making its interface as simple as `start` and `stop`. This might prove useful in future, in case we'll want to remove the file watcher concrete class and use an interface to define objects that "reads the contract from the outside, and publish updates on the event bus"
* Run the file watcher out of the main thread

## Proposed Changes

- :broom: Now `FileWatcher` hides a bit more the internal details, and its API surface is as easy as `start` and `stop`, where internally there is an owned `Thread` that is managed to execute the watch out of the main thread.


matzew:
@slinkydeveloper can u rebase?

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/file/FileWatcher.java
```diff
@@ -39,73 +38,103 @@
...
-   * @throws IOException watch service cannot be registered.
    */
-  public FileWatcher(
-    final WatchService watcher,
-    final Consumer<DataPlaneContract.Contract> contractConsumer,
-    final File file)
-    throws IOException {
-
-    Objects.requireNonNull(contractConsumer, "provide consumer");
+  public FileWatcher(File file, Consumer<DataPlaneContract.Contract> contractConsumer) {
```
matzew, line 65: I like moving the `WatchService` to the inside here

matzew (approved):
/lgtm
/approve

PR #1162, opened 2021-08-23
Channel lifecycle

aliok:
Fixes #

## Proposed Changes

- Implement lifecycle for KafkaChannel
- Mostly taken from distributed ch impl
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
+++ b/control-plane/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
@@ -77,6 +77,22 @@ func (*KafkaChannel) GetConditionSet() apis.ConditionSet {
 	return kc
 }
 
+// IsReady returns true if the Status condition KafkaChannelConditionReady
+// is true and the latest spec has been observed.
+func (kc *KafkaChannel) IsReady() bool {
+	cs := kc.Status
+	return cs.ObservedGeneration == kc.Generation &&
+		cs.GetCondition(KafkaChannelConditionReady).IsTrue()
+}
+
+// IsFailed returns true if the resource has observed the latest generation
+// and ready is false.
+func (kc *KafkaChannel) IsFailed() bool {
+	kcs := kc.Status
+	return kcs.ObservedGeneration == kc.Generation &&
+		kcs.GetCondition(KafkaChannelConditionReady).IsFalse()
+}
+
 // GetConditionSet retrieves the condition set for this resource.
 func (*KafkaChannelStatus) GetConditionSet() apis.ConditionSet {
 	channelCondSetLock.RLock()
--- a/control-plane/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
+++ b/control-plane/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 The Knative Authors
+ * Copyright 2020 The Knative Authors
  *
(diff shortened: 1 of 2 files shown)


control-plane/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
```diff
@@ -17,30 +17,72 @@
...
-// ConditionAddressable is a reference to the Addressable condition in the base package.
-// KafkaChannel uses the common Condition set defined in the base package.
+var kc = apis.NewLivingConditionSet(
+	KafkaChannelConditionTopicReady,
+	KafkaChannelConditionDispatcherReady,
+	KafkaChannelConditionServiceReady,
+	KafkaChannelConditionEndpointsReady,
+	KafkaChannelConditionAddressable,
+	KafkaChannelConditionChannelServiceReady,
+	KafkaChannelConditionConfigReady)
```
aliok, line 35: `KafkaChannelConditionDispatcherReady` --> we won't need readiness condition for the receiver part. It is very well possible that there are no subscribers.

aliok: Well, ok, now I remember. `KafkaChannelConditionEndpointsReady` is pretty much the same with that.

aliok:
cc @devguyio @matzew 
Now you can comment about the lifecycle related stuff !

control-plane/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
matzew: Would it make sense to rename it, to have a better distinction between "receiver" and "dispatcher" ? To me  `KafkaChannelConditionEndpointsReady` sounds more generic, less specific

matzew:
The discussion around doing this on the lifecycle:

```go
// IsReady returns true if the Status condition KafkaChannelConditionReady
// is true and the latest spec has been observed.
func (kc *KafkaChannel) IsReady() bool {
	cs := kc.Status
	return cs.ObservedGeneration == kc.Generation &&
		cs.GetCondition(KafkaChannelConditionReady).IsTrue()
}

// IsFailed returns true if the resource has observed the latest generation
// and ready is false.
func (kc *KafkaChannel) IsFailed() bool {
	kcs := kc.Status
	return kcs.ObservedGeneration == kc.Generation &&
		kcs.GetCondition(KafkaChannelConditionReady).IsFalse()
}

```

I'd like to also separate in a new PR (upstream _and_ here). I will send one upstream first

aliok:
@matzew I added these methods here in this PR

control-plane/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
aliok: overall, the naming of these are not good. I will now try to find better names for all

matzew:
I added them upstream too: https://github.com/knative-sandbox/eventing-kafka/pull/825  with tests and stuff

Since here we just get started, I think we can move forward and fix any other issues there (since eventually we reused the code directly from there, instead of copy-paste here)

/lgtm
/approve

control-plane/pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
aliok: The conditions we now have are solid. Won't touch any of these.

PR #1217, opened 2021-09-07
Improve test coverage for security.config/TestSecret

aliok:
Fixes #

## Proposed Changes

- Improve test coverage for security.config/TestSecret
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/sink/auth_provider.go
+++ b/control-plane/pkg/reconciler/sink/auth_provider.go
@@ -24,13 +24,16 @@ type SecretLocator struct {
 	*eventing.KafkaSink
 }
 
-func (ks *SecretLocator) SecretName() (string, bool, error) {
+func (ks *SecretLocator) SecretName() (string, bool) {
 	if !ks.Spec.HasAuthConfig() {
-		return "", false, nil
+		return "", false
 	}
-	return ks.Spec.Auth.Secret.Ref.Name, true, nil
+	return ks.Spec.Auth.Secret.Ref.Name, true
 }
 
-func (ks *SecretLocator) SecretNamespace() (string, bool, error) {
-	return ks.Namespace, true, nil
+func (ks *SecretLocator) SecretNamespace() (string, bool) {
+	if !ks.Spec.HasAuthConfig() {
+		return "", false
+	}
+	return ks.Namespace, true
 }
--- a/control-plane/pkg/reconciler/sink/auth_provider.go
+++ b/control-plane/pkg/reconciler/sink/auth_provider.go
@@ -32,8 +32,5 @@ func (ks *SecretLocator) SecretName() (string, bool) {
 }
 
 func (ks *SecretLocator) SecretNamespace() (string, bool) {
-	if !ks.Spec.HasAuthConfig() {
-		return "", false
-	}
-	return ks.Namespace, true
+	return ks.Namespace, ks.Spec.HasAuthConfig()
 }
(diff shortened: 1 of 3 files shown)


control-plane/pkg/reconciler/sink/auth_provider.go
```diff
@@ -24,13 +24,16 @@ type SecretLocator struct {
 	*eventing.KafkaSink
 }
 
-func (ks *SecretLocator) SecretName() (string, bool, error) {
+func (ks *SecretLocator) SecretName() (string, bool) {
```
aliok, line 27: Got rid of the 3rd param here as we never return an error

control-plane/pkg/reconciler/sink/auth_provider.go
```diff
@@ -24,13 +24,16 @@ type SecretLocator struct {
...
-		return "", false, nil
+		return "", false
 	}
-	return ks.Spec.Auth.Secret.Ref.Name, true, nil
+	return ks.Spec.Auth.Secret.Ref.Name, true
 }
 
-func (ks *SecretLocator) SecretNamespace() (string, bool, error) {
-	return ks.Namespace, true, nil
+func (ks *SecretLocator) SecretNamespace() (string, bool) {
```
aliok, line 34: Same story here

control-plane/pkg/security/config_test.go
```diff
@@ -92,6 +92,29 @@ func TestNewSaramaSecurityOptionFromSecret(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
```
aliok, line 95: added 2 new test cases

control-plane/pkg/security/config_test.go
```diff
@@ -50,7 +50,7 @@ func (sp *SecretProviderFuncMock) F(ctx context.Context, namespace, name string)
 	return sp.secret, sp.err
 }
 
-func TestNewSaramaSecurityOptionFromSecret(t *testing.T) {
+func TestSecret(t *testing.T) {
```
aliok, line 53: Test func name was bad

aliok:
cc @pierDipi @devguyio @matzew

control-plane/pkg/security/config_test.go
```diff
@@ -92,6 +92,29 @@ func TestNewSaramaSecurityOptionFromSecret(t *testing.T) {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "my-ns",
+						Name:      "my-name",
+					},
+					// No ref to secret here
+					Data: map[string]string{},
+				},
+			},
+			wantSecret: nil,
+			wantErr:    false,
```
pierDipi, line 116: can you remove this `wantError`?

control-plane/pkg/reconciler/sink/auth_provider.go
```diff
@@ -24,13 +24,16 @@ type SecretLocator struct {
...
+	if !ks.Spec.HasAuthConfig() {
+		return "", false
+	}
+	return ks.Namespace, true
```
pierDipi, line 38: what about this?
```suggestion
	return ks.Namespace, ks.Spec.HasAuthConfig()
```

aliok: I can revert that one, not a big issue.

I thought:
- If there's no secret specified in the spec, we should also make `SecretNamespace()` return nil, similar to `SecretName()`

In the end, the `SecretName()` is always called before `SecretNamespace()` and the execution stops before `SecretNamespace()` call, but I like consistency.

I don't have a strong opinion and can revert

aliok: Oh, you have a suggestion. That's actually better. Committing that now

pierDipi (approved):

pierDipi:
Thanks!

aliok:

Data plane issue

PR #1221, opened 2021-09-08
Inline config.Env in channel reconciler

aliok:
Fixes #

## Proposed Changes

- Inline config.Env in channel reconciler
- Add a TODO about Channel webhook
-

**Release Note**

```release-note

```

**Docs**


aliok:
cc @pierDipi

PR #1223, opened 2021-09-08
Register event listeners in KafkaChannel controller and register custom condition set

aliok:
Fixes #

## Proposed Changes

- Register event listeners in KafkaChannel controller
- Register custom condition set
-

**Release Note**

```release-note

```

**Docs**


aliok:
cc @pierDipi

pierDipi (approved):

PR #1254, opened 2021-09-20
Add namespace to referenced addressable before resolving destination

### LINKED ISSUE #1011: Dead Letter Sink call not happening after verified failed event delivery
## Expected Behavior
An event delivery sink (KNative service) is called after an event reaches a service with the service returning an error response.

## Actual Behavior
The event delivery sink is not called, and an error message can be found in the `kafka-broker-dispatcher` logs (details below).

## Steps to Reproduce the Problem

1. Create a working setup with kafka cluster and KNative's source -> broker -> trigger -> service, with the service configured to return an error, and the broker configured with a dead letter sink.
2. Send a message to the topic that source and broker are listening to.

## Additional Info
The wider context where this issue is happening:

- Windows machine with Windows 10 and WSL2.
- A Kubernetes cluster built with KinD and MetalLb.
- A Kafka cluster configured within Kubernetes with Strimzi.
- KNative Serving and KNative eventing set up (both 0.22.1), both confirmed working.

I have a test setup with Serving - Eventing, and in order to test that Dead Letter Sinks are working properly, I've configured a service which expected behaviour is to fail. The source - broker - trigger setup seems to be working a-ok, and the event delivery fails as expected, however I can't see the `deadLetterSink` being called.

I've checked logs:
`kubectl logs kafka-broker-dispatcher-54cddc9484-s2kjw -n knative-eventing`

And can find these entries:
```
{"@timestamp":"2021-06-17T12:27:08.522Z","@version":"1","message":"failed to send event to subscriber subscriberURI=http://event-failer.default.svc.cluster.local/nothing statusCode=501","logger_name":"dev.knative.eventing.kafka.broker.dispatcher.http.HttpConsumerRecordSender","thread_name":"vert.x-eventloop-thread-1","level":"ERROR","level_value":40000,"subscriberURI":"http://event-failer.default.svc.cluster.local/nothing","statusCode":501}
{"@timestamp":"2021-06-17T12:27:08.523Z","@version":"1","message":"Failed to send event to subscriber topic=knative-broker-default-default partition=0 offset=7","logger_name":"dev.knative.eventing.kafka.broker.dispatcher.RecordDispatcher","thread_name":"vert.x-eventloop-thread-1","level":"ERROR","level_value":40000,"stack_trace":"io.vertx.core.impl.NoStackTraceThrowable: response status code is not 2xx - got: 501\n","topic":"knative-broker-default-default","partition":0,"offset":7}
{"@timestamp":"2021-06-17T12:27:08.523Z","@version":"1","message":"Failed to send event to dead letter queue topic=knative-broker-default-default partition=0 offset=7","logger_name":"dev.knative.eventing.kafka.broker.dispatcher.RecordDispatcher","thread_name":"vert.x-eventloop-thread-1","level":"ERROR","level_value":40000,"stack_trace":"io.vertx.core.impl.NoStackTraceThrowable: No DLQ set\n","topic":"knative-broker-default-default","partition":0,"offset":7} 
```

The last error log is quite explicit (No DLQ set) but reading the documentation I haven't been able to find references to setting up a DLQ. In addition to this, I've also checked the kafka topics in case something was missing:
`kubectl exec -it my-cluster-kafka-0 -n kafka -c kafka -- bin/kafka-topics.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --list`

```
__consumer_offsets
__strimzi-topic-operator-kstreams-topic-store-changelog
__strimzi_store_topic
custom-topic
knative-broker-default-default
```

The topics looked as expected:
- `custom-topic` is the topic I have created, that all other elements of the system use for the test.
- `knative-broker-default-default` is the topic that I was expecting the broker to create, and that appears in the error log.

The broker is configured like this:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: kafka-broker-config
  namespace: knative-eventing
data:
  # Number of topic partitions
  default.topic.partitions: "1"
  # Replication factor of topic messages.
  # Broker URL. Replace this with the URLs for your kafka cluster,
  # which is in the format of my-cluster-kafka-bootstrap.my-kafka-namespace:9092.
  bootstrap.servers: my-cluster-kafka-bootstrap.kafka:9092
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: config-br-defaults
  namespace: knative-eventing
data:
  default-br-config: |
    clusterDefault:
      brokerClass: Kafka
      apiVersion: v1
      kind: ConfigMap
      name: kafka-broker-config
      namespace: knative-eventing
    namespaceDefaults:
      namespace1:
        brokerClass: Kafka
        apiVersion: v1
        kind: ConfigMap
        name: kafka-broker-config
        namespace: knative-eventing
      namespace2:
        brokerClass: Kafka
        apiVersion: v1
        kind: ConfigMap
        name: kafka-broker-config
        namespace: knative-eventing 
---
kind: Broker
metadata:
  annotations:
    # case-sensitive
  name: default
  namespace: default
spec:
  # Configuration specific to this broker.
  config:
    apiVersion: v1
    kind: ConfigMap
    name: kafka-broker-config
    namespace: knative-eventing
  # Optional dead letter sink, you can specify either:
  #  - deadLetterSink.ref, which is a reference to a Callable
  #  - deadLetterSink.uri, which is an absolute URI to a Callable (It can potentially be out of the Kubernetes cluster)
  delivery:
    deadLetterSink:
      ref:
        apiVersion: serving.knative.dev/v1
        kind: Service
        name: event-deadletter-sink
    backoffDelay: PT2S  # Initial 2 second delay
    backoffPolicy: exponential
    retry: 3
```

pierDipi:
The resolver uses `ref.namespace` to lookup the destination
addressable.
`ref.namespace` for the `deadLetterSink` is not defaulted by the
webhook like for `trigger.subscriber.ref`, so we might receive a
destination with an empty string for
`deadLetterSink.ref.namespace` which leads to not finding the
referenced addressable service.

Fixes #1011 

## Proposed Changes

- Add namespace to referenced addressable before resolving destination

**Release Note**

```release-note
The fields `Broker.spec.delivery.deadLetterSink.ref.namespace` and `Trigger.spec.delivery.deadLetterSink.ref.namespace` are optional.
By default the controller uses the object namespace. 
```

### DIFF
--- a/control-plane/pkg/core/config/utils.go
+++ b/control-plane/pkg/core/config/utils.go
@@ -62,6 +62,10 @@ func EgressConfigFromDelivery(
 	egressConfig := &contract.EgressConfig{}
 
 	if delivery.DeadLetterSink != nil {
+		destination := *delivery.DeadLetterSink // Do not update object Spec, so copy destination.
+		if destination.Ref != nil && destination.Ref.Namespace == "" {
+			destination.Ref.Namespace = parent.GetNamespace()
+		}
 		deadLetterSinkURL, err := resolver.URIFromDestinationV1(ctx, *delivery.DeadLetterSink, parent)
 		if err != nil {
 			return nil, fmt.Errorf("failed to resolve Spec.Delivery.DeadLetterSink: %w", err)
--- a/control-plane/pkg/reconciler/broker/broker_test.go
+++ b/control-plane/pkg/reconciler/broker/broker_test.go
@@ -218,6 +218,63 @@ func brokerReconciliation(t *testing.T, format string, configs Configs) {
 			},
 		},
 		{
+			Name: "Reconciled normal - with DLS - no DLS ref namespace",
+			Objects: []runtime.Object{
+				NewBroker(
+					WithDelivery(WithNoDeadLetterSinkNamespace),
+				),
+				NewConfigMapFromContract(&contract.Contract{
+					Generation: 1,
+				}, &configs),
+				NewService(WithServiceNamespace(BrokerNamespace)),
+				BrokerReceiverPod(configs.SystemNamespace, map[string]string{base.VolumeGenerationAnnotationKey: "3"}),
+				BrokerDispatcherPod(configs.SystemNamespace, map[string]string{base.VolumeGenerationAnnotationKey: "1"}),
+			},
+			Key: testKey,
+			WantEvents: []string{
+				finalizerUpdatedEvent,
+			},
+			WantUpdates: []clientgotesting.UpdateActionImpl{
+				ConfigMapUpdate(&configs, &contract.Contract{
+					Resources: []*contract.Resource{
+						{
+							Uid:              BrokerUUID,
+							Topics:           []string{BrokerTopic()},
+							Ingress:          &contract.Ingress{ContentMode: contract.ContentMode_BINARY, IngressType: &contract.Ingress_Path{Path: receiver.Path(BrokerNamespace, BrokerName)}},
+							BootstrapServers: bootstrapServers,
+							EgressConfig:     &contract.EgressConfig{DeadLetter: ServiceURLFrom(BrokerNamespace, ServiceName)},
+						},
+					},
+					Generation: 2,
+				}),
+				BrokerReceiverPodUpdate(configs.SystemNamespace, map[string]string{
+					base.VolumeGenerationAnnotationKey: "2",
+				}),
+				BrokerDispatcherPodUpdate(configs.SystemNamespace, map[string]string{
+					base.VolumeGenerationAnnotationKey: "2",
+				}),
+			},
+			WantPatches: []clientgotesting.PatchActionImpl{
+				patchFinalizers(),
+			},
+			WantStatusUpdates: []clientgotesting.UpdateActionImpl{
+				{
+					Object: NewBroker(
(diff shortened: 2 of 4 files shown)


aliok:
@pierDipi 
Have you checked if the namespace is optional in the spec?
If so, is it documented that it defaults to parent object namespace?

pierDipi:
The spec doesn't explicitly talk about this field being optional or the default value (I'll open an issue).

The namespace for a `KReference` is (implementation-wise) optional and we historically defaulted it to the object namespace, for example in Trigger subscriber [1] (there are many examples).

That is also the logically expected behavior of our users as you can see in the linked issue.

I opened an issue in eventing to add this behavior https://github.com/knative/eventing/issues/5747 (and a PR).

1. https://github.com/knative/eventing/blob/f031ba23b23d20ed8d13491000feb0abcf50d562/pkg/apis/eventing/v1/trigger_defaults.go#L40-L41

---

Regarding this PR, I'd keep, regardless of the outcome of https://github.com/knative/eventing/issues/5747, the logic of this PR to make sure we don't pass invalid data to the resolver that requires a namespace to be set.

aliok (approved):
/lgtm
/approve

Thanks @pierDipi for fixing this issue and also taking care of the same issue in other places

pierDipi:
`No tests were recorded.`  🤔 ❓ ❓ 
/retest

PR #1281, opened 2021-10-05
[minor] Use Maven wrapper to run unit tests

pierDipi:
It's better to use the maven wrapper since it doesn't require a maven installation.

## Proposed Changes

- Use Maven wrapper to run unit tests


matzew (approved):
/lgtm
/approve

PR #1282, opened 2021-10-05
[WIP] Custom eventshub to support kafka consumer 

pierDipi:
Add a new custom eventshub that can perform all the eventshub functionalities and can also be used as kafka consumer 

## Proposed Changes

- :gift: Add a new eventshub that can replace our existing kafka test images, including options and asserts


pierDipi:
I'm going to close this, for now, we can use KafkaSink and KafkaSource to mimic Kafka consumers and producers.

Issue #785, opened 2021-04-07
Consumers partitioning and Trigger parallelism


pierDipi:
**Problem**

Design doc: https://docs.google.com/document/d/1L-gWDV2sCaSXYikEEHIgn3KY11YxExYvjcslpoh1Pj4/edit

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?
- Event Consumer

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

15

pierDipi:
Tracked in: https://github.com/knative-sandbox/eventing-kafka-broker/issues/1537

PR #1293, opened 2021-10-07
Enhancement: Consumers Scheduling

pierDipi:
## PoC status

- not yet working
- branch: https://github.com/pierDipi/eventing-kafka-broker/tree/poc-scheduling
  - Important files to look at:
    - trigger_v2.go
    - trigger_v2_pod_controller.go
    - trigger_v2_vpod.go
    - controller_v2.go

### DIFF
--- a/docs/enhancements/consumers-scheduling.md
+++ b/docs/enhancements/consumers-scheduling.md
@@ -99,8 +99,6 @@ To mark a resource, identified by `<resource_namespace>/<resource_name>`, as rea
    `kafka.eventing.knative.dev/scheduler.placeable.Placements.<resource_namespace>/<resource_name> >= resource.status.ObservedGeneration`
 1. To make sure that there are no further changes to the scheduled resource, we need to check that: `metadata.generation == resource.status.ObservedGeneration`
 
-In other words, the `Pod` reconciler has seen the `Trigger` reconciler scheduling decision.
-
 ## Known Issues and TODOs:
 
 - Current scheduler implementation in `eventing-kafka` depends on `StatefulSet`
@@ -111,12 +109,3 @@ In other words, the `Pod` reconciler has seen the `Trigger` reconciler schedulin
 - KafkaChannel:
     - How to embed `N` placeables one per subscription in `KafkaChannel` status
     - How to get `Subscription` replicas
-
-## PoC status
-
-- branch: https://github.com/pierDipi/eventing-kafka-broker/tree/poc-scheduling
-  - Important files to look at:
-    - trigger_v2.go
-    - trigger_v2_pod_controller.go
-    - trigger_v2_vpod.go
-    - controller_v2.go


docs/enhancements/consumers-scheduling.md
```diff
@@ -0,0 +1,111 @@
...
+1. Transform them to the internal contract representation
+1. Save the contract in the ConfigMap associtated with the pod
+1. Update pod annotations as follow:
+    ```yaml
+    apiVersion: v1
+    kind: Pod
+    metadata:
+    annotations:
+        # <resource_namespace>/<resource_name> represents the resource key
+        kafka.eventing.knative.dev/scheduler.placeable.Placements.<resource_namespace>/<resource_name>: <resource.status.ObservedGeneration>
```
devguyio, line 86: How will this look like if there are a significant number of Triggers, specifically:
1. The UX: Are we okay for cluster admins dealing with a pod that has 100+ annotations? 
2. Are we gonna hit any [etcd limits](https://etcd.io/docs/v3.5/dev-guide/limit/#request-size-limit)?

devguyio (commented):
Interesting proposal! Here're my thoughts:
1. I've some concerns/questions about the use of annotations that I left in the comments.
2. On a macro-level: this design introduces:
    1.  a place for providing a data plane runtime configuration.
    2. a control plane - data plane coordination mechanism. 

 Given that we already need to address the control plane - data plane communication and probably might need to set our runtime configuration story straight, here are some questions:

1. Would adopting yet another narrow scope solution for cp-dp config & coordination impose future constraints when solving the cp-dp problem?
2. Are we confident this solution will still hold when we address the cp-dp config & coordination problem? or is there a possibility it could be designed differently had we solved that first?

docs/enhancements/consumers-scheduling.md
pierDipi: > How will this look like if there are a significant number of Triggers, specifically:
> 
> 1. The UX: Are we okay for cluster admins dealing with a pod that has 100+ annotations?

In general, I don't think that would be a problem, given that an annotation is not indexed while labels are, so it's comparable to additional fields of the spec. 

Besides that, I'm thinking that the best thing you want to do in Kubernetes is to have "low" resource limits and requests (whatever that means) for each pod and scale them out (if the application allows that) in order to take advantage to the high-density scheduling of the k8s pods, which leads to a lower number of annotations per pod.

> 2. Are we gonna hit any [etcd limits](https://etcd.io/docs/v3.5/dev-guide/limit/#request-size-limit)?

In the worst-case, in addition to other pod's fields, the `kafka.eventing.knative.dev/scheduler.placeable.Placements.<resource_namespace>/<resource_name>` (when variables are 63 chars) is 740 bytes x 100 = 74000 bytes, which shouldn't be a problem for Etcd or the API server.

pierDipi:
> > Thanks for your review @devguyio!
> > > Interesting proposal! Here're my thoughts:
> > > 
> > > 1. I've some concerns/questions about the use of annotations that I left in the comments.
> > > 2. On a macro-level: this design introduces:
> > >    
> > >    1. a place for providing a data plane runtime configuration.
> > 
> > 
> > If you're referring to configuring replicas for a `Trigger` or `KafkaSource` replicas, I'd say it is more a control plane configuration than a data plane one.
> 
> Sorry I guess I wasn't very accurate in this "this design introduces" . I am referring to the end result, the end-to-end chain, or the intention if you may. I understand this design on its own introduces more a scheduling configuration that'd be translated by the control plane to the update of the ConfigMap. But the "intention" is that end-to-end, this introduces a new API that would result eventually in changing the data plane runtime configuration.
> 
> > > 1. a control plane - data plane coordination mechanism.
> > 
> > 
> > The coordination mechanism is between 2 control plane components (Resource reconciler and Pod reconciler), the coordination mechanism between the control plane and the data plane is still the old one (ConfigMaps).
> 
> Ack, again poor choice of "introduces" from my side, But the intent is to have a controller-controllers-dispatcher coordination more or less. no? Scheduler-Resource ctor - Pod ctor - dp reread config - pod ctor - resource ctor.
> 
> > In addition, I'm decreasing the importance of the CM being the mechanism
> 
> Ack, and introducing at the same time another configuration mechanism. Which I'll reference later down.
> 
> > since with this design the CM is coupled with a single Pod, and its short-lived lifecycle, while before it was coupled to 2 long-lived Deployments, so now it becomes an ephemeral component for the dispatcher, that could be replaced in the future.
> 
> > > Given that we already need to address the control plane - data plane communication and probably might need to set our runtime configuration story straight, here are some questions:
> > > 
> > > 1. Would adopting yet another narrow scope solution for cp-dp config & coordination impose future constraints when solving the cp-dp problem?
> > > 2. Are we confident this solution will still hold when we address the cp-dp config & coordination problem? or is there a possibility it could be designed differently had we solved that first?
> > 
> > 
> > Given the above point (2), and with the recent changes to consumer groups initialization [knative-sandbox/eventing-kafka#911](https://github.com/knative-sandbox/eventing-kafka/pull/911) (TLDR; remove direct coordination between controller and data plane, and use Kafka instead), I think this will not be a problem _on the dispatcher side_ (which is what we're changing in this proposal).
> 
> So I guess the gist of how I see this now is like this:
> 
> This introduces: **1. A scheduler API for assignment / sharding of resources to data plane pods.** I generally see this is very valuable, and we 100% need that. I don't yet have a strong opinion on that specific solution and need to take a deeper look at it and whether or not there were alternative solutions. Would be awesome if any of y'all have more info to help here.
> 

I created an alternative solution a while ago that is different from this but I was never convinced it is the approach I'd move forward (https://github.com/knative-sandbox/eventing-kafka-broker/issues/785)

> **2. A mechanism to reflect the assignment/shard on the ConfigMaps that includes a reload of the dispatcher pods when updated (i.e annotations)** My only concern here is the admin UX as described above. Specifically the dynamic, growing aspect of it. Would there be an alternative that solves this dynamic growing aspect?

Again let's not focus too much on the ConfigMap aspect, we can change that later easily.
The annotation is a mechanism to coordinate 2 controllers, _then_ to refresh the volume.
Also, this seems relevant https://github.com/knative-sandbox/eventing-kafka-broker/pull/1293#discussion_r724801214

devguyio:
> > **2. A mechanism to reflect the assignment/shard on the ConfigMaps that includes a reload of the dispatcher pods when updated (i.e annotations)** My only concern here is the admin UX as described above. Specifically the dynamic, growing aspect of it. Would there be an alternative that solves this dynamic growing aspect?
> 
> Again let's not focus too much on the ConfigMap aspect, we can change that later easily.
> The annotation is a mechanism to coordinate 2 controllers (status updates and requeuing), _then_ to refresh the volume.

TBH I'm not sure exactly where is the focus on the ConfigMap 😀 I'm focusing on the dynamically growing list of annotations.

> Also, this seems relevant https://github.com/knative-sandbox/eventing-kafka-broker/pull/1293#discussion_r724801214
> 

Responded.

docs/enhancements/consumers-scheduling.md
pierDipi: > > > > In general, I don't think that would be a problem, given that an annotation is not indexed while labels are, so it's comparable to additional fields of the spec.
> > > 
> > > 
> > > Can you elaborate more why is annotation indexing relevant here?
> > 
> > 
> > It's not that relevant because it's not done. (forget about what I wrote :) )
> > > > Besides that, I'm thinking that the best thing you want to do in Kubernetes is to have "low" resource limits and requests (whatever that means) for each pod and scale them out (if the application allows that) in order to take advantage to the high-density scheduling of the k8s pods, which leads to a lower number of annotations per pod.
> > > 
> > > 
> > > This is still assuming a specific user behavior, the question is, given that we're using this in a serverless environment where scale to 0 of subscribers is assumed, is the use case of say two dispatcher pods handling 1K triggers totally out of question here? and how would the admin UX be like in that case?
> > 
> > 
> > If you scale all triggers to 0, we will have 0 scheduling-relevant annotations on both pods.
> 
> Is there a misunderstanding here? I'm referring to scaling of the _subscriber_ of the trigger, not the trigger itself.

Sorry for the misunderstanding, but then my comment below is the answer.

> 
> > We have a new annotation for each trigger that has `replicas > 0`.
> > Assuming that all 1k triggers have `replicas > 0` and we have only 2 pods, then yes 500 annotations per pod.
> > (2 pods for 1k Kafka consumers is definitely low)
> 
> I'm interested honestly to know more concrete numbers or sizing guides behind that, care to elaborate more?
> 

I don't have concrete numbers but scheduling 500 consumers in a pod means that if one of them is rescheduled you will get a rebalance across multiple consumer groups (when there are more consumers for a group in the same pod the rebalance could even be cascading (one join the group, then another one join group, and so on) which leads to a lot of spike traffic to Kafka and message processing delays since joining a group is a stateful operation that needs to be coordinated.

Don't you see an underprovisioned environment in running 500 Kafka consumers in a single process?

> > I think that since those are admin decisions, an admin needs to know how the system works to best operate it, so I'd say spelling out in docs that the preference is for small pods over big pods is a reasonable solution IMHO.
> 
> I can totally be convinced with that when it's unavoidable, but is having a dynamically growing list of annotations in the pod unavoidable? I think that's what we should focus on. If it is, sure let's fall back to this solution.

A max trigger per pod configuration could be introduced but that is not relevant for scheduling that is an autoscaling problem because once we have reached the capacity of a pod we need to add more pods (or remove) and then the scheduler will do its own things as described here.

pierDipi:
> This introduces: 1. A scheduler API for assignment / sharding of resources to data plane pods. I generally see this is very valuable, and we 100% need that. I don't yet have a strong opinion on that specific solution and need to take a deeper look at it and whether or not there were alternative solutions. Would be awesome if any of y'all have more info to help here.
> > I created an alternative solution a while ago that is different from this but I was never convinced it is the approach I'd move forward (#785)

Also, I explored another idea that was using Kafka itself to do the sharding (I can't find the document) but the limitation is that you need some capacity planning beforehand which is not a good UX.

docs/enhancements/consumers-scheduling.md
pierDipi: > > I don't have concrete numbers but scheduling 500 consumers in a pod means that if one of them is rescheduled you will get a rebalance across multiple consumer groups (when there are more consumers for a group in the same pod the rebalance could even be cascading (one join the group, then another one join group, and so on)
> 
> I need to think more of this, quick question, are vertx consumers "verticles" ?
> 

Yes, every KafkaConsumer instance is wrapped in a vertx verticle

> > A max trigger per pod configuration could be introduced but that is not relevant for scheduling that is an autoscaling problem because once we have reached the capacity of a pod we need to add more pods (or remove) and then the scheduler will do its own things as described here.
> 
> Why do we need an annotation entry per trigger?

To coordinate with the `Trigger` reconciler (queue the trigger, verifying the observed generation, etc)

devguyio: > To coordinate with the `Trigger` reconciler (queue the trigger, verifying the observed generation, etc)

I'm sorry I meant on the pod resource, why an annotation entry per trigger?

docs/enhancements/consumers-scheduling.md
```diff
@@ -0,0 +1,111 @@
...
+type Placement struct {
+	// PodName is the name of the pod where the resource is placed
+	PodName string `json:"podName,omitempty"`
+	// VReplicas is the number of virtual replicas assigned to in the pod
+	VReplicas int32 `json:"vreplicas,omitempty"`
+}
+```
+
+The `KafkaSource` type embeds the `Placeable` type in the status, while the Trigger will use
+a different representation of that type in the `status.annotations` field like the following:
```
devguyio, line 33: will the Trigger then be implementing the duck type? i.e any code dealing with `Placeable` can still deal with the Triggers?

lionelvillard: /cc @aavarghese

aavarghese: @devguyio since Placements info will be added as Annotations for Trigger (and could do the same for KafkaSource, as well), don't think Trigger will implement this duck type. Pod reconcilers dealing with these placements use them independent of resource being a Trigger or Kafkasource. (Apologize if I misunderstood what I'm learning about the scheduling proposal)

PR #1306, opened 2021-10-12
WIP: Remove config-tracing, since already there via eventing

matzew:

Fixes #

## Proposed Changes

- Remove empty `config-tracing`, which is already there via eventing core, so would override that (potentially)

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/hack/update-checksums.sh
+++ b/hack/update-checksums.sh
@@ -26,4 +26,4 @@ fi
 
 source $(dirname $0)/../vendor/knative.dev/hack/library.sh
 
-go run "${REPO_ROOT_DIR}/vendor/knative.dev/pkg/configmap/hash-gen" "${REPO_ROOT_DIR}"/control-plane/config/100-config-tracing.yaml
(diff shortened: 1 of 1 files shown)


hack/update-checksums.sh
```diff
@@ -26,4 +26,4 @@ fi
 
 source $(dirname $0)/../vendor/knative.dev/hack/library.sh
 
-go run "${REPO_ROOT_DIR}/vendor/knative.dev/pkg/configmap/hash-gen" "${REPO_ROOT_DIR}"/control-plane/config/100-config-tracing.yaml
+#go run "${REPO_ROOT_DIR}/vendor/knative.dev/pkg/configmap/hash-gen" "${REPO_ROOT_DIR}"/control-plane/config/100-config-tracing.yaml
```
matzew, line 29: I think we can either remove the file or just remove this line

matzew:
@pierDipi Since this file gets installed via `eventing`, not sure we need to have the duplicate on ours here

matzew:
Ok, so I'd think we fine-tweak the actual yaml that get's produced? 

E.g. broker VS sink

pierDipi:
Yes, we're fine as long as `config-tracing` is in the system namespace (the data plane doesn't start without config-tracing in the right place)

Issue #312, opened 2020-10-19
Implement KafkaSource controller


slinkydeveloper:
**Problem**
We need to implement the KafkaSource controller for the data plane in this repo

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for? Users

**Exit Criteria**
A working kafka source implementation using the data-plane in this project. All e2e tests of kafka source passing.

**Time Estimate (optional):** 
10

matzew:
Interesting, could be aligned w/ the `eventing-kafka` repo? or is that competing ?

slinkydeveloper:
The idea is to reuse the data plane we have here and build the control plane for it. I guess it's mostly competing, because by the nature of this data plane, this implementaiton will be a "multi-tenant" kafka source

lionelvillard:
did you say multi-tenant, as of 1 data-plane component for all source instances? @slinkydeveloper ?

slinkydeveloper:
@lionelvillard yep that's the idea

lionelvillard:
Of course :-) Do you have a design document describing how the scaling works? I'm assuming it's using Keda, but I'm most curious to know how source instances scale independently to each other. @slinkydeveloper

pierDipi:
Roughly the rough idea is to:
- add the number of consumers to run for each KafkaSource or Trigger (Egress in our terminology)
  - how?
     - KafkaSource: scale subresource (your PR) or an annotation
     - Trigger: an internal CRD with the scale subresource or an annotation
- use the `RangeAssignor` to partition consumer instances (your design document) + defining a total order for each Egress config so that each pod has the same view of the system.

lionelvillard:
ah cool thanks!

PR #1326, opened 2021-10-15
Add Key deserializer for KafkaSource

pierDipi:
This patch adds a `KeyDeserializer` that uses the
`kafkasources.sources.knative.dev/key-type` label to
deserialize a record key.

Part of #312

## Proposed Changes

- Add Key deserializer for KafkaSource

**Release Note**

```release-note
None
```

**Docs**

`None`

### DIFF
--- a/control-plane/pkg/reconciler/source/source.go
+++ b/control-plane/pkg/reconciler/source/source.go
@@ -185,6 +185,7 @@ func (r *Reconciler) getResource(ctx context.Context, ks *sources.KafkaSource, s
 		EgressConfig:  egressConfig,
 		DeliveryOrder: DefaultDeliveryOrder,
 	}
+	// Set key type hint (if any).
 	if keyType, ok := ks.Labels[sources.KafkaKeyTypeLabel]; ok {
 		egress.KeyType = coreconfig.KeyTypeFromString(keyType)
 	}
--- a/data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/KeyDeserializerTest.java
+++ b/data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/KeyDeserializerTest.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright © 2018 Knative Authors (knative-dev@googlegroups.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package dev.knative.eventing.kafka.broker.dispatcher.impl.consumer;
 
 import com.google.common.base.Charsets;
--- a/control-plane/pkg/core/config/egress.go
+++ b/control-plane/pkg/core/config/egress.go
@@ -65,6 +65,7 @@ func AddOrUpdateEgressConfig(ct *contract.Contract, brokerIndex int, egress *con
 	return EgressChanged
 }
 
+// KeyTypeFromString returns the contract.KeyType associated to a given string.
 func KeyTypeFromString(s string) contract.KeyType {
 	switch s {
 	case "byte-array":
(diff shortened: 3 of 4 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/RecordDispatcher.java
```diff
@@ -32,6 +32,6 @@
    * @param record record to handle.
    * @return the completion future.
    */
-  Future<Void> dispatch(KafkaConsumerRecord<String, CloudEvent> record);
+  Future<Void> dispatch(KafkaConsumerRecord<Object, CloudEvent> record);
```
matzew, line 35: is this _break_ change? (well String is Object...)

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/InvalidCloudEventInterceptor.java
```diff
@@ -61,6 +61,8 @@
 
   public static final String TYPE = "dev.knative.kafka.event";
 
+  private static final String KEY_EXTENSION = "key";
```
matzew, line 64: can we - somewhere, get a little comment/description on the relationship between `PARTITION_KEY_EXTENSION` and `KEY_EXTENSION` ?

control-plane/pkg/reconciler/source/source.go
```diff
@@ -185,6 +185,10 @@ func (r *Reconciler) getResource(ctx context.Context, ks *sources.KafkaSource, s
 		EgressConfig:  egressConfig,
 		DeliveryOrder: DefaultDeliveryOrder,
 	}
+	// Set key type hint (if any).
+	if keyType, ok := ks.Labels[sources.KafkaKeyTypeLabel]; ok {
```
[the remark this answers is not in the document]
pierDipi: I am open to making this change, but why not a proper spec field?
```
spec:
  key:
    type: <int | string | byte-array | float>
```

A spec field has many benefits over an annotation like validation since if I mistype the annotation key or label key I won't get any error or warning.

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/Main.java
```diff
@@ -64,9 +66,11 @@ public static void main(final String[] args) throws IOException {
 
     // Read consumer and producer kafka config
     Properties producerConfig = Configurations.readPropertiesSync(env.getProducerConfigFilePath());
+    producerConfig.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName());
```
matzew, line 69: why did we have this before? B/c was just string - no SerDe needed?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/InvalidCloudEventInterceptor.java
pierDipi: The only reason to use `key` is for backward compatibility with the exiting source in `eventing-kafka`.

I'd propose to deprecate that field if possible.

data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/KeyDeserializerTest.java
```diff
@@ -0,0 +1,168 @@
...
+    assertThat(got).isEqualTo(new DoubleDeserializer().deserialize("t1", data));
+  }
+
+  @Test
+  public void shouldDeserializeFloat() {
+    final var deserializer = new KeyDeserializer();
+
+    final var configs = new HashMap<String, Object>();
+    configs.put(KEY_TYPE, DataPlaneContract.KeyType.Double);
+    deserializer.configure(configs, true);
```
matzew, line 98: should we name it `keyDeserializer` ? the configure alreay also has a flag that this is for the `key` of the kafka-record (the `true`)

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/Main.java
pierDipi: We got this config from the CM with all producer configurations but now I want to force it to be `String` so that no one can mess with it.

control-plane/pkg/reconciler/source/source.go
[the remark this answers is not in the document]
matzew: Oh, that's an interesting point, making this part of the spec 

I don't actually know why... I guess less "heavy" to do as label/annotation.

But sounds reasonable to me...  but we should keep the existing `label`, and just at some point "migrate" 

At least my $0.02

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/InvalidCloudEventInterceptor.java
matzew: > So, I'd deprecate key and the alternative will be partitionkey.

sounds good, @pierDipi

PR #1335, opened 2021-10-15
Reorganize manifests of the control plane

pierDipi:
This patch reorganizes manifests of the control plane in a way
that makes more sense moving forward, in particular:

- it moves webhook-related manifests in a directory called
  `200-webhook` rather than being in a directory called `sink`.
- it moves controller-related manifests in a directory called
  `200-controller` rather than being in the root directory of our
  control plane configuration.
- In the directories `100-broker`, `100-channel`, `100-sink` there
  are files related to the associated component like CRDs, CMs, etc.

Part of #312

## Proposed Changes

- Reorganize manifests of the control plane

**Release Note**

```release-note
None
```

**Docs**

```
None
```

### DIFF
--- a/control-plane/config/100-source/100-kafka-source.yaml
+++ b/control-plane/config/100-source/100-kafka-source.yaml
@@ -1,0 +1,1 @@
+../../../third_party/eventing-kafka-latest/source-crds.yaml


matzew:
wanna do a `100-source` too, that links to the third-party CRD for `KafkaSource` ?

PR #1348, opened 2021-10-19
Update aggregated roles for KafkaChannel

aliok:
Fixes #

## Proposed Changes

- Add rules for addressable-resolver role for KafkaChannel
- Move addressable-resolver YAML to `200-controller` directory
- Add channelable-manipulator rules for KafkaChannel

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/config/200-controller/200-addressable-resolver-cluster-role.yaml
+++ b/control-plane/config/200-controller/200-addressable-resolver-cluster-role.yaml
@@ -15,7 +15,7 @@
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: kafka-addressable-resolver
+  name: knative-kafka-addressable-resolver
   labels:
     kafka.eventing.knative.dev/release: devel
     duck.knative.dev/addressable: "true"
--- a/control-plane/config/200-controller/200-channelable-manipulator-clusterrole.yaml
+++ b/control-plane/config/200-controller/200-channelable-manipulator-clusterrole.yaml
@@ -15,7 +15,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: kafka-channelable-manipulator
+  name: knative-kafka-channelable-manipulator
   labels:
     kafka.eventing.knative.dev/release: devel
     duck.knative.dev/channelable: "true"


control-plane/config/200-controller/200-addressable-resolver-cluster-role.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: kafka-addressable-resolver
+  labels:
+    kafka.eventing.knative.dev/release: devel
+    duck.knative.dev/addressable: "true"
+# Do not use this role directly. These rules will be added to the "addressable-resolver" role.
+rules:
+  - apiGroups:
+      - eventing.knative.dev
+    resources:
+      - kafkasinks
```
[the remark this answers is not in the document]
matzew: should we split this file?

control-plane/config/200-controller/200-addressable-resolver-cluster-role.yaml
```diff
@@ -0,0 +1,42 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: kafka-addressable-resolver
```
pierDipi, line 18: Since this is a `ClusterRole` and it could conflict with other unrelated projects, I'd recommend adding `knative-` as prefix
```suggestion
  name: knative-kafka-addressable-resolver
```

control-plane/config/200-controller/200-addressable-resolver-cluster-role.yaml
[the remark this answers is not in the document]
pierDipi: There is already an addressable resolver for `KafkaSinks` in https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/config/200-webhook/200-addressable-resolver-cluster-role.yaml (I'd change the location to be under `200-controller` rather than `200-webhook`)

control-plane/config/200-controller/200-addressable-resolver-cluster-role.yaml
pierDipi: I noticed we don't do this in the other resources but we should.

aliok: +1 Done

PR #1350, opened 2021-10-19
Remove Trigger finalizer when Broker is removed

pierDipi:
This bug is reproducible in this situation:

- Create Kafka Broker `b1`
- Create a trigger `t1` associated with `b1`
- Remove Kafka Broker `b1`
- Remove Trigger `t1`

`t1` will be in `Terminating` state forever since our
reconciler will not reconcile it since we can't know
if Trigger is our trigger given that there is not `b1`.

## Proposed Changes

- Remove Trigger finalizer when Broker is removed

**Release Note**

```release-note
None
```

**Docs**

```
None
```


matzew (approved):
/lgtm
/approve

PR #1353, opened 2021-10-20
Add support for KafkaSource auth in data plane

pierDipi:
Part of #312

This patch adds support for KafkaSource auth.

KafkaSource auth needs changes to the data plane because it can be
configured to reference multiple secrets for the entire auth context
(TLS, SASL, etc) that we, so far, didn't need.

## Proposed Changes

- Add support for KafkaSource auth in data plane

**Release Note**

```release-note
None
```

**Docs**

```
None
```


pierDipi:
As always when the contract changes, linter complains, please ignore it :smile:

PR #1374, opened 2021-10-27
Kafka Channel Subscription reconciliation - control plane

aliok:
Fixes #

## Proposed Changes

- Reconcile subscribers in the control plane
- Init offsets and mark subscriptions ready
- Data plane changes (the check to see if the offsets are initialized) are to be done in a separate PR. 

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/channel/channel.go
+++ b/control-plane/pkg/reconciler/channel/channel.go
@@ -65,8 +65,9 @@ type Reconciler struct {
 	// mock the function used during the reconciliation loop.
 	NewKafkaClusterAdmin kafka.NewClusterAdminFunc
 
-	// TODO: rename
-	KafkaClient kafka.NewClientFunc
+	// NewKafkaClient creates new sarama Client. It's convenient to add this as Reconciler field so that we can
+	// mock the function used during the reconciliation loop.
+	NewKafkaClient kafka.NewClientFunc
 
 	ConfigMapLister corelisters.ConfigMapLister
 
@@ -157,7 +158,7 @@ func (r *Reconciler) reconcileKind(ctx context.Context, channel *messagingv1beta
 	}
 	defer kafkaClusterAdmin.Close()
 
-	kafkaClient, err := r.KafkaClient(topicConfig.BootstrapServers, kafkaClientSaramaConfig)
+	kafkaClient, err := r.NewKafkaClient(topicConfig.BootstrapServers, kafkaClientSaramaConfig)
 	if err != nil {
 		return statusConditionManager.FailedToCreateTopic(topicName, fmt.Errorf("error getting sarama config: %w", err))
 	}
--- a/control-plane/pkg/reconciler/channel/channel.go
+++ b/control-plane/pkg/reconciler/channel/channel.go
@@ -177,24 +177,29 @@ func (r *Reconciler) reconcileKind(ctx context.Context, channel *messagingv1beta
 	logger.Debug("Topic created", zap.Any("topic", topic))
 	statusConditionManager.TopicReady(topic)
 
-	err = r.reconcileSubscribers(ctx, kafkaClient, kafkaClusterAdmin, channel, statusConditionManager)
-	if err != nil {
-		return fmt.Errorf("error reconciling subscribers %v", err)
-	}
-
-	// Get contract config map to write into it.
+	// Get data plane config map.
 	contractConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx)
 	if err != nil {
-		return statusConditionManager.FailedToGetConfigMap(err)
+		return statusConditionManager.FailedToGetConfig(err)
 	}
 	logger.Debug("Got contract config map")
 
-	// Get contract data
+	// Get data plane config data.
 	ct, err := r.GetDataPlaneConfigMapData(logger, contractConfigMap)
-	if err != nil && ct == nil {
+	if err != nil || ct == nil {
 		return statusConditionManager.FailedToGetDataFromConfigMap(err)
 	}
 	logger.Debug("Got contract data from config map", zap.Any(base.ContractLogKey, ct))
+
+	channelIndex := coreconfig.FindResource(ct, channel.UID)
+	if channelIndex == coreconfig.NoResource {
+		return statusConditionManager.FailedToResolveConfig(errors.New("missing channel in contract"))
+	}
+
+	err = r.reconcileSubscribers(ctx, kafkaClient, kafkaClusterAdmin, channel, statusConditionManager, ct, channelIndex, contractConfigMap)
+	if err != nil {
+		return fmt.Errorf("error reconciling subscribers %v", err)
+	}
 
 	// Get resource configuration
 	channelResource, err := r.getChannelContractResource(ctx, topic, channel, secret, topicConfig)
@@ -204,7 +209,6 @@ func (r *Reconciler) reconcileKind(ctx context.Context, channel *messagingv1beta
 	coreconfig.SetDeadLetterSinkURIFromEgressConfig(&channel.Status.DeliveryStatus, channelResource.EgressConfig)
 
 	// Update contract data with the new contract configuration
-	channelIndex := coreconfig.FindResource(ct, channel.UID)
 	changed := coreconfig.AddOrUpdateResourceConfig(ct, channelResource, channelIndex, logger)
 	logger.Debug("Change detector", zap.Int("changed", changed))
 
@@ -380,41 +384,8 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
 	return nil
 }
 
-func (r *Reconciler) reconcileSubscribers(ctx context.Context, kafkaClient sarama.Client, kafkaClusterAdmin sarama.ClusterAdmin, channel *messagingv1beta1.KafkaChannel, statusConditionManager base.StatusConditionManager) error {
+func (r *Reconciler) reconcileSubscribers(ctx context.Context, kafkaClient sarama.Client, kafkaClusterAdmin sarama.ClusterAdmin, channel *messagingv1beta1.KafkaChannel, statusConditionManager base.StatusConditionManager, ct *contract.Contract, channelIndex int, contractConfigMap *corev1.ConfigMap) error {
 	logger := kafkalogging.CreateReconcileMethodLogger(ctx, channel)
-
-	// Get data plane config map.
-	contractConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx)
-	if err != nil {
-		// In case of errors like this, which is not related,
-		// to a single subscription, we should still set the error
-		// on the KafkaChannel status.
-		// For problems about the subscription, we should set it on the
-		// subscriber.
-		// However, core eventing-controller is setting the status
-		// of individual subscriptions, we cannot set the status of
-		// the subscription itself. We can set it on the
-		// channel.Spec.Subscribers though.
-		return statusConditionManager.FailedToGetConfig(err)
-	}
-
-	logger.Debug("Got contract config map")
-
-	// Get data plane config data.
-	ct, err := r.GetDataPlaneConfigMapData(logger, contractConfigMap)
-	if err != nil || ct == nil {
-		return statusConditionManager.FailedToGetDataFromConfigMap(err)
-	}
-
-	logger.Debug(
-		"Got contract data from config map",
-		zap.Any(base.ContractLogKey, ct),
-	)
-
-	channelIndex := coreconfig.FindResource(ct, channel.UID)
-	if channelIndex == coreconfig.NoResource {
-		return statusConditionManager.FailedToResolveConfig(errors.New("missing channel in contract"))
-	}
 
 	after := channel.DeepCopy()
 	after.Status.Subscribers = make([]v1.SubscriberStatus, 0)
--- a/control-plane/pkg/reconciler/channel/channel.go
+++ b/control-plane/pkg/reconciler/channel/channel.go
@@ -428,7 +428,7 @@ func (r *Reconciler) reconcileInitialOffset(ctx context.Context, channel *messag
 	}
 
 	topicName := topic(TopicPrefix, channel)
-	groupID := fmt.Sprintf("kafka.%s.%s.%s", channel.Namespace, channel.Name, string(sub.UID))
+	groupID := consumerGroup(channel, sub)
 	_, err := offset.InitOffsets(ctx, kafkaClient, kafkaClusterAdmin, []string{topicName}, groupID)
 	if err != nil {
 		logger := logging.FromContext(ctx)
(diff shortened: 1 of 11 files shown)


control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +380,185 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+
+	patch, err := jsonPatch.MarshalJSON()
+	if err != nil {
+		return fmt.Errorf("marshaling JSON patch: %w", err)
+	}
+	patched, err := r.kafkaClientSet.MessagingV1beta1().KafkaChannels(channel.Namespace).Patch(ctx, channel.Name, types.JSONPatchType, patch, metav1.PatchOptions{}, "status")
+
+	if err != nil {
+		return fmt.Errorf("failed patching: %w", err)
+	}
```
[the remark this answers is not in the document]
aliok: I checked the IMC code. There, the dispatcher is responsible for marking the subscription ready/unready and it is done using a patch. This makes sense since the dispatcher is not the owner for the channel and some manual API calls are required.

I guess that code was carried to knative-sandbox/eventing-kafka and I carried it over to here.

I will now remove this code and see what happens.

aliok:
/test pull-knative-sandbox-eventing-kafka-broker-unit-tests

Maven crap

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +375,88 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+
+	topicName := topic(TopicPrefix, channel)
+	groupID := consumerGroup(channel, sub)
+	_, err := offset.InitOffsets(ctx, kafkaClient, kafkaClusterAdmin, []string{topicName}, groupID)
+	if err != nil {
+		logger := logging.FromContext(ctx)
+		logger.Errorw("error reconciling initial offset", zap.String("channel", fmt.Sprintf("%s.%s", channel.Namespace, channel.Name)), zap.Any("subscription", sub), zap.Error(err))
+	}
+	return err
```
pierDipi, line 441: The caller will enrich the error with a bit more context and then it will be logged.
```suggestion
	return err
```

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +375,88 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+}
+
+func (r *Reconciler) getSubscriberConfig(ctx context.Context, channel *messagingv1beta1.KafkaChannel, subscriber *v1.SubscriberSpec) (*contract.Egress, error) {
+	egress := &contract.Egress{
+		Destination:   subscriber.SubscriberURI.String(),
+		ConsumerGroup: string(subscriber.UID),
```
pierDipi, line 447:
```suggestion
		ConsumerGroup: consumerGroup(channel, subscriber),
```

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +375,88 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+		Uid:           string(subscriber.UID),
+	}
+
+	egressConfig, err := coreconfig.EgressConfigFromDelivery(ctx, r.Resolver, channel, subscriber.Delivery, r.Configs.DefaultBackoffDelayMs)
+	if err != nil {
+		return nil, err
+	}
+	egress.EgressConfig = egressConfig
```
pierDipi, line 455: If there is no `subscriber.Delivery`, we should fall back to channel delivery parameters. 
In the trigger reconciler, we do a similar thing where we merge Broker and Trigger delivery params.

https://github.com/knative-sandbox/eventing-kafka-broker/blob/b36793cead0d4c165ad207e6460d2626c207307b/control-plane/pkg/reconciler/trigger/trigger.go#L288-L297

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -175,7 +196,9 @@ func (r *Reconciler) reconcileKind(ctx context.Context, channel *messagingv1beta
 	}
 	coreconfig.SetDeadLetterSinkURIFromEgressConfig(&channel.Status.DeliveryStatus, channelResource.EgressConfig)
 
-	// Update contract data with the new contract configuration
+	r.reconcileSubscribers(ctx, kafkaClient, kafkaClusterAdmin, channel, channelResource)
```
pierDipi, line 199: I'd expect this method call to return an error.

When there is an error, we still want to update the CM but then `ReconcileKind` should return an error so that we retry later.

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +375,88 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+		} else {
+			logger.Debug("marking subscription as ready. ", zap.String("channel", fmt.Sprintf("%s.%s", channel.Namespace, channel.Name)), zap.Any("subscription", s))
+			channel.Status.Subscribers = append(channel.Status.Subscribers, v1.SubscriberStatus{
+				UID:                s.UID,
+				ObservedGeneration: s.Generation,
+				Ready:              corev1.ConditionTrue,
+			})
+		}
+	}
+}
```
pierDipi, line 401:
```suggestion
func (r *Reconciler) reconcileSubscribers(ctx context.Context, kafkaClient sarama.Client, kafkaClusterAdmin sarama.ClusterAdmin, channel *messagingv1beta1.KafkaChannel, channelContractResource *contract.Resource) error {
	logger := kafkalogging.CreateReconcileMethodLogger(ctx, channel)

	channel.Status.Subscribers = make([]v1.SubscriberStatus, 0)
	var globalErr error
	for _, s := range channel.Spec.Subscribers {
		logger = logger.With(zap.Any("subscription", s))
		err := r.reconcileSubscriber(ctx, kafkaClient, kafkaClusterAdmin, channel, s, channelContractResource)
		if err != nil {
			logger.Error("error reconciling subscription. marking subscription as not ready", zap.Error(err))
			channel.Status.Subscribers = append(channel.Status.Subscribers, v1.SubscriberStatus{
				UID:                s.UID,
				ObservedGeneration: s.Generation,
				Ready:              corev1.ConditionFalse,
				Message:            fmt.Sprint("Subscription not ready", err),
			})
			globalErr = multierr.Append(globalErr, err)
		} else {
			logger.Debug("marking subscription as ready")
			channel.Status.Subscribers = append(channel.Status.Subscribers, v1.SubscriberStatus{
				UID:                s.UID,
				ObservedGeneration: s.Generation,
				Ready:              corev1.ConditionTrue,
			})
		}
	}
	return globalErr
}
```

pierDipi: `multierr` is a library we should already have as dependency `go.uber.org/multierr`

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +375,88 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+	subscriberConfig, err := r.getSubscriberConfig(ctx, channel, &subscriberSpec)
+	if err != nil {
+		return fmt.Errorf("failed to resolve subscriber config: %v", zap.Error(err))
+	}
+
+	changed := coreconfig.AddOrUpdateEgressConfigForResource(channelContractResource, subscriberConfig, subscriberIndex)
+
+	logger.Debug("Egress changes", zap.Int("changed", changed))
+
+	return nil
```
pierDipi, line 424: I think this `changed` is always `EgressUnchanged` because `channelContractResource` is always a new object without Egresses set, which is good (so that we don't need to mess with proto unknown fields), so
```suggestion
	coreconfig.AddOrUpdateEgressConfigForResource(channelContractResource, subscriberConfig, subscriberIndex)
	return nil
```

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +375,88 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+	}
+}
+
+func (r *Reconciler) reconcileSubscriber(ctx context.Context, kafkaClient sarama.Client, kafkaClusterAdmin sarama.ClusterAdmin, channel *messagingv1beta1.KafkaChannel, subscriberSpec v1.SubscriberSpec, channelContractResource *contract.Resource) error {
+	logger := kafkalogging.CreateReconcileMethodLogger(ctx, channel)
+
+	logger.Debug("Reconciling initial offset for subscription", zap.Any("subscription", subscriberSpec), zap.Any("channel", channel))
+	err := r.reconcileInitialOffset(ctx, channel, subscriberSpec, kafkaClient, kafkaClusterAdmin)
+	if err != nil {
+		logger.Error("reconcile failed to initial offset for subscription. Marking the subscription not ready", zap.String("channel", fmt.Sprintf("%s.%s", channel.Namespace, channel.Name)), zap.Any("subscription", subscriberSpec), zap.Error(err))
```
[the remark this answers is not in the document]
aliok: Done

control-plane/pkg/reconciler/channel/channel.go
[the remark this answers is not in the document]
aliok: Done

control-plane/pkg/reconciler/channel/channel.go
aliok: Done

control-plane/pkg/reconciler/channel/channel.go
aliok: Done

control-plane/pkg/reconciler/channel/channel.go
aliok: Good point. Done

control-plane/pkg/reconciler/channel/channel.go
aliok: I adapted your suggestion. Thanks

control-plane/pkg/reconciler/channel/channel.go
aliok: Good catch

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +382,90 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+}
+
+func (r *Reconciler) getSubscriberConfig(ctx context.Context, channel *messagingv1beta1.KafkaChannel, subscriber *v1.SubscriberSpec) (*contract.Egress, error) {
+	egress := &contract.Egress{
+		Destination:   subscriber.SubscriberURI.String(),
+		ConsumerGroup: consumerGroup(channel, subscriber),
+		Uid:           string(subscriber.UID),
+	}
+
+	subscriptionEgressConfig, err := coreconfig.EgressConfigFromDelivery(ctx, r.Resolver, channel, subscriber.Delivery, r.Configs.DefaultBackoffDelayMs)
```
aliok, line 454: Passing `channel` as the parent here should work fine I guess. I have doubts if I should pass something else.

pierDipi: Yes, parent is the object that is tracking changes to the referenced object, so I think this is correct.

pierDipi:
@googlebot I consent.

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +382,90 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
 	return nil
 }
 
+func (r *Reconciler) reconcileSubscribers(ctx context.Context, kafkaClient sarama.Client, kafkaClusterAdmin sarama.ClusterAdmin, channel *messagingv1beta1.KafkaChannel, channelContractResource *contract.Resource) error {
+	logger := kafkalogging.CreateReconcileMethodLogger(ctx, channel)
+
+	channel.Status.Subscribers = make([]v1.SubscriberStatus, 0)
+	var globalErr error
+	for _, s := range channel.Spec.Subscribers {
+		logger = logger.With(zap.Any("subscription", s))
```
pierDipi, line 391:
```suggestion
	for _, s := range channel.Spec.Subscribers {
		s := s // capture loop variable
		logger = logger.With(zap.Any("subscription", s))
```

aliok: I did it differently, but thanks for pointing at the problem

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +382,91 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
 	return nil
 }
 
+func (r *Reconciler) reconcileSubscribers(ctx context.Context, kafkaClient sarama.Client, kafkaClusterAdmin sarama.ClusterAdmin, channel *messagingv1beta1.KafkaChannel, channelContractResource *contract.Resource) error {
+	logger := kafkalogging.CreateReconcileMethodLogger(ctx, channel)
+
+	channel.Status.Subscribers = make([]v1.SubscriberStatus, 0)
+	var globalErr error
+	for i, _ := range channel.Spec.Subscribers {
```
pierDipi, line 390:
```suggestion
	for i := range channel.Spec.Subscribers {
```

pierDipi (approved):
/lgtm
/approve

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -135,17 +142,31 @@ func (r *Reconciler) reconcileKind(ctx context.Context, channel *messagingv1beta
...
-	kafkaClusterAdmin, err := r.NewKafkaClusterAdmin(topicConfig.BootstrapServers, saramaConfig)
+	// Manually commit the offsets in KafkaChannel controller.
+	// That's because we want to make sure we initialize the offsets within the controller
+	// before dispatcher actually starts consuming messages.
+	kafkaClientSaramaConfig, err := kafka.GetSaramaConfig(saramaSecurityOption, kafka.DisableOffsetAutoCommitConfigOption)
+	if err != nil {
+		return statusConditionManager.FailedToCreateTopic(topicName, fmt.Errorf("error getting cluster admin sarama config: %w", err))
+	}
+
+	kafkaClusterAdmin, err := r.NewKafkaClusterAdmin(topicConfig.BootstrapServers, kafkaClusterAdminSaramaConfig)
```
matzew, line 158: Perhaps we should name this to `NewKafkaClusterAdminClient` ? I mean the `ClusterAdmin` in Sarama land is the administrative client for Kafka. 

It always confuses me when looking at code related to this (`NewKafkaClusterAdmin` VS `NewKafkaClient`).

I guess upstream saram is named poor there, but perhaps for us we could do the change, and make that `NewKafkaClusterAdminClient` 

I am happy to do that refactors

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -352,6 +382,91 @@ func (r *Reconciler) finalizeKind(ctx context.Context, channel *messagingv1beta1
...
+	subscriptionStatus := findSubscriptionStatus(channel, sub.UID)
+	if subscriptionStatus != nil && subscriptionStatus.Ready == corev1.ConditionTrue {
+		// subscription is ready, the offsets must have been initialized already
+		return nil
+	}
+
+	topicName := topic(TopicPrefix, channel)
+	groupID := consumerGroup(channel, sub)
+	_, err := offset.InitOffsets(ctx, kafkaClient, kafkaClusterAdmin, []string{topicName}, groupID)
+	return err
```
matzew, line 445: some nice reuse :smile:

control-plane/pkg/reconciler/channel/channel.go
aliok: Thanks for https://github.com/knative-sandbox/eventing-kafka-broker/pull/1398

PR #1393, opened 2021-10-29
Remove egresses copy when comparing resources for change detection

pierDipi:
The assumption that `contract.resources[*].egreeses` don't change
when we update a resource in the contract is not true anymore
(and it is not really needed for most of our resources,
only Broker needs that),
so removing the copy.

Part of #312

## Proposed Changes

- Remove egresses copy when comparing resources

**Release Note**

```release-note
None
```

**Docs**

```
None
```

### DIFF
--- a/control-plane/pkg/core/config/resource.go
+++ b/control-plane/pkg/core/config/resource.go
@@ -46,6 +46,12 @@ const (
 	ResourceUnchanged
 )
 
+func SetResourceEgressesFromContract(contract *contract.Contract, resource *contract.Resource, index int) {
+	if index != NoResource {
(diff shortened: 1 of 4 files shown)


control-plane/pkg/core/config/resource.go
```diff
@@ -46,14 +46,19 @@ const (
 	ResourceUnchanged
 )
 
+func SetResourceEgressesFromContract(contract *contract.Contract, resource *contract.Resource, index int) {
+	if index != NoResource {
```
matzew, line 50: can we get a comment here on a follow up PR?

PR #1401, opened 2021-11-03
Add probes cache

pierDipi:
This patch adds a probe cache that allows to associate a `Status`
to a given key.

The cache prunes expired entries and calls a callback when
an entry is expired, so that controllers can re-queue associated
resources.

This is part of our work to reduce e2e tests' flakiness.

## Proposed Changes

- Add probes cache

**Release Note**

```release-note
None
```

**Docs**

```
None
```

### DIFF
--- a/control-plane/pkg/prober/cache.go
+++ b/control-plane/pkg/prober/cache.go
@@ -41,13 +41,13 @@ type Cache interface {
 	GetStatus(key string) Status
 	// UpsertStatus add or updates the status associated with the given key.
 	// Once the given key expires the onExpired callback will be called passing the arg parameter.
-	UpsertStatus(key string, status Status, arg interface{}, onExpired ExpiredFunc)
+	UpsertStatus(key string, status Status, arg interface{}, onExpired ExpirationFunc)
 }
 
-// ExpiredFunc is a callback called once an entry in the cache is expired.
-type ExpiredFunc func(key string, arg interface{})
+// ExpirationFunc is a callback called once an entry in the cache is expired.
+type ExpirationFunc func(key string, arg interface{})
 
-type inMemoryLocalCache struct {
+type localExpiringCache struct {
 	// mu protects targets and entries
 	mu sync.RWMutex
 	// targets is a map of key-pointer to an element in the entries list.
@@ -56,7 +56,7 @@ type inMemoryLocalCache struct {
 	// This allows fast deletion of expired entries.
 	entries *list.List
 
-	expireDuration time.Duration
+	expiration time.Duration
 }
 
 type value struct {
@@ -64,22 +64,22 @@ type value struct {
 	lastUpsert time.Time
 	key        string
 	arg        interface{}
-	onExpired  ExpiredFunc
+	onExpired  ExpirationFunc
 }
 
-func NewInMemoryLocalCache(ctx context.Context, expireDuration time.Duration) Cache {
-	c := &inMemoryLocalCache{
-		mu:             sync.RWMutex{},
-		targets:        make(map[string]*list.Element, 64),
-		entries:        list.New().Init(),
-		expireDuration: expireDuration,
+func NewLocalExpiringCache(ctx context.Context, expiration time.Duration) Cache {
+	c := &localExpiringCache{
+		mu:         sync.RWMutex{},
+		targets:    make(map[string]*list.Element, 64),
+		entries:    list.New().Init(),
+		expiration: expiration,
 	}
 	go func() {
 		for {
 			select {
 			case <-ctx.Done():
 				return
-			case <-time.After(expireDuration):
+			case <-time.After(expiration):
 				c.removeExpiredEntries(time.Now())
 			}
 		}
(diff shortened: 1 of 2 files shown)


aliok:
Looks good, but where's this gonna be used? Like, there's no integration right now

control-plane/pkg/prober/cache.go
```diff
@@ -0,0 +1,134 @@
...
+// Status represents the resource status.
+type Status int
+
+// Cache is a key-status store.
+type Cache interface {
+	// GetStatus retries the status associated with the given key.
+	GetStatus(key string) Status
+	// UpsertStatus add or updates the status associated with the given key.
+	// Once the given key expires the onExpired callback will be called passing the arg parameter.
+	UpsertStatus(key string, status Status, arg interface{}, onExpired ExpiredFunc)
```
matzew, line 44: How about something like `PutIfAbsent` ? :smile:

control-plane/pkg/prober/cache.go
```diff
@@ -0,0 +1,134 @@
...
+
+type value struct {
+	status     Status
+	lastUpsert time.Time
+	key        string
+	arg        interface{}
+	onExpired  ExpiredFunc
+}
+
+func NewInMemoryLocalCache(ctx context.Context, expireDuration time.Duration) Cache {
```
matzew, line 70: `expireDuration`  -> `expiration` ?

control-plane/pkg/prober/cache.go
```diff
@@ -0,0 +1,134 @@
...
+type inMemoryLocalCache struct {
+	// mu protects targets and entries
+	mu sync.RWMutex
+	// targets is a map of key-pointer to an element in the entries list.
+	targets map[string]*list.Element
+	// entries is a doubly-linked list of all entries present in the cache.
+	// This allows fast deletion of expired entries.
+	entries *list.List
+
+	expireDuration time.Duration
```
matzew, line 59: expireDuration -> expiration ?

control-plane/pkg/prober/cache.go
```diff
@@ -0,0 +1,134 @@
...
+type Cache interface {
+	// GetStatus retries the status associated with the given key.
+	GetStatus(key string) Status
+	// UpsertStatus add or updates the status associated with the given key.
+	// Once the given key expires the onExpired callback will be called passing the arg parameter.
+	UpsertStatus(key string, status Status, arg interface{}, onExpired ExpiredFunc)
+}
+
+// ExpiredFunc is a callback called once an entry in the cache is expired.
+type ExpiredFunc func(key string, arg interface{})
```
matzew, line 48: `ExpirationFunc`?

control-plane/pkg/prober/cache.go
```diff
@@ -0,0 +1,134 @@
...
+
+type inMemoryLocalCache struct {
```
matzew, line 50: I'd call that different, and express that the cache has some built-in `expiration`

`localExpiringCache` ? 

Do we really need the "inMemory" in the name?  If so: `inMemoryLocalExpiringCache` ?

control-plane/pkg/prober/cache.go
devguyio: isn't that typically called `ttl` ?

matzew: not typically. I like `expiration` over `ttl`. I think expiring is _stronger_ than `ttl`

(heavily biased based on heavy usage of a sweet library like https://github.com/jhalterman/expiringmap - which this PR is bascially also trying to do)

pierDipi:
> Looks good, but where's this gonna be used? Like, there's no integration right now

Integration in a separate PR.

This cache will be added as a reconciler field and used in `ReconcilerKind` and `FinalizeKind`.

control-plane/pkg/prober/cache.go
pierDipi: Should it be just `put`?
This is doing a `replace`

control-plane/pkg/prober/cache.go
pierDipi: Done, `localExpiringCache`

control-plane/pkg/prober/cache.go
pierDipi: Done!

control-plane/pkg/prober/cache.go
pierDipi: I'm using expiration, for now, I think we all get the meaning of this variable with that name.

control-plane/pkg/prober/cache.go
pierDipi: Done

control-plane/pkg/prober/cache.go
pierDipi: `PutStatus` works for me, even though I think `UpsertStatus` conveys better the idea of adding or replacing something.

matzew (approved):
/lgtm
/approve

thanks for addressing the comments. 

Perhaps even worth to extract as a lib :-)

PR #1402, opened 2021-11-03
Check verticle exception handler presence

pierDipi:
The exception handler may be null.

Part of #312

## Proposed Changes

- Check verticle exception handler presence

**Release Note**

```release-note
None
```

**Docs**

```
None
```


matzew (approved):
/lgtm
/approve

PR #1414, opened 2021-11-05
Remove unicode control characters in Strimzi operator YAML

pierDipi:
The diff won't make sense but there is a non-visible character in
between `_<hostname>_:` and `_<port>_`.

## Proposed Changes

- Remove unicode control characters in Strimzi operator YAML

**Release Note**

```release-note
None
```

**Docs**

```
None
```

### DIFF
--- a/test/kafka/strimzi-cluster-operator.yaml
+++ b/test/kafka/strimzi-cluster-operator.yaml
@@ -11094,7 +11094,7 @@ spec:
               bootstrapServers:
                 type: string
                 description: Bootstrap servers to connect to. This should be given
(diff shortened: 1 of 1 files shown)


matzew:
/lgtm
/approve

While we are removing it here. Should we follow up on this, by filing an issue directly for Strimzi - So that folks like @ppatierno, @tombentley or @scholzj see it too ?

matzew (approved):

scholzj:
@matzew I already removed it in strimzi/strimzi-kafka-operator#5839. So it should be removed in 0.27 release.

PR #1419, opened 2021-11-05
Better naming and consistency for config.Env and broker.Config

aliok:
## Proposed Changes

- Better naming and consistency for `config.Env`
- Got rid of `broker.Config` which only had 1 additional field. That `broker.Config` thing was used in source, channel and sink. Changed those usages to `config.Env`.
-

**Release Note**

```release-note

```

**Docs**


pierDipi (approved):
/lgtm

Issue #1421, opened 2021-11-06
Need to upgrade Sarama To v1.30.0


odacremolbap:
**Describe the bug**

Eventing e2e is failing due to dependency updates resolving Sarama to v1.30.0.
See sibling issue at eventing-kafka:

- workaround: https://github.com/knative-sandbox/eventing-kafka/pull/972
- fix: https://github.com/knative-sandbox/eventing-kafka/pull/975

That same upgrade should also happen at eventing-kafka-broker. This is one of the eventing PR logs showing the issue:

https://github.com/knative/eventing/runs/4120996132?check_suite_focus=true

```
# knative.dev/eventing-kafka-broker/control-plane/pkg/kafka [knative.dev/eventing-kafka-broker/control-plane/pkg/kafka.test]
control-plane/pkg/kafka/consumer_group_lag_test.go:136:117: cannot use tc.client (type *saramaClientMock) as type sarama.ClusterAdmin in return argument:
	*saramaClientMock does not implement sarama.ClusterAdmin (missing AlterClientQuotas method)
ok  	knative.dev/eventing-kafka-broker/control-plane/pkg/core/config	0.002s
FAIL	knative.dev/eventing-kafka-broker/control-plane/pkg/kafka [build failed]
?   	knative.dev/eventing-kafka-broker/control-plane/pkg/logging	[no test files]
ok  	knative.dev/eventing-kafka-broker/control-plane/pkg/receiver	0.001s
# knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/kafka/testing
control-plane/pkg/reconciler/kafka/testing/admin_mock.go:27:5: cannot use &MockKafkaClusterAdmin{} (type *MockKafkaClusterAdmin) as type sarama.ClusterAdmin in assignment:
	*MockKafkaClusterAdmin does not implement sarama.ClusterAdmin (missing AlterClientQuotas method)
ok  	knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/base	0.001s
FAIL	knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/broker [build failed]
ok  	knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/channel	0.003s
FAIL	knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/kafka [build failed]
FAIL	knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/sink [build failed]
[...]

PR #1424, opened 2021-11-07
Pin sarama version to v1.29.1

odacremolbap:
Fixes https://github.com/knative-sandbox/eventing-kafka-broker/issues/1421

Workarond sticks to Samara v1.29.1 so that Knative tests do not fail.
This is not intended to be permanent and should be removed when a PR akin to [this](https://github.com/knative-sandbox/eventing-kafka/pull/975) gets merged.

There is also an update of the kn/hack repo which I don't think is strictly needed here, but will resolve tags from `knative-v1.0.0` and on for any test that needs it.

## Proposed Changes

- Pin the samara version to v1.29.1. Will need to be unpinned when it is upgraded at this repo.


pierDipi:
Hi @odacremolbap, I sent a PR to bump Sarama to 1.30 https://github.com/knative-sandbox/eventing-kafka-broker/pull/1435

odacremolbap:
Thanks so much @pierDipi 
Closing this.

PR #1438, opened 2021-11-09
Set KafkaChannel delivery order to ordered by default

pierDipi:
The existing channel uses ordered delivery and I think we should
keep it ordered by default for the new channel for backward
compatibility and seamless upgrade.

## Proposed Changes

- Set KafkaChannel delivery order to ordered by default

**Release Note**

```release-note
None
```

**Docs**

```
None
```


control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -447,6 +448,7 @@ func (r *Reconciler) getSubscriberConfig(ctx context.Context, channel *messaging
 	egress := &contract.Egress{
 		Destination:   subscriber.SubscriberURI.String(),
 		ConsumerGroup: consumerGroup(channel, subscriber),
+		DeliveryOrder: contract.DeliveryOrder_ORDERED,
```
matzew, line 451: perhaps.... not sure it really makes sense, but:

https://github.com/knative-sandbox/eventing-kafka-broker/blob/d3815d268fb6dd3d2a39c3043ea8606232a2165a/control-plane/pkg/reconciler/source/source.go#L50-L52


is that a change from the other `KafkaChannel`s?

pierDipi: I'm not following what do you mean here.

aliok: I also am not sure about the behaviour of delivery order in KafkaChannel implementations.

matzew: is the `eventing-kafka` unordered ?

pierDipi: eventing-kafka is ordered (_only_)

matzew (approved):
/lgtm
/approve

PR #1508, opened 2021-11-18
Sink URI resolver tests update

tzununbekov:
Sink resolver updated behavior introduced in knative/pkg#2322 requires `v1.Service` Informer injection to retrieve destination port details.

`v1.Service` informers manually injected and prepopulated with objects if they are required in the tests.


matzew:
@tzununbekov do we still need this, or what's the status here ?

tzununbekov:
> @tzununbekov do we still need this, or what's the status here ?

@matzew without the https://github.com/knative/pkg/pull/2322, this PR can be closed. I don't think that I'll have time to take care of it any time soon.

PR #1551, opened 2021-11-29
Make configmap reading reusable

aliok:
Fixes #

## Proposed Changes

- Make configmap reading reusable
- Validate the topicConfig that's built, later in the broker itself. This way, each component can have its own validation.
-

**Release Note**

```release-note

```

**Docs**


control-plane/pkg/reconciler/kafka/topic.go
```diff
@@ -31,6 +39,38 @@ type TopicConfig struct {
 	BootstrapServers []string
 }
 
+func TopicConfigFromConfigMap(logger *zap.Logger, cm *corev1.ConfigMap) (*TopicConfig, error) {
```
pierDipi, line 42: can we pass a `Validator` here and reconcilers can just pass their own validators?

pierDipi (commented):
Do you see `TopicConfig` being validated differently in different reconcilers?

control-plane/pkg/reconciler/kafka/topic.go
aliok: I thought about that. But, to me, returning the topicConfig and then explicitly do the validation in the reconciler seemed cleaner.

aliok:
> Do you see TopicConfig being validated differently in different reconcilers?

Yes, in KafkaChannel it will be different.

pierDipi:
> > Do you see TopicConfig being validated differently in different reconcilers?
> 
> Yes, in KafkaChannel it will be different.

Can you list a difference?

aliok:
> Can you list a difference?

Ok, I think you're right. The default hardcoded config in the broker code got me confused.

Let me move the validation back in to the `TopicConfigFromConfigMap`

control-plane/pkg/reconciler/kafka/topic.go
```diff
@@ -31,6 +39,61 @@ type TopicConfig struct {
...
+	err := configmap.Parse(cm.Data,
+		configmap.AsInt32(DefaultTopicNumPartitionConfigMapKey, &topicDetail.NumPartitions),
+		configmap.AsInt32(DefaultTopicReplicationFactorConfigMapKey, &replicationFactor),
+		configmap.AsString(BootstrapServersConfigMapKey, &bootstrapServers),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse config map %s/%s: %w", cm.Namespace, cm.Name, err)
+	}
+
+	topicDetail.ReplicationFactor = int16(replicationFactor)
```
pierDipi, line 76: FYI, I sent this PR https://github.com/knative/pkg/pull/2365 to pkg so that we can drop this hack on `ReplicationFactor`.

pierDipi (commented):
/lgtm
/approve

PR #1555, opened 2021-11-30
Remove broker.spec.config defaulting magic

pierDipi:
## Proposed Changes

- Remove broker.spec.config defaulting magic
- Add webhook validation (i think we're better off rejecting invalid Brokers from the start)

**Release Note**

```release-note
`broker.spec.config` is now required.
```

**Docs**

- I'll check if docs mentions this feature (I don't thing it does)

/assign @aliok

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker_test.go
+++ b/control-plane/pkg/reconciler/broker/broker_test.go
@@ -171,9 +171,6 @@ func brokerReconciliation(t *testing.T, format string, env config.Env) {
 					),
 				},
 			},
-			OtherTestData: map[string]interface{}{
-				kafka.BootstrapServersConfigMapKey: bootstrapServers,
-			},
 		},
 		{
 			Name: "Reconciled failed - probe " + prober.StatusNotReady.String(),
@@ -232,8 +229,7 @@ func brokerReconciliation(t *testing.T, format string, env config.Env) {
 				},
 			},
 			OtherTestData: map[string]interface{}{
-				kafka.BootstrapServersConfigMapKey: bootstrapServers,
-				testProber:                         probertesting.MockProber(prober.StatusNotReady),
+				testProber: probertesting.MockProber(prober.StatusNotReady),
 			},
 		},
 		{
@@ -293,8 +289,7 @@ func brokerReconciliation(t *testing.T, format string, env config.Env) {
 				},
 			},
 			OtherTestData: map[string]interface{}{
-				kafka.BootstrapServersConfigMapKey: bootstrapServers,
-				testProber:                         probertesting.MockProber(prober.StatusUnknown),
+				testProber: probertesting.MockProber(prober.StatusUnknown),
 			},
 		},
 		{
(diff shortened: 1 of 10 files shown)


matzew:
> `broker.spec.config` is now required.

Isn't that quiet some implication? Or are we generally sure our users do specifiy that ? And don't default on any "defaulting" ?

pierDipi:
> > `broker.spec.config` is now required.
> 
> Isn't that quiet some implication? Or are we generally sure our users do specifiy that ? And don't default on any "defaulting" ?

I'm removing the magic defaulting meaning that if users don't specify any `config` and there is no defaulting provided by the eventing (core) webhook (cluster or namespace defaults), then it will fail.

aliok:
> Isn't that quiet some implication? Or are we generally sure our users do specifiy that ? And don't default on any "defaulting" ?

There was a mechanism that auto-creates the broker, using the defaults in the configmap. Is this still the case? Do we care?

control-plane/pkg/reconciler/broker/broker_test.go
```diff
@@ -121,6 +118,7 @@ func brokerReconciliation(t *testing.T, format string, env config.Env) {
 			Name: "Reconciled normal - no DLS",
 			Objects: []runtime.Object{
 				NewBroker(),
+				BrokerConfig(bootstrapServers, 20, 5),
```
aliok, line 121: I like this better than my approach

pierDipi:
This feature will still provide the defaulting mechanism https://knative.dev/docs/eventing/broker/kafka-broker/#set-as-default-broker-implementation

control-plane/pkg/reconciler/broker/broker_test.go
aliok: Can you remove these though?

```
OtherTestData: map[string]interface{}{
				kafka.BootstrapServersConfigMapKey: bootstrapServers,
			},
```

control-plane/pkg/reconciler/broker/broker_test.go
```diff
@@ -1970,6 +1995,7 @@ func brokerFinalization(t *testing.T, format string, env config.Env) {
 			Name: "Reconciled normal - no broker found in config map",
 			Objects: []runtime.Object{
 				NewDeletedBroker(),
+				BrokerConfig(bootstrapServers, 20, 5),
```
aliok, line 1998: Are there any tests that make sure the config is actually read from the configmap?

control-plane/pkg/reconciler/broker/broker_test.go
pierDipi: Yes, we can remove that.

control-plane/pkg/reconciler/broker/broker_test.go
pierDipi: Yes, all of them, and there is a test that fails because we can't find the ConfigMap.

aliok:
lgtm

Leaving final approval to @matzew

PR #1561, opened 2021-12-01
Kafka Ch reconciler tests with auth

aliok:
Fixes #

## Proposed Changes

- Kafka Ch reconciler tests with auth
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/channel/channel.go
+++ b/control-plane/pkg/reconciler/channel/channel.go
@@ -126,7 +126,7 @@ func (r *Reconciler) reconcileKind(ctx context.Context, channel *messagingv1beta
 	statusConditionManager.ConfigResolved()
 
 	// get the secret to access Kafka
-	secret, err := r.secret(ctx, channelConfigMap)
+	secret, err := security.Secret(ctx, &security.MTConfigMapSecretLocator{ConfigMap: channelConfigMap}, r.SecretProviderFunc())
 	if err != nil {
 		return fmt.Errorf("failed to get secret: %w", err)
 	}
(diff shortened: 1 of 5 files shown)


control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -126,7 +126,7 @@ func (r *Reconciler) reconcileKind(ctx context.Context, channel *messagingv1beta
 	statusConditionManager.ConfigResolved()
 
 	// get the secret to access Kafka
-	secret, err := r.secret(ctx, channelConfigMap)
+	secret, err := security.Secret(ctx, &security.MTConfigMapSecretLocator{ConfigMap: channelConfigMap}, r.SecretProviderFunc())
```
pierDipi, line 129: I think for the channel we want a different `SecretLocator` since the `MTConfigMapSecretLocator` uses, as namespace, the namespace of the ConfigMap itself, while for the channel to be compatible with the existing one we need to read the namespace from the field of ConfigMap (I guess is called `authSecretNamespace`).

aliok: KafkaChannel will also use the new configmap format.

pierDipi: This means that the secret has to be in the `knative-eventing` namespace, is that ok (just wondering)?

aliok: I guess for now it is good. We can improve this when we're working on namespace-scoped channel config (there are tickets already).

pierDipi (approved):
/lgtm

matzew:
This is auth like we do for the brokeR?

aliok:
> This is auth like we do for the broker?

Yes. No e2e tests yet, only reconciler unit tests.

aliok:

```
I1201 14:49:21.407977   63478 request.go:668] Waited for 1.013722884s due to client-side throttling, not priority and fairness, request: GET:https://34.74.49.124/api/v1/namespaces/test-fklnkgui/events?limit=500&resourceVersion=0
```

PR #1574, opened 2021-12-07
Handle non-retryable HTTP status codes as reported in the spec

pierDipi:
As reported in the spec:
https://github.com/knative/specs/blob/c348f501de9eb998b4fd010c54d9127033ee41be/specs/eventing/data-plane.md#event-acknowledgement-and-delivery-retry

Similar to `SelectiveRetry` function in eventing.

## Proposed Changes

- Handle non-retryable HTTP status codes

**Release Note**

```release-note
Handle non-retryable HTTP status codes as reported in the spec: 
For more information: https://github.com/knative/specs/blob/c348f501de9eb998b4fd010c54d9127033ee41be/specs/eventing/data-plane.md#event-acknowledgement-and-delivery-retry
```

**Docs**

```
None
```

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
@@ -26,12 +26,11 @@ import io.vertx.core.Promise;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.ext.web.client.HttpResponse;
 import io.vertx.ext.web.client.WebClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.net.URI;
 import java.util.Objects;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
```diff
@@ -129,4 +134,12 @@ private void logError(final CloudEvent event, Throwable ex) {
     this.client.close();
     return Future.succeededFuture();
   }
+
+  static boolean isRetryableStatusCode(final int statusCode) {
+    // From https://github.com/knative/specs/blob/c348f501de9eb998b4fd010c54d9127033ee41be/specs/eventing/data-plane.md#event-acknowledgement-and-delivery-retry
+    return statusCode >= 500 || // Generic error
+      statusCode == 404 || // Endpoint does not exist
+      statusCode == 409 || // Conflict / Processing in progress
+      statusCode == 429;   // Too Many Requests / Overloaded
```
matzew, line 143: for completeness we have `-1`  there as well, in upstream ...  

I guess that's due to a Golang thingy/bug... but just wanted to mention it here

matzew (approved):
/lgtm
/approve

PR #1591, opened 2021-12-13
[main] Use Broker's own configmap for Kafka

mgencur:
Fixes an issue with using wrong default values when `config-br-defaults` configmap points to another cluster-default configmap that might not include configuration for Kafka. It might include for example MTChannelBasedBroker configuration in which case the Broker creation fails with an error like this
```
invalid configuration - numPartitions: 0 - replicationFactor: 0 - bootstrapServers:  - ConfigMap data: map[channelTemplateSpec:apiVersion: messaging.knative.dev/v1
        kind: InMemoryChannel
        ]
```

## Proposed Changes

- Creation a new ConfigMap in the test namespace that will be used by the Brokers.
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/pkg/broker/broker.go
+++ b/test/pkg/broker/broker.go
@@ -28,7 +28,6 @@ import (
 	"knative.dev/eventing/test/lib/resources"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 
-	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/broker"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/kafka"
 	testingpkg "knative.dev/eventing-kafka-broker/test/pkg/testing"
 )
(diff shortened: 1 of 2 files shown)


test/pkg/broker/broker.go
```diff
@@ -33,9 +40,32 @@ func Creator(client *eventingtestlib.Client, version string) string {
...
+		if err != nil && !apierrors.IsAlreadyExists(err) {
+			client.T.Fatalf("Failed to create ConfigMap %s/%s: %v", namespace, cm.GetName(), err)
+		}
 		client.CreateBrokerOrFail(
 			name,
 			resources.WithBrokerClassForBroker(kafka.BrokerClass),
+			resources.WithConfigForBroker(&duckv1.KReference{
+				Kind:       "ConfigMap",
+				Namespace:  cm.GetNamespace(),
+				Name:       cm.GetName(),
```
pierDipi, line 66: If `apierrors.IsAlreadyExists(err) == true`, cm at line 65 and 66 is `nil`, we should just use `namespace` and a new name variable set to `kafka-broker-upgrade-config`.

pierDipi (approved):

Issue #1537, opened 2021-11-24
Kafka Consumers Scheduling and Scaling


pierDipi:
**Problem**

Currently, each Dispatcher replica instantiates a Consumer for each Trigger with the same consumer group. With an increasing number of Triggers, the Dispatcher resources need to be increased as well and there is no way to configure how many consumers for a specific Trigger we need to run. 
This proposal will allow partitioning consumers across Dispatcher replicas and fine-grained control over the parallelism of each Trigger.

Design doc: https://docs.google.com/document/d/1UktwiDyqq07MtA7pUlahEpux5CCdAsyI6k3nkQeeqXw/edit

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?
- Event Consumer

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

30

**Additional context (optional)**

/priority important-soon
/area broker
/area channel
/area source
/kind feature-request

devguyio:
I know I'm coming late to the party. So please excuse me. I am worried about the consequences of choosing the "ConsumerGroup" name. 
1. It has a direct connotation with Kafka's consumer groups and its symantics.
2. If some Kafka operator (say Strimzi) decides to create a ConsumerGroup CRD for whatever reason, it'll collide with ours and will be confusing for the users.

I'm thinking of a more generic term. Maybe "DispatcherPartition" and "DispatcherPartitionGroup" or something along those lines.

aavarghese:
@pierDipi I was able to deploy KafkaSource with consumer replicas (under and over [pod capacity](https://github.com/aavarghese/eventing-kafka-broker/blob/685ce5045ab537285408da6639b28460f3e0c98c/control-plane/pkg/reconciler/consumergroup/controller.go#L84)) and the HA scheduler with this set of predicates and priorities in the [scheduler CM](https://github.com/aavarghese/eventing-kafka-broker/blob/685ce5045ab537285408da6639b28460f3e0c98c/control-plane/config/200-controller/100-config-kafka-scheduler.yaml).

With 25 VReplicas for example,
```
  Placements:
    Pod Name:      kafka-source-dispatcher-0
    Vreplicas:     13
    Pod Name:      kafka-source-dispatcher-1
    Vreplicas:     12
  Replicas:        25
  ```
  
  But I was seeing an issue when running with priority plugin `{"Name": "AvailabilityZonePriority",}`
  
[...]

PR #1611, opened 2021-12-14
KafkaSource Reconciler for ConsumerGroup scaling

aavarghese:

Part of #1537 (sourcev2_test.go coming soon)

## Proposed Changes

- Adding new KafkaSource reconciler to create ConsumerGroup resources

**Release Note**

```release-note

```

**Docs**


control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -0,0 +1,137 @@
...
+	consgroup, err := r.reconcileConsumerGroup(ctx, ks)
+	if err != nil {
+		ks.GetConditionSet().Manage(&ks.Status).MarkFalse(consumergroup.KafkaConditionConsumerGroupNotReady, "failed to reconcile consumer group", err.Error())
+		return err
+	}
+	ks.GetConditionSet().Manage(&ks.Status).MarkTrue(consumergroup.KafkaConditionConsumerGroupReady)
+
+	ks.Status.MarkSink(consgroup.Status.SubscriberURI)
+	ks.Status.Placeable = consgroup.Status.Placeable
+	ks.Status.Consumers = *consgroup.Spec.Replicas
```
pierDipi, line 64: Maybe we need a `consgroup.Status.Replicas` field?

control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -0,0 +1,137 @@
...
+	if apierrors.IsNotFound(err) {
+		cg, err := r.InternalsClient.InternalV1alpha1().ConsumerGroups(newcg.GetNamespace()).Create(ctx, newcg, metav1.CreateOptions{})
+		if err != nil && !apierrors.IsAlreadyExists(err) {
+			return nil, fmt.Errorf("failed to create consumer group %s/%s: %w", newcg.GetNamespace(), newcg.GetName(), err)
+		}
+		if apierrors.IsAlreadyExists(err) {
+			return newcg, nil
+		}
+		return cg, nil
+	}
```
pierDipi, line 120: If the `ks.Spec.ConsumerGroup` is not immutable on the KafkaSource CR this might leak ConsumerGroup resources since 
- a user creates a KafkaSource with `Spec.ConsumerGroup: c1`
- we create a `ConsumerGroup` called `c1`
- a user updates the KafkaSource with `Spec.ConsumerGroup: c2`
- c1 will keep being in the system even though it shouldn't be.

What about using the UUID as name?

control-plane/pkg/reconciler/source/v2/sourcev2.go
aavarghese: Yeah. I guess so - to reflect the actual replicas...

control-plane/pkg/reconciler/source/v2/sourcev2.go
aavarghese: Ah yes, right. `c1` will still have `ks` as owner which means it wont be deleted until `ks` is deleted too

control-plane/pkg/reconciler/source/v2/controllerv2.go
```diff
@@ -0,0 +1,58 @@
...
+		ConsumerGroupLister: consumerGroupInformer.Lister(),
+		InternalsClient:     consumergroupclient.Get(ctx),
+	}
+
+	impl := kafkasource.NewImpl(ctx, r)
+
+	kafkaInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: kafka.BrokerClassFilter(),
+		Handler:    controller.HandleAll(impl.Enqueue),
+	})
```
pierDipi, line 50: We shouldn't filter KafkaSources by broker class.

control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+type ReconcilerV2 struct {
+	ConsumerGroupLister internalslst.ConsumerGroupLister
+	InternalsClient     internalsclient.Interface
+}
+
+func (r *ReconcilerV2) ReconcileKind(ctx context.Context, ks *sources.KafkaSource) reconciler.Event {
+
+	//todo initialize consumer group offsets here?
```
pierDipi, line 51: We should do this in the ConsumerGroup reconciler.

control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -0,0 +1,135 @@
...
+			OwnerReferences: []metav1.OwnerReference{
+				*kmeta.NewControllerRef(ks),
+			},
+		},
+		Spec: internalscg.ConsumerGroupSpec{
+			Template: internalscg.ConsumerTemplateSpec{
+				Spec: internalscg.ConsumerSpec{
+					Topics: ks.Spec.Topics,
+					Configs: internalscg.ConsumerConfigs{Configs: map[string]string{
+						"group.id":          string(ks.UID),
```
pierDipi, line 82: This should be `ks.Spec.ConsumerGroup`.

control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -0,0 +1,135 @@
...
+		TypeMeta:   cg.TypeMeta,
+		ObjectMeta: cg.ObjectMeta,
+		Spec:       newcg.Spec,
+		Status:     cg.Status,
+	}
+	if _, err := r.InternalsClient.InternalV1alpha1().ConsumerGroups(cg.GetNamespace()).Update(ctx, newCg, metav1.UpdateOptions{}); err != nil {
+		return nil, fmt.Errorf("failed to update consumer group %s/%s: %w", newCg.GetNamespace(), newCg.GetName(), err)
+	}
+
+	return cg, nil
```
pierDipi, line 134: Should we return the `cg` returned by the `Update` call?

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -138,13 +138,15 @@ func (r ReconcilerV2) reconcileConsumerGroup(ctx context.Context, trigger *event
...
-	//todo validate newcg
-
 	cg, err := r.ConsumerGroupLister.ConsumerGroups(trigger.GetNamespace()).Get(string(trigger.UID)) //Get by consumer group name
 	if err != nil && !apierrors.IsNotFound(err) {
 		return nil, err
 	}
 
+	if err := newcg.Validate(ctx); err != nil {
+		return nil, fmt.Errorf("failed to validate expected consumer group %s/%s: %w", newcg.GetNamespace(), newcg.GetName(), err)
+	}
```
pierDipi, line 148: I think there shouldn't validate the CG here since if it's invalid we get an error once we create it.

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -85,10 +86,10 @@ func (r *ReconcilerV2) ReconcileKind(ctx context.Context, trigger *eventing.Trig
 
 	consgroup, err := r.reconcileConsumerGroup(ctx, trigger)
 	if err != nil {
-		trigger.Status.MarkDependencyFailed("failed to reconcile consumer group", err.Error())
+		trigger.GetConditionSet().Manage(&trigger.Status).MarkFalse(consumergroup.KafkaConditionConsumerGroupNotReady, "failed to reconcile consumer group", err.Error())
 		return err
 	}
-	trigger.Status.MarkDependencySucceeded()
+	trigger.GetConditionSet().Manage(&trigger.Status).MarkTrue(consumergroup.KafkaConditionConsumerGroupReady)
```
pierDipi, line 92: The problem with this approach is that the condition `KafkaConditionConsumerGroupNotReady` won't be considered for computing the top-level `Ready` condition by the reconciler framework, can we keep using the `Dependency` condition?

control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -0,0 +1,135 @@
...
+func (r *ReconcilerV2) ReconcileKind(ctx context.Context, ks *sources.KafkaSource) reconciler.Event {
+
+	//todo initialize consumer group offsets here?
+
+	consgroup, err := r.reconcileConsumerGroup(ctx, ks)
+	if err != nil {
+		ks.GetConditionSet().Manage(&ks.Status).MarkFalse(consumergroup.KafkaConditionConsumerGroupNotReady, "failed to reconcile consumer group", err.Error())
+		return err
+	}
+	ks.GetConditionSet().Manage(&ks.Status).MarkTrue(consumergroup.KafkaConditionConsumerGroupReady)
```
pierDipi, line 58: We need to make sure that the condition `KafkaConditionConsumerGroupNotReady` is registered and the call `GetConditionSet()` method on the KafkaSource object returns all conditions that are relevant to this reconciler.

See `RegisterAlternateConditionSet` for the Broker.

control-plane/pkg/reconciler/consumergroup/consumergroup.go
```diff
@@ -33,6 +34,14 @@ import (
 	kafkainternalslisters "knative.dev/eventing-kafka-broker/control-plane/pkg/client/internals/kafka/listers/eventing/v1alpha1"
 )
 
+const (
+	// KafkaConditionConsumerGroupReady has status True when the KafkaChannel has created a consumer group.
+	KafkaConditionConsumerGroupReady apis.ConditionType = "ConsumerGroupReady"
+
+	// KafkaConditionConsumerGroupNotReady has status False when the KafkaChannel has failed to create a consumer group.
+	KafkaConditionConsumerGroupNotReady apis.ConditionType = "ConsumerGroupNotReady"
+)
```
pierDipi, line 43: I think we should have only one condition `ConsumerGroup` (or something like that) and then we can mark that condition with true or false based on the state or errors we get.

Conditions like this one should be marked as true to mark the entire KafkaSource as ready, so `ConsumerGroupNotReady` won't be true in this case.

control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
```diff
@@ -93,6 +93,13 @@ type ConsumerGroupStatus struct {
 	// DeliveryStatus contains a resolved URL to the dead letter sink address, and any other
 	// resolved delivery options.
 	eventingduckv1.DeliveryStatus `json:",inline"`
+
+	// Replicas is the desired number of replicas of the given Template.
+	// These are replicas in the sense that they are instantiations of the
+	// same Template, but individual replicas also have a consistent identity.
+	// If unspecified, defaults to 1.
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
```
pierDipi, line 102: The `replicas` field in the status shouldn't be the "desired` number of replicas.

maybe the latest observed number of replicas or we can take a look at the replicas field in the status of the `Deployment` resource.

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
aavarghese: The idea here was to validate the `newcg` expected consumer group respresenation that we build before calling `.Create(ctx....)`. Such that `newcg` must have all the fields expected by Validate to be set. Shouldn't that be necessary?

aavarghese: Right now, we aren't setting `Replicas`, `Selector`, etc in `newcg` so Validate will fail..

control-plane/pkg/reconciler/source/v2/sourcev2.go
aavarghese: Sure. Where can I find `RegisterAlternateConditionSet`?

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
pierDipi: > Shouldn't that be necessary?

Yes, since the webhook will reject invalid consumer groups when we call `Create(ctx, ...)`.

The other problem with this is that a resource is validated after calling `SetDefaults(ctx)` on it which is done by the webhook as well, so it's ok to call `Create(ctx, ...)` with an invalid resource since after the mutating webhook runs the resource might become valid.

control-plane/pkg/reconciler/source/v2/sourcev2.go
pierDipi: Sorry, it's called `RegisterAlternateBrokerConditionSet`

https://github.com/knative/eventing/blob/d98ae8c06a46d930d08d8158a3f2962a99d9efca/pkg/apis/eventing/v1/broker_lifecycle.go#L44

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
aavarghese: I see, now I understand how the SetDefaults and Validate apis are used. I'm following. Thanks @pierDipi.

control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -0,0 +1,136 @@
...
+					Delivery: &internalscg.DeliverySpec{
+						Ordering: DefaultDeliveryOrder,
+					},
+					Subscriber: ks.Spec.Sink,
+					CloudEventOverrides: &duckv1.CloudEventOverrides{
+						Extensions: ks.Spec.CloudEventOverrides.Extensions,
+					},
+				},
+			},
+		},
```
pierDipi, line 102: Shouldn't we set `spec.replicas = ks.spec.replicas` too?

aavarghese: Yes! That was missed - added now.

control-plane/pkg/reconciler/source/v2/controllerv2.go
```diff
@@ -0,0 +1,60 @@
...
+	impl := kafkasource.NewImpl(ctx, r)
+
+	kafkaInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		Handler: controller.HandleAll(impl.Enqueue),
+	})
+
+	// ConsumerGroup changes and enqueue associated KafkaSource
+	consumerGroupInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: consumergroup.Filter("kafkasource"),
+		Handler:    controller.HandleAll(impl.Enqueue),
```
pierDipi, line 57: Here we need to enqueue the KafkaSource key associated with the ConsumerGroup and not the ConsumerGroup resource.

pierDipi: I think we're doing this wrong in the trigger controller too.

pierDipi: Example https://github.com/knative-sandbox/eventing-kafka-broker/pull/1708

pierDipi (approved):
/lgtm

PR #1612, opened 2021-12-15
Verify Trigger finalizer isn't set on Triggers of another class

pierDipi:
Enabling Kafka Broker while running channel-based Broker with
Kafka sets incorrect finalizer.

## Proposed Changes

- Verify Trigger finalizer isn't set on Triggers of another class

**Release Note**

```release-note
None
```

**Docs**

```
None
```

PR #1616, opened 2021-12-15
Add promote filter function to Broker and Trigger reconciler

pierDipi:
When a reconciler gets promoted to be the leader, every object
gets re-queued even if it shouldn't be.

Tested in #1612

## Proposed Changes

- Add promote filter function to Broker and Trigger reconciler

**Release Note**

```release-note
None
```

**Docs**

```
None
```


matzew (approved):
/lgtm
/approve

PR #1618, opened 2021-12-15
KafkaChannel Reconciler for ConsumerGroup scaling

aavarghese:

Part of #1537 (channelv2_test.go coming soon)

## Proposed Changes

- Adding new KafkaChannel reconciler to create ConsumerGroup resources

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/channel/v2/channelv2.go
+++ b/control-plane/pkg/reconciler/channel/v2/channelv2.go
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2021 The Knative Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package v2
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"knative.dev/pkg/kmeta"
+	"knative.dev/pkg/reconciler"
+
+	messagingv1beta1 "knative.dev/eventing-kafka/pkg/apis/messaging/v1beta1"
+
+	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/consumergroup"
+	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/kafka"
+
+	internals "knative.dev/eventing-kafka-broker/control-plane/pkg/apis/internals/kafka/eventing"
+	internalscg "knative.dev/eventing-kafka-broker/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1"
+	internalsclient "knative.dev/eventing-kafka-broker/control-plane/pkg/client/internals/kafka/clientset/versioned"
+	internalslst "knative.dev/eventing-kafka-broker/control-plane/pkg/client/internals/kafka/listers/eventing/v1alpha1"
+)
+
+const (
+	// TopicPrefix is the Kafka Channel topic prefix - (topic name: knative-messaging-kafka.<channel-namespace>.<channel-name>).
+	TopicPrefix          = "knative-messaging-kafka"
+	DefaultDeliveryOrder = internals.Ordered
+)
+
+type Reconciler struct {
+	ConsumerGroupLister internalslst.ConsumerGroupLister
+	InternalsClient     internalsclient.Interface
+}
+
+func (r *Reconciler) ReconcileKind(ctx context.Context, channel *messagingv1beta1.KafkaChannel) reconciler.Event {
+
+	consgroup, err := r.reconcileConsumerGroup(ctx, channel)
+	if err != nil {
+		channel.GetConditionSet().Manage(&channel.Status).MarkFalse(consumergroup.KafkaConditionConsumerGroupNotReady, "failed to reconcile consumer group", err.Error())
+		return err
+	}
+	channel.GetConditionSet().Manage(&channel.Status).MarkTrue(consumergroup.KafkaConditionConsumerGroupReady)
+
+	//todo other status fields relevant for channels
+
+	//SubscribableStatus
+	//AddressStatus
+
+	channel.Status.DeadLetterSinkURI = consgroup.Status.DeadLetterSinkURI //DeliveryStatus
+	return nil
+}
+
+func (r Reconciler) reconcileConsumerGroup(ctx context.Context, channel *messagingv1beta1.KafkaChannel) (*internalscg.ConsumerGroup, error) {
+
+	topicName := kafka.ChannelTopic(TopicPrefix, channel)
+	newcg := &internalscg.ConsumerGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      string(channel.UID),
+			Namespace: channel.Namespace,
+			OwnerReferences: []metav1.OwnerReference{
+				*kmeta.NewControllerRef(channel),
+			},
+		},
+		Spec: internalscg.ConsumerGroupSpec{
+			Template: internalscg.ConsumerTemplateSpec{
+				Spec: internalscg.ConsumerSpec{
+					Topics: []string{topicName},
+					Configs: internalscg.ConsumerConfigs{Configs: map[string]string{
+						"group.id": string(channel.UID),
+					}},
+					Delivery: &internalscg.DeliverySpec{
+						DeliverySpec: channel.Spec.Delivery,
+						Ordering:     DefaultDeliveryOrder,
+					},
+				},
+			},
+		},
+	}
+
+	cg, err := r.ConsumerGroupLister.ConsumerGroups(channel.GetNamespace()).Get(string(channel.UID)) //Get by consumer group id
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+
+	if err := newcg.Validate(ctx); err != nil {
+		return nil, fmt.Errorf("failed to validate expected consumer group %s/%s: %w", newcg.GetNamespace(), newcg.GetName(), err)
+	}
+
+	if apierrors.IsNotFound(err) {
+		cg, err := r.InternalsClient.InternalV1alpha1().ConsumerGroups(newcg.GetNamespace()).Create(ctx, newcg, metav1.CreateOptions{})
+		if err != nil && !apierrors.IsAlreadyExists(err) {
+			return nil, fmt.Errorf("failed to create consumer group %s/%s: %w", newcg.GetNamespace(), newcg.GetName(), err)
+		}
+		if apierrors.IsAlreadyExists(err) {
+			return newcg, nil
+		}
+		return cg, nil
+	}
+
+	if equality.Semantic.DeepDerivative(newcg.Spec, cg.Spec) {
+		return cg, nil
+	}
+
+	newCg := &internalscg.ConsumerGroup{
+		TypeMeta:   cg.TypeMeta,
+		ObjectMeta: cg.ObjectMeta,
+		Spec:       newcg.Spec,
+		Status:     cg.Status,
+	}
+	if _, err := r.InternalsClient.InternalV1alpha1().ConsumerGroups(cg.GetNamespace()).Update(ctx, newCg, metav1.UpdateOptions{}); err != nil {
+		return nil, fmt.Errorf("failed to update consumer group %s/%s: %w", newCg.GetNamespace(), newCg.GetName(), err)
+	}
+
+	return cg, nil
+}
--- a/control-plane/pkg/reconciler/channel/v2/channelv2.go
+++ b/control-plane/pkg/reconciler/channel/v2/channelv2.go
@@ -18,6 +18,7 @@ package v2
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,132 @@
...
+	//AddressStatus
+
+	channel.Status.DeadLetterSinkURI = consgroup.Status.DeadLetterSinkURI //DeliveryStatus
+	return nil
+}
+
+func (r Reconciler) reconcileConsumerGroup(ctx context.Context, channel *messagingv1beta1.KafkaChannel) (*internalscg.ConsumerGroup, error) {
+
+	topicName := kafka.ChannelTopic(TopicPrefix, channel)
+	newcg := &internalscg.ConsumerGroup{
```
pierDipi, line 73: We need one consumer group per subscriber here, one consumer group per channel isn't enough.

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,132 @@
...
+	TopicPrefix          = "knative-messaging-kafka"
+	DefaultDeliveryOrder = internals.Ordered
+)
+
+type Reconciler struct {
+	ConsumerGroupLister internalslst.ConsumerGroupLister
+	InternalsClient     internalsclient.Interface
+}
+
+func (r *Reconciler) ReconcileKind(ctx context.Context, channel *messagingv1beta1.KafkaChannel) reconciler.Event {
```
pierDipi, line 52: In this reconciler we need to handle the ingress part, are you planning on doing it in a different PR?

pierDipi: The ingress part should be set like here https://github.com/knative-sandbox/eventing-kafka-broker/blob/e32734a003082847ed1d98a27f07f6a362e26248/control-plane/pkg/reconciler/channel/channel.go#L529-L555

Then for each subscriber like here: https://github.com/knative-sandbox/eventing-kafka-broker/blob/e32734a003082847ed1d98a27f07f6a362e26248/control-plane/pkg/reconciler/channel/channel.go#L406-L429 instead of this logic: https://github.com/knative-sandbox/eventing-kafka-broker/blob/e32734a003082847ed1d98a27f07f6a362e26248/control-plane/pkg/reconciler/channel/channel.go#L433-L451 we just create a new `ConsumerGroup` resource accordingly.

aavarghese: Done

aavarghese:
@pierDipi if you have time for another review please? :)

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,260 @@
...
+
+type Reconciler struct {
+	*base.Reconciler
+	*config.Env
+	ConfigMapLister     corelisters.ConfigMapLister
+	ConsumerGroupLister internalslst.ConsumerGroupLister
+	InternalsClient     internalsclient.Interface
+}
+
+func (r *Reconciler) ReconcileKind(ctx context.Context, channel *messagingv1beta1.KafkaChannel) reconciler.Event {
```
pierDipi, line 78: I think we're missing some bits:

- [Probing and setting the channel address](https://github.com/knative-sandbox/eventing-kafka-broker/blob/3f3f37ca49a64f6d3181b4296ce8c2b7b39e7028/control-plane/pkg/reconciler/channel/channel.go#L268-L283)
- [CM tracking](https://github.com/knative-sandbox/eventing-kafka-broker/blob/3f3f37ca49a64f6d3181b4296ce8c2b7b39e7028/control-plane/pkg/reconciler/channel/channel.go#L116)
- [Topic creation (with and without auth)](https://github.com/knative-sandbox/eventing-kafka-broker/blob/3f3f37ca49a64f6d3181b4296ce8c2b7b39e7028/control-plane/pkg/reconciler/channel/channel.go#L121-L182)
- [Annotation update](https://github.com/knative-sandbox/eventing-kafka-broker/blob/3f3f37ca49a64f6d3181b4296ce8c2b7b39e7028/control-plane/pkg/reconciler/channel/channel.go#L229-L246)
- [Finalizer](https://github.com/knative-sandbox/eventing-kafka-broker/blob/3f3f37ca49a64f6d3181b4296ce8c2b7b39e7028/control-plane/pkg/reconciler/channel/channel.go#L292)

Overall we need pretty much the same logic of the other reconciler, the logic that changes is [reconcileSubscribers](https://github.com/knative-sandbox/eventing-kafka-broker/blob/3f3f37ca49a64f6d3181b4296ce8c2b7b39e7028/control-plane/pkg/reconciler/channel/channel.go#L205-L207)

IMO, if we can refactor the old channel and in a way that we can reuse the logic and plug our own `reconcileSubscribers` that would be better.

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,579 @@
...
+			})
+			globalErr = multierr.Append(globalErr, err)
+		} else {
+			logger.Debug("marking subscription as ready")
+			channel.Status.Subscribers = append(channel.Status.Subscribers, v1.SubscriberStatus{
+				UID:                s.UID,
+				ObservedGeneration: s.Generation,
+				Ready:              corev1.ConditionTrue,
+			})
+		}
```
pierDipi, line 425: we should propagate the readiness of the ConsumerGroup here, so that if a ConsumerGroup for a subscriber is ready we set the subscription as ready.

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,579 @@
...
+	}
```
pierDipi, line 426: How do we handle the case that a subscriber might be deleted from this list?

pierDipi: If we leave the logic as it is, the ConsumerGroup associated with that subscriber will be leaked.

aavarghese: adding this cg cleanup to latest commit - can you TAL please?

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,624 @@
...
+	if err := r.UpdateDispatcherPodsAnnotation(ctx, logger, ct.Generation); err != nil {
+		return err
+	}
+
+	// TODO probe (as in #974) and check if status code is 404 otherwise requeue and return.
+	//  Rationale: after deleting a topic closing a producer ends up blocking and requesting metadata for max.block.ms
+	//  because topic metadata aren't available anymore.
+	// 	See (under discussions KIPs, unlikely to be accepted as they are):
+	// 	- https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=181306446
+	// 	- https://cwiki.apache.org/confluence/display/KAFKA/KIP-286%3A+producer.send%28%29+should+not+block+on+metadata+update
```
pierDipi, line 342: We did this in the other channel, maybe you need to rebase to get the latest code?

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,624 @@
...
+	newcg := &internalscg.ConsumerGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      string(s.UID),
+			Namespace: channel.Namespace,
+			OwnerReferences: []metav1.OwnerReference{
+				*kmeta.NewControllerRef(channel),
+			},
+			Labels: map[string]string{
+				KafkaChannelNameLabel: channel.Name, // identifies the new ConsumerGroup as associated with this channel (same namespace)
+			},
```
pierDipi, line 488: We need to make sure that this label is set and immutable in the ConsumerGroup validation.

control-plane/pkg/reconciler/channel/v2/channelv2.go
aavarghese: Oh yes, I do. Thanks. Fixing.

control-plane/pkg/reconciler/channel/v2/channelv2.go
aavarghese: Currently, we only have this label for consumer groups created by KafkaChannel reconcilers. If adding as part of validation, do you recommend we add this label in KafkaSource and Trigger reconciler code too for CGs? @pierDipi

pierDipi: Eh, that's right, so let's not do this, probably we can only add the immutability check.

When the label is present in the previous version, it must be present in the current version of the ConsumerGroup.

aavarghese: ok - that works - will add

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,624 @@
...
+					//globalErr = multierr.Append(globalErr, errors.New(msg))
+				} else {
+					msg := fmt.Sprint("Subscription not ready. ", topLevelCondition.Reason, topLevelCondition.Message)
+					channel.Status.Subscribers = append(channel.Status.Subscribers, v1.SubscriberStatus{
+						UID:                s.UID,
+						ObservedGeneration: s.Generation,
+						Ready:              corev1.ConditionFalse,
+						Message:            msg,
+					})
+					//globalErr = multierr.Append(globalErr, errors.New(msg))
```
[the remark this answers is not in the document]
aavarghese: Yeah, possibly. I'm figuring out whether we want to return an error if the CG status is just unknown. Will update accordingly.

aavarghese:
Rebased and added CG selector label

aavarghese:
ping @pierDipi :)

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,636 @@
...
+	address := receiver.Address(r.IngressHost, channel)
+	proberAddressable := prober.Addressable{
+		Address: address,
+		ResourceKey: types.NamespacedName{
+			Namespace: channel.GetNamespace(),
+			Name:      channel.GetName(),
+		},
+	}
+	if status := r.Prober.Probe(ctx, proberAddressable, prober.StatusNotReady); status != prober.StatusNotReady {
+		return nil // Object will get re-queued once probe status changes.
```
pierDipi, line 348: Like the other channel reconciler: https://github.com/knative-sandbox/eventing-kafka-broker/blob/683e88c5c96531e9b1293ab6b8e7b2325cb13a66/control-plane/pkg/reconciler/channel/channel.go#L356-L358
```suggestion
		// Return a requeueKeyError that doesn't generate an event and it re-queues the object
		// for a new reconciliation.
		return controller.NewRequeueAfter(5 * time.Second)
```

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,636 @@
...
+		globalErr = multierr.Append(globalErr, err)
+	}
+	for _, cg := range channelCgs {
+		found := false
+		for _, ccg := range currentCgs {
+			if ccg.Name == cg.Name {
+				found = true
+			}
+		}
+		if !found { //ConsumerGroup needs to be deleted since it isn't associated with an existing subscriber (subscriber may have been deleted)
```
pierDipi, line 474:
```suggestion
		if !found { // ConsumerGroup needs to be deleted since it isn't associated with an existing subscriber (subscriber may have been deleted)
```

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,636 @@
...
+
+	// Get all consumer groups associated with this Channel
+	selector := labels.SelectorFromSet(map[string]string{internalscg.KafkaChannelNameLabel: channel.Name})
+	channelCgs, err := r.ConsumerGroupLister.ConsumerGroups(channel.GetNamespace()).List(selector)
+	if err != nil {
+		globalErr = multierr.Append(globalErr, err)
+	}
+	for _, cg := range channelCgs {
+		found := false
+		for _, ccg := range currentCgs {
```
pierDipi, line 469: I'm wondering if we can use a map for `currentCgs` `name -> ConsumerGroup` so that we can avoid the double loop.
`found` would be just this:
```golang
found, _ := currentCgs[cg.Name]
```

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,636 @@
...
+			},
+			Labels: map[string]string{
+				internalscg.KafkaChannelNameLabel: channel.Name, // identifies the new ConsumerGroup as associated with this channel (same namespace)
+			},
+		},
+		Spec: internalscg.ConsumerGroupSpec{
+			Template: internalscg.ConsumerTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"eventing.knative.dev/subscriberUID": string(s.UID),
```
pierDipi, line 501: Nit: maybe we can use this label?
```suggestion
						"kafka.eventing.knative.dev/subscriber.uid": string(s.UID),
```

pierDipi (commented):
Thanks!

Some minor comments

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -0,0 +1,634 @@
...
+
+	return nil
+}
+
+func (r *Reconciler) reconcileSubscribers(ctx context.Context, channel *messagingv1beta1.KafkaChannel, topicName string, bootstrapServers []string) error {
+	logger := kafkalogging.CreateReconcileMethodLogger(ctx, channel)
+
+	channel.Status.Subscribers = make([]v1.SubscriberStatus, 0)
+	var globalErr error
+	currentCgs := make(map[string]*internalscg.ConsumerGroup)
```
pierDipi, line 416: Nit:
```suggestion
	currentCgs := make(map[string]*internalscg.ConsumerGroup, len(channel.Spec.Subscribers))
```

pierDipi (approved):
/lgtm
/approve

We can address the nit in another PR, thanks!

control-plane/pkg/reconciler/channel/v2/channelv2.go
aavarghese: Will modify in unti test PR.

PR #1651, opened 2021-12-22
Add Source Observer ClusterRole

pierDipi:
Eventing controller error:
```
kafkasources.sources.knative.dev is forbidden:
User \"system:serviceaccount:knative-eventing:eventing-controller\"
cannot list resource \"kafkasources\" in API group
\"sources.knative.dev\" at the cluster scope
```

/assign @aliok


aliok (approved):
/lgtm
/approve

Why did we add this now? Is it newly needed or was it missing and we didn't notice?

pierDipi:
> /lgtm /approve
> 
> Why did we add this now? Is it newly needed or was it missing and we didn't notice?

this one: 
> it was missing and we didn't notice

Issue #1683, opened 2022-01-04
Eventing Kafka Sink GA audit


pierDipi:
See [Eventing Kafka GA requirements document](https://docs.google.com/document/d/1H0nGnnT1TV27tSbApJVoDdMHkpobP2vWfHAS3xymNpU/edit#bookmark=id.awnfw2naaptw)

Issue #1682, opened 2022-01-04
Eventing Kafka Broker GA audit


pierDipi:
See [Eventing Kafka GA requirements document](https://docs.google.com/document/d/1H0nGnnT1TV27tSbApJVoDdMHkpobP2vWfHAS3xymNpU/edit#bookmark=id.os4u9qwbk42k)

PR #1686, opened 2022-01-04
Use main as branch in codecov config

pierDipi:
It uses master by default, so it can't show coverage data:
https://app.codecov.io/gh/knative-sandbox/eventing-kafka-broker/blob/master/control-plane/pkg/kafka/consumer_group_lag.go

Part of #1682 #1683


matzew (approved):
/lgtm
/approve

PR #1687, opened 2022-01-04
Add test for PartitionLag.String() method

pierDipi:
As per title, to increase code coverage.

Part of #1682


matzew (approved):
/lgtm
/approve

PR #1695, opened 2022-01-05
Align Receiver metrics to Eventing core Ingress metrics

pierDipi:
See https://knative.dev/docs/eventing/observability/metrics/eventing-metrics/#broker-ingress

_Note: Eventing Core uses `broker_name` for the tag but since we have Broker/Channel/Sink, I'm using `name` instead._

Example metrics:
```
# HELP event_count_total Number of events received
# TYPE event_count_total counter
event_count_total{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",} 15915.0

# HELP event_dispatch_latencies_max The time spent dispatching an event to Kafka
# TYPE event_dispatch_latencies_max gauge
event_dispatch_latencies_max{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",} 99.0
# HELP event_dispatch_latencies The time spent dispatching an event to Kafka
# TYPE event_dispatch_latencies histogram
event_dispatch_latencies_bucket{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",le="1.0",} 7499.0
event_dispatch_latencies_bucket{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",le="2.0",} 14566.0
event_dispatch_latencies_bucket{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",le="5.0",} 15742.0
event_dispatch_latencies_bucket{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",le="10.0",} 15862.0
event_dispatch_latencies_bucket{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",le="20.0",} 15894.0
event_dispatch_latencies_bucket{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",le="50.0",} 15907.0
event_dispatch_latencies_bucket{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",le="100.0",} 15915.0
event_dispatch_latencies_bucket{event_type="com.example.FullEvent",name="",namespace_name="",response_code="202",response_code_class="2xx",le="500.0",} 15915.0
[...]

PR #1701, opened 2022-01-07
Set resource reference in Broker, Channel, and Sink contracts

pierDipi:
To align receiver metrics to Eventing core #1695, we need
to have the resource reference in the data plane.


matzew (approved):
/lgtm
/approve

PR #1708, opened 2022-01-07
Enqueue trigger when ConsumerGroups change

pierDipi:
We need to enqueue the Trigger key associated with the 
ConsumerGroup and not the ConsumerGroup resource.

Related https://github.com/knative-sandbox/eventing-kafka-broker/pull/1611#discussion_r780340587

## Proposed Changes

- Enqueue trigger when ConsumerGroups change


aavarghese (approved):
Great! (Maybe we can name the file something other than `filter.go`, but not impt for now)
/lgtm

PR #1723, opened 2022-01-11
Support replies in KafkaChannel

aliok:
Fixes #

## Proposed Changes

- Support replies in KafkaChannel
- Create new `ResponseHandler` that does it
- Rename `KafkaResponseHandler` to `ResponseToKafkaTopicHandler` because the original name meant "handle the response from Kafka" to me
- Enable e2e tests that use sink replies

**Release Note**

```release-note

```

**Docs**


pierDipi (commented):
Have you considered using composition over inheritance?

the Response Handler (that isn't the noop) extracts an event and forwards it to a ReplyHandler if there is an event.

ReplyHandler would be implemented by `ResponseToHttpEndpointHandler` or `ResponseToKafkaTopicHandler`

aliok:
> Have you considered using composition over inheritance?

I thought about it. I prefer leaving this as is for now.

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
```diff
@@ -202,19 +186,46 @@ public AbstractVerticle get(final DataPlaneContract.Resource resource, final Dat
...
+    if(egress.hasReplyUrl()){
+      return httpSupplier.get();
+    } else if(egress.hasReplyToOriginalTopic()){
+      return kafkaSupplier.get();
+    } else if (egress.hasDiscardReply()) {
       return new NoopResponseHandler();
     }
-    return defaultSupplier.get();
+    //TODO: log
+    return kafkaSupplier.get();
```
aliok, line 200: I had to carry over the logic because so many things failed when I changed the logic to return `Noop` when nothing was set.

I think we should do one of these:
- Check for requiredness of the `replyStrategy` and reject when it is not set
- If not, at least default to `Noop` and not to Kafka

I don't want to do this in this PR though.

If you guys agree, I will create an issue about that.

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
```diff
@@ -171,7 +153,9 @@ public AbstractVerticle get(final DataPlaneContract.Resource resource, final Dat
           new AttributesFilter(egress.getFilter().getAttributesMap()) :
           Filter.noop();
 
-        final var responseHandler = getNoopResponseHandlerOrDefault(egress, () -> getKafkaResponseHandler(vertx, producerConfigs, resource));
+        final var responseHandler = getResponseHandler(egress,
+          () -> getResponseToKafkaTopicHandler(vertx, producerConfigs, resource),
+          () -> getResponseToHttpEndpointHandler(vertx, egress.getReplyUrl()));
```
pierDipi, line 158: Can't we do this?

```suggestion
          () -> new ResponseToHttpEndpointHandler(createConsumerRecordSender(vertx, egress.getReplyUrl(), egressConfig));
```

pierDipi: Then we can remove the `getResponseToHttpEndpointHandler` method or wrap the suggested logic in that method.

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
pierDipi: > Check for requiredness of the replyStrategy and reject when it is not set
> If not, at least default to Noop and not to Kafka

I'd do the second one since we don't have a way of telling the control plane that the reply strategy wasn't set properly.

In general, yes, controllers must set the reply field (we can safety checks to the common logic to update the CM in the controller)

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
aliok: Yes, thanks for catching that up. I forgot about my TODOs I added there to do something similar

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
aliok: I see this as an implementation problem and after giving some thought, I would prefer #1.

Created a ticket here: https://github.com/knative-sandbox/eventing-kafka-broker/issues/1725 where we can discuss more

Issue #1732, opened 2022-01-12
Thread blocked and broker crash after more messages 


jalonsoa:
**Describe the bug**

After tread blocked message we have more logs about session timeout, coordinator unavailable, consumer closed, etc .. and finaly a new exception and the broker stop.

```
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher Jan 12, 2022 1:55:32 PM io.vertx.core.impl.BlockedThreadChecker
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher WARNING: Thread Thread[vert.x-eventloop-thread-1,5,main] has been blocked for 13731 ms, time limit is 2000 ms
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher io.vertx.core.VertxException: Thread blocked
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher 	at java.base@17.0.1/java.time.format.DateTimeFormatterBuilder$CompositePrinterParser.format(Unknown Source)
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher 	at java.base@17.0.1/java.time.format.DateTimeFormatterBuilder$CompositePrinterParser.format(Unknown Source)
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher 	at java.base@17.0.1/java.time.format.DateTimeFormatterBuilder$CompositePrinterParser.format(Unknown Source)
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher 	at java.base@17.0.1/java.time.format.DateTimeFormatterBuilder$CompositePrinterParser.format(Unknown Source)
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher 	at java.base@17.0.1/java.time.format.DateTimeFormatter.formatTo(Unknown Source)
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher 	at java.base@17.0.1/java.time.format.DateTimeFormatter.format(Unknown Source)
kafka-broker-dispatcher-65c76bf5f-pc6s8 kafka-broker-dispatcher 	at app//io.cloudevents.types.Time.writeTime(Time.java:71)
[...]

PR #1750, opened 2022-01-17
Bounded in-flight dispatch requests in unordered delivery for back-pressure

pierDipi:
Vert.x auto-subscribe and handling of records might grow
unbounded, and it is particularly evident when the consumer
is slow to consume messages.

To apply backpressure, we need to bound the number of outbound
in-flight requests, so we need to manually poll for new records
as we dispatch them to the subscriber service.

The maximum number of outbound in-flight requests is already configurable
with the consumer parameter `max.poll.records`, and it's critical to
control the memory consumption of the dispatcher.

Part of #1732 

**Release Note**

```release-note
Apply back-pressure by limiting the number of in-flight dispatch requests in the unordered event consumption 
```


matzew (approved):
/lgtm
/approve

good call to customize this.

PR #1752, opened 2022-01-17
Fix metrics naming for Kafka clients

pierDipi:
Kafka Consumer and Producer metrics can't be scraped if we use
the prometheus naming convention since internally the micrometer
libary uses micromether specific logic to convert Kafka clients
metric names to micrometer metric names.


matzew (approved):
/lgtm
/approve

PR #1789, opened 2022-01-24
Promote some logs to info in data plane reconciler

pierDipi:
These logs are useful to watch control plane activities in
the data plane and auditing the behavior and they should be
IMO set to `info` since they are not a lot or in the hot path.


matzew (approved):
/lgtm
/approve

PR #1791, opened 2022-01-24
Add KafkaSink ConfigMap watcher for tests

pierDipi:
The CM watcher is a pod used for collecting CM changes for
debugging during tests.
We had one only for the Broker/Trigger CM, this patch adds
one for `KafkaSink`.


matzew (approved):
/lgtm
/approve

PR #1796, opened 2022-01-25
Add Pod webhook to inject contract ConfigMap for scheduling

pierDipi:
To be able to mount a `ConfigMap` per dispatcher pod, we need
to use a webhook that reacts when a _dispatcher_ `Pod` is created.

Dispatcher pods are selected based on this label:
```
app.kubernetes.io/component: kafka-dispatcher
```
(so the StatefulSets kafka-source-dispatcher,
kafka-broker-dispatcher, etc must have these labels).
and only when they are in a namespace labeled with:
```
app.kubernetes.io/name: knative-eventing
```
which is the label used for the `knative-eventing` namespace.

To every dispatcher pod, we inject a volume in `spec.volumes`
`contract-resources` which then reference a ConfigMap which
has the same name as the `Pod`.

The volume can be (statically) referenced by the `VolumeMount`
section of the (to be added) StatefulSets.

### DIFF
--- a/control-plane/cmd/webhook-kafka/main.go
+++ b/control-plane/cmd/webhook-kafka/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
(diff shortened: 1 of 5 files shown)


control-plane/cmd/webhook-kafka/main.go
```diff
@@ -77,6 +84,35 @@ func NewDefaultingAdmissionController(ctx context.Context, _ configmap.Watcher)
...
+	return defaulting.NewAdmissionController(ctx,
+		// Name of the resource webhook.
+		"pods.defaulting.webhook.kafka.eventing.knative.dev",
+
+		// The path on which to serve the webhook.
+		"/pods-defaulting",
+
+		// The resources to default.
+		// We use only defaulting callbacks for pods.
+		map[schema.GroupVersionKind]resourcesemantics.GenericCRD{},
```
aavarghese, line 103: Do we add Pod{} type to this map here?

pierDipi: No, this is done by `defaultingCallbacks` which is from `Pod` -> `custom function`.

aavarghese: Ok. Thanks!

aavarghese (approved):
/lgtm

PR #1809, opened 2022-01-27
[WIP] KafkaChannel run e2e tests with auth

aliok:
Fixes #

## Proposed Changes

- 
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/kafka/kafka_setup.sh
+++ b/test/kafka/kafka_setup.sh
@@ -103,6 +103,7 @@ function create_sasl_secrets() {
     --from-literal=user="my-sasl-user" \
     --from-literal=protocol="SASL_SSL" \
     --from-literal=sasl.mechanism="SCRAM-SHA-512" \
+    --from-literal=saslType="SCRAM-SHA-512" \
     --dry-run=client -o yaml | kubectl apply -n "${SYSTEM_NAMESPACE}" -f -
 
   kubectl create secret --namespace "${SYSTEM_NAMESPACE}" generic strimzi-sasl-plain-secret \
@@ -110,6 +111,7 @@ function create_sasl_secrets() {
     --from-literal=user="my-sasl-user" \
     --from-literal=protocol="SASL_PLAINTEXT" \
     --from-literal=sasl.mechanism="SCRAM-SHA-512" \
+    --from-literal=saslType="SCRAM-SHA-512" \
     --dry-run=client -o yaml | kubectl apply -n "${SYSTEM_NAMESPACE}" -f -
 }
 
--- a/test/kafka/kafka_setup.sh
+++ b/test/kafka/kafka_setup.sh
@@ -66,7 +66,7 @@ function create_tls_secrets() {
   ca_cert_secret="my-cluster-cluster-ca-cert"
   tls_user="my-tls-user"
 
-  echo "Waiting until secrets: ${ca_cert_secret}, ${tls_user} don't exist"
+  echo "Waiting until secrets: ${ca_cert_secret}, ${tls_user} exist"
   wait_until_object_exists secret ${ca_cert_secret} kafka
   wait_until_object_exists secret ${tls_user} kafka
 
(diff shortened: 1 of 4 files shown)


test/kafka/kafka_setup.sh
```diff
@@ -66,7 +66,7 @@ function create_tls_secrets() {
   ca_cert_secret="my-cluster-cluster-ca-cert"
   tls_user="my-tls-user"
 
-  echo "Waiting until secrets: ${ca_cert_secret}, ${tls_user} don't exist"
```
pierDipi, line 69: :man_facepalming:

test/kafka/kafka_setup.sh
```diff
@@ -80,27 +80,36 @@ function create_tls_secrets() {
...
-  echo "Creating SASL Kafka secret"
+  echo "Creating SASL_SSL and SASL_PLAINTEXT Kafka secrets"
 
   STRIMZI_CRT=$(kubectl -n kafka get secret ${ca_cert_secret} --template='{{index .data "ca.crt"}}' | base64 --decode )
   SASL_PASSWD=$(kubectl -n kafka get secret ${sasl_user} --template='{{index .data "password"}}' | base64 --decode )
 
   kubectl create secret --namespace "${SYSTEM_NAMESPACE}" generic strimzi-sasl-secret \
     --from-literal=ca.crt="$STRIMZI_CRT" \
     --from-literal=password="$SASL_PASSWD" \
-    --from-literal=saslType="SCRAM-SHA-512" \
```
pierDipi, line 102: I think this would break KafkaSource tests.

aliok: Hm, interesting...

I added these below:
```
--from-literal=protocol="SASL_SSL" \
--from-literal=sasl.mechanism="SCRAM-SHA-512" \
```

I thought these are necessary and `saslType` is actually not used.
Can't really find where `saslType` is read from the secret.

However, then on the other hand; how was it working before?

aliok: @pierDipi Yes, that change breaks KafkaSource. But, I don't see how.

pierDipi: It's hard-coded here in `eventing-kafka`,
https://github.com/knative-sandbox/eventing-kafka/blob/15a64fb7261fdb9e66bb2f7ce8171a49b6df4641/test/e2e/helpers/kafka_source.go#L326

aliok:
What I do in this PR is:
- Fix the secret creation (irrelevant to the point I would like to discuss)
- Introduce a `CHANNEL_AUTH_SCENARIO` env var.
- That env var has 4 different values and each value configures the KafkaChannel config in a special way.
- I added the 4 values in GH action test matrix. So, we now have 4 runs for every other scenario in the matrix.
- I am **planning** to create 3 additional Prow jobs for each of the e2e tests (integration-tests and reconciler-tests) that test TLS, SASL_SSL, SASL_PLAINTEXT.

Issues:
1. I didn't limit the additional runs to KafkaChannel only. So, currently all the other tests are run as well.
2. More jobs - more resources

As I wrote in a separate thread, auth is set globally until https://github.com/knative-sandbox/eventing-kafka/issues/1005 is done.
So, in order to test auth scenarios, we would need to make sure these tests are run at the end in a sequential order. Also required is to modify the configmap...
Until that issue is done, I think it is less problematic to have separate Prow jobs. I will go that route.

Now, back to issue 1 above... I think I can somehow hack the Go build tags to only run the KafkaChannel tests in the additional jobs. The question is, is it worth it...
Or, while we're at it, would we want to run the entire suites for broker, sink and source with auth? Currently only some smoke tests are done.

aliok:
> can we use a different directory for channel tests?

That actually makes sense. Instead of playing with the buildtags.

I created this PR: https://github.com/knative-sandbox/eventing-kafka-broker/pull/1813

Once that's merged, I can then create the additional Prow jobs and GH actions for auth options, but only for KafkaChannel.

aliok:
/close

Closing in favor of https://github.com/knative-sandbox/eventing-kafka-broker/pull/1816

PR #1812, opened 2022-01-27
[WIP] Enable tracing test for Channel

aliok:
Fixes #

## Proposed Changes

- Enable tracing test for Channel
-
-

**Release Note**

```release-note

```

**Docs**


aliok:
/close

Doesn't work out of the box. I will check this later.

pierDipi:
(yes, we aren't aligned with eventing core in terms of span names, etc, the metrics work that I'm doing is a little step in that direction)

PR #1813, opened 2022-01-27
Move KafkaChannel e2e and Rekt tests to different dirs

aliok:
Fixes #

## Proposed Changes

- Move KafkaChannel e2e and Rekt tests to different dirs

Motivation is being able to run channel e2e and Rekt tests independently with auth.
This would require 2 things:
1. Tooling to only run KafkaChannel tests when requested: https://github.com/knative-sandbox/eventing-kafka-broker/pull/1816
2. 3 new prow jobs for each of the auth options that run the e2e and Rekt tests for the Channel.

**Release Note**

```release-note

```

**Docs**


aliok:
cc @pierDipi

pierDipi (approved):
/lgtm

PR #1816, opened 2022-01-27
KafkaChannel run e2e tests with auth - 2nd try

aliok:
Fixes #

## Proposed Changes

Motivation is being able to run channel e2e and Rekt tests independently with auth.

- Tooling to only run KafkaChannel tests when requested.
- When `CHANNEL_AUTH_SCENARIO` env var is set, e2e and reconciler test scripts will only run KafkaChannel tests and with the given auth scenario.
- 3 new prow jobs are to be added for each of the auth options that run the e2e and Rekt tests for the Channel here: https://github.com/knative/test-infra/pull/3062

**Release Note**

```release-note
:gift: KafkaChannel is now conformant to Knative Eventing spec also while supporting auth.
```

**Docs**

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -24,6 +24,9 @@ jobs:
         eventing-config:
           - "./third_party/eventing-latest/"
 
+        channel-auth-scenario:
+          - ""
+
         test-suite:
           - ./test/e2e
           - ./test/e2e_new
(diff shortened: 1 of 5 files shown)


.github/workflows/kind-e2e.yaml
```diff
@@ -120,10 +143,15 @@ jobs:
...
         run: |
+          source test/e2e-common.sh
+
           export KO_FLAGS="--platform=linux/amd64"
           export SYSTEM_NAMESPACE=knative-eventing
           export CLUSTER_DOMAIN=${CLUSTER_SUFFIX}
 
+          # Setup auth config for KafkaChannel. Uses CHANNEL_AUTH_SCENARIO env var.
+          setup_kafka_channel_auth
+
```
pierDipi, line 154: Can we make this an `hack/run.sh` target and use that instead here?

So that we can use that target for development too.

aliok: Would you mind if I do that separately?

pierDipi: sounds good!

/lgtm

aliok:
I wanted to hold this because I realize something like this can happen:
- Say we merge this PR and then https://github.com/knative/test-infra/pull/3062
- Say there's something wrong with this PR... After https://github.com/knative/test-infra/pull/3062 is merged every new PR will be failing.

So, let's just merge https://github.com/knative/test-infra/pull/3062 first. It won't cause any failures because it just adds new jobs with new arguments. Those arguments are currently ignored (before this PR), so there won't be any issues.

After that is merge, we can re-test this PR to see if the changes are actually good.

aliok:
```
./test/presubmit-tests.sh --run-test "./test/e2e-tests.sh --channel-sasl-ssl"
```

This fails and I can't figure out why. Or, I can't figure out a very similar command in eventing-kafka repository doesn't fail:

```
 ./test/presubmit-tests.sh --run-test "./test/e2e-tests.sh --consolidated"
```

This will block all PRs on this repository.

aliok:
/close

Closing in favor of https://github.com/knative-sandbox/eventing-kafka-broker/pull/1858 and https://github.com/knative-sandbox/eventing-kafka-broker/pull/1876

PR #1840, opened 2022-02-01
[1.0] Fix Upgrade and reconciler tests

pierDipi:
Prow expects upgrade test file in `test/upgrade-tests.sh` and 
a `test/reconciler-tests.sh` file.


pierDipi:
A little PR to unblock v1.0 backports.

/assign @matzew @devguyio @aliok

test/reconciler-tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
```
aliok, line 2: I would echo some f word here :D

PR #1850, opened 2022-02-02
Consumer: check if pod is running after creating ConfigMap

pierDipi:
Check if the pod is running after trying to get or create the
associated ConfigMap, since it won't become ready until we have
created the ConfigMap.

Part of #1537


aavarghese (approved):
/lgtm

PR #1852, opened 2022-02-02
Fix KafkaSource v2 controller handlers

pierDipi:
`FilteringResourceEventHandler` doesn't work without a
filter function and we don't need to filter Kafka source in
the Kafka source reconciler.

Part of #1537


aavarghese (approved):
Thanks!!
/lgtm

PR #1880, opened 2022-02-08
Use Git patches for pkg issue and E2E tests failures

pierDipi:
See the linked issue in `TODO: ...`

this is causing errors like:
```
channel_impl.go:76: admission webhook "validation.webhook.kafka.eventing.knative.dev" denied the request: validation failed: invalid value: : spec.retentionDuration
```
In this case, the pod webhook is the leader over the resource webhook.


aliok (approved):
/lgtm

PR #1901, opened 2022-02-15
Add Consumer reply validation

pierDipi:
Add enabled flag to distinguish between the enabled
reply strategy.

Update KafkaSource reconciler to set the reply strategy
to `NoReply`.

Update Trigger reconciler to set the reply strategy
to `TopicReply`.

Part of #1537

## Proposed Changes

- Add Consumer reply validation


aavarghese (approved):
/lgtm

PR #1909, opened 2022-02-15
Upgrade to latest dependencies and fix KafkaChannel tests

pierDipi:
Fixes https://github.com/knative-sandbox/eventing-kafka-broker/pull/1843

- Upgrade to latest dependencies
- Fix KafkaChannel reconciler tests


matzew:
wat the test!

/retest

/go-in!

pierDipi:
There is a real problem with this update:
```
request: validation failed: Immutable fields changed (-old +new): spec
        {v1beta1.KafkaChannelSpec}.NumPartitions:
        	-: "10"
        	+: "1"
```

matzew:
@pierDipi 
Option 1) reviert and do nothing
Option 2) revert the default to old/nasty settings and override them only _here_
Option 3) look for conversion/migration, before we cut for 1.3

I think I'd go with 1) now, and do 2) once we can...

pierDipi:
> @pierDipi Option 1) reviert and do nothing Option 2) revert the default to old/nasty settings and override them only _here_ Option 3) look for conversion/migration, before we cut for 1.3
> 
> I think I'd go with 1) now, and do 2) once we can...

Revert what?

pierDipi:
> > Revert what?
> 
> the default changes to 10/3 (instead of 1/1)

Yes, but these defaults should only affect channels without any partitions or replication factor set

matzew:
/close

will manually follow up

PR #1920, opened 2022-02-16
Fix for no label selector in cg causing reusing consumer between kafkasource and trigger objects

aavarghese:

Part of https://github.com/knative-sandbox/eventing-kafka-broker/issues/1537

Fixes issue described in title. More info in https://github.com/knative-sandbox/eventing-kafka-broker/issues/1537#issuecomment-1042380156

### DIFF
--- a/control-plane/pkg/reconciler/source/v2/sourcev2.go
+++ b/control-plane/pkg/reconciler/source/v2/sourcev2.go
@@ -101,6 +101,11 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, ks *sources.Kafk
 		},
 		Spec: internalscg.ConsumerGroupSpec{
 			Template: internalscg.ConsumerTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"eventing.knative.dev/sourceUID": string(ks.UID),
+					},
+				},
 				Spec: internalscg.ConsumerSpec{
 					Topics: ks.Spec.Topics,
 					Configs: internalscg.ConsumerConfigs{Configs: map[string]string{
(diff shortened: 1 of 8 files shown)


matzew:
the unit tests fail b/c the new labels.One example:
```
          				DeletionGracePeriodSeconds: nil,
        - 				Labels:                     nil,
        + 				Labels:                     map[string]string{"eventing.knative.dev/sourceUID": "uuid"},

```

control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -101,6 +101,11 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, ks *sources.Kafk
 		},
 		Spec: internalscg.ConsumerGroupSpec{
 			Template: internalscg.ConsumerTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"eventing.knative.dev/sourceUID": string(ks.UID),
```
pierDipi, line 106: nit: can we use `kafka.eventing.knative.dev/source.uid` ? can we extract it as constant?

should we add a label for `cg` to the trigger reconciler too? maybe in a different PR?

aavarghese: @pierDipi sounds good. I added as constant[ here ](https://github.com/knative-sandbox/eventing-kafka-broker/pull/1920/files#diff-678e94010cb81bf07cc8bf28ed64cc0c94c44541d7398c96f2f63296bf0ac7cfR28)

Label for trigger reconciler is already in this PR. I'll update the PR title to reflect that. Is that ok?

pierDipi (approved):
/lgtm
/approve

Thanks!

Issue #1763, opened 2022-01-20
Implement the new trigger filters experimental feature


devguyio:
### Description

Eventing landed a new experimental feature called "new-trigger-filters" which adds a new `filters` field in the Trigger spec that conforms to the CNCF CouldEvents Subscriptions API filters field. It support more advanced filters via a set if filter dialect. 

The kafka broker should implement the new experimental feature as defined by the feature track found in knative/eventing#5204

### [Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md) 
Event Consumer (developer)

PR #1921, opened 2022-02-17
Introducing dialected-filters to the contract for supporting new-trigger-filters experimental feature

devguyio:
Part of #1763 

This PR introduces dialected-filters to the egress contract as a first step of implementing the new-trigger-filers experimental feature.

## Proposed Changes

- Add dialected filters to egress contract


matzew (approved):
/lgtm
/approve

I like the commit message details

proto/contract.proto
```diff
@@ -11,6 +11,48 @@ option java_outer_classname = "DataPlaneContract";
...
+
+message Prefix {
+  string attribute = 1;
+  string prefix = 2;
+}
+
+message Suffix {
+  string attribute = 1;
+  string suffix = 2;
+}
```
pierDipi, line 26: I'm reading [this example in spec](https://github.com/cloudevents/spec/blob/main/subscriptions/spec.md#filters):
```
[ {"prefix": { "type": "com.github.issue" } } ]
```

why doesn't this need to be a `map<string, string>`?

In addition, at this layer, I would design for extensibility here.

pierDipi: I see why:
> Use of this MUST include exactly one nested property

It seems weird that exact supports multiple attributes while prefix and suffix don't

PR #1932, opened 2022-02-21
Consumer reconciler sets ConfigMap owner reference to dispatcher pod

pierDipi:
When a consumer is removed and its binding dispatcher pod also
gone (autoscaled down), the config map is cleaned up.

Part of #1537


aavarghese (approved):
Thank you @pierDipi !!!!

/lgtm

PR #1957, opened 2022-03-02
Global resync consumers and consumergroups on statefulset changes

pierDipi:
To keep resource statuses in sync with the actual current state, we
need to global resync consumer groups when the statefulsets changes.

Part of #1537

### DIFF
--- a/control-plane/pkg/reconciler/consumergroup/controller.go
+++ b/control-plane/pkg/reconciler/consumergroup/controller.go
@@ -25,12 +25,15 @@ import (
 
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
+	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apiserver/pkg/storage/names"
+	"k8s.io/client-go/tools/cache"
 	"knative.dev/eventing/pkg/scheduler"
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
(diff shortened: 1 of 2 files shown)


control-plane/pkg/reconciler/consumergroup/controller.go
```diff
@@ -99,12 +102,32 @@ func NewController(ctx context.Context) *controller.Impl {
 	impl := cgreconciler.NewImpl(ctx, r)
 	consumerInformer := consumer.Get(ctx)
 
+	consumerGroupInformer := consumer.Get(ctx)
```
aavarghese, line 105: Same as line 103 with diff name

control-plane/pkg/reconciler/consumergroup/controller.go
```diff
@@ -99,12 +102,32 @@ func NewController(ctx context.Context) *controller.Impl {
...
+
 	consumergroup.Get(ctx).Informer().AddEventHandler(controller.HandleAll(impl.Enqueue))
 	consumerInformer.Informer().AddEventHandler(controller.HandleAll(enqueueConsumerGroupFromConsumer(impl.EnqueueKey)))
 
+	globalResync := func(interface{}) {
+		impl.GlobalResync(consumerGroupInformer.Informer())
```
aavarghese, line 111: can change to `consumerInformer` from line 103

pierDipi: This is the consumer group reconciler isn't consumerGroupInformer what we should use?

aavarghese: Yes yes. Misread the field. Changes on 105 looks good.

aavarghese (approved):
/lgtm

PR #1965, opened 2022-03-07
Failed to build storage version migrator

pierDipi:
As per title.

storage version migrator wasn't included in the tools.go file:
```
_ "knative.dev/pkg/apiextensions/storageversion/cmd/migrate"
```

### DIFF
--- a/hack/tools.go
+++ b/hack/tools.go
@@ -24,6 +24,9 @@ import (
 	_ "knative.dev/pkg/configmap/hash-gen"
 	_ "knative.dev/pkg/hack"
 
+	// Needed for the storage version migration.
+	_ "knative.dev/pkg/apiextensions/storageversion/cmd/migrate"
+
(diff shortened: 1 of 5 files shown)


hack/tools.go
```diff
@@ -24,6 +24,9 @@ import (
 	_ "knative.dev/pkg/configmap/hash-gen"
 	_ "knative.dev/pkg/hack"
 
+	// Needed for the storage version migration.
+	_ "knative.dev/pkg/apiextensions/storageversion/cmd/migrate"
```
matzew, line 28: :facepalm:

matzew (approved):
/lgtm
/approve

hack/tools.go
pierDipi: I added a test for these kinds of problems: https://github.com/knative-sandbox/eventing-kafka-broker/pull/1976

PR #1968, opened 2022-03-07
Add support for ARM64

### LINKED ISSUE #1908: knative-kafka-broker-receiver is not published for aarch64/arm64
**Describe the bug**

While all other containers are available for aarch64/arm64, `gcr.io/knative-releases/knative-kafka-broker-receiver` is not.

 **Expected behavior**

Image should be available for `aarch64` too.

 **To Reproduce**

Deploy knative-eventing on aarch64.

 **Knative release version** 0.26.5

 **Additional context**

```
    State:       Waiting
      Reason:    CrashLoopBackOff
    Last State:  Terminated
      Reason:    Error
      Message:   standard_init_linux.go:228: exec user process caused: exec format error

      Exit Code:    1
```

pierDipi:
Fixes #1908

To build ARM64 images I'm using Jib in this PR.

I have investigated 2 options:

1. docker/podman buildx
     - advantages: 
       - keep using docker/podman for the build
       - (unclear if it would still possible to use this) smaller images since we can keep using jlink to reduce the image size (1/2x)
     - disadvantages:
       - complex build (additional bash scripting) because some docker/podman drivers don't natively support multi-arch builds (like mine) so we need to manually do multiple builds and publish the final multi-arch manifest
2. jib
     - advantages: 
       - simpler multi-arch build
       - reduces bash scripting
       - removes the dependency on docker for building and for local development (`jib` is `ko` like tool)
     - disadvantages:
       - optional additional maven plugin dependencies
       - bigger images (2x)

## Proposed Changes

- Add support for ARM64

**Release Note**

```release-note
Add support for ARM64.
```


pierDipi:
This is ready for review, I'm adding hold to wait for the release

/hold

aliok:
> disadvantages:
> complex build (additional bash scripting) because some docker/podman drivers don't natively support multi-arch builds (like mine) so we need to manually do multiple builds and publish the final multi-arch manifest

I don't think we're at this stage to deal these kind of things in this project. Let's go with the simplest option and we can optimize later.

So, IMO, jib is better.

aliok:
Maybe the productization people in downstream Red Hat won't like it, but I don't have too many concerns.

.github/workflows/kind-e2e.yaml
```diff
@@ -66,7 +66,8 @@ jobs:
 
     env:
       GOPATH: ${{ github.workspace }}
-      KO_DOCKER_REPO: kind.local
+      KO_DOCKER_REPO: localhost:5001
+      REGISTRY_PORT: 5001
```
aliok, line 70: Why this change was necessary?

data-plane/pom.xml
```diff
@@ -349,6 +351,28 @@
...
+            <allowInsecureRegistries>true</allowInsecureRegistries>
+            <from>
+              <image>gcr.io/distroless/java17-debian11:nonroot</image>
+              <platforms>
+                <platform>
+                  <architecture>amd64</architecture>
+                  <os>linux</os>
+                </platform>
+                <platform>
+                  <architecture>arm64</architecture>
```
aliok, line 369: Awesome!

hack/create-kind-cluster.sh
```diff
@@ -22,11 +22,25 @@ set -o pipefail
...
 NODE_SHA=${NODE_SHA:-"sha256:0df8215895129c0d3221cda19847d1296c4f29ec93487339149333bd9d899e5a"}
+REGISTRY_NAME=${REGISTRY_NAME:-"kind-registry"}
+REGISTRY_PORT=${REGISTRY_PORT:-"5001"}
+
+# create registry container unless it already exists
+if [ "$(docker inspect -f '{{.State.Running}}' "${REGISTRY_NAME}" 2>/dev/null || true)" != 'true' ]; then
+  docker run \
+    -d --restart=always -p "127.0.0.1:${REGISTRY_PORT}:5000" --name "${REGISTRY_NAME}" \
+    docker.io/registry:2
+fi
```
aliok, line 33: Similar to my question above, I didn't get the reason for this

hack/create-kind-cluster.sh
```diff
@@ -48,3 +62,21 @@ nodes:
   image: kindest/node:${NODE_VERSION}@${NODE_SHA}
 EOF
 
+# connect the registry to the cluster network if not already connected
+if [ "$(docker inspect -f='{{json .NetworkSettings.Networks.kind}}' "${REGISTRY_NAME}")" = 'null' ]; then
+  docker network connect "kind" "${REGISTRY_NAME}"
+fi
```
aliok, line 68: Boy you must've spent a lot of time on this, thanks for that!

.github/workflows/kind-e2e.yaml
pierDipi: Because `kind.local` is `ko` only feature and it does some magic with `KinD` (it basically does the equivalent of `kind load image ...`), so we need a proper local container registry

hack/create-kind-cluster.sh
pierDipi: I copy-pasted logic from the documentation https://kind.sigs.k8s.io/docs/user/local-registry/

hack/create-kind-cluster.sh
pierDipi: See https://github.com/knative-sandbox/eventing-kafka-broker/pull/1968#discussion_r822713028

pierDipi:
> Maybe the productization people in downstream Red Hat won't like it, but I don't have too many concerns.

This is optional, we can still produce the final fat jar

aliok:
LGTM

data-plane/config/channelv2/500-dispatcher.yaml
```diff
@@ -99,12 +99,6 @@ spec:
             # This should be set according to initial delay seconds
             - name: WAIT_STARTUP_SECONDS
               value: "8"
-          command:
-            - "java"
-          args:
-            - "-Dlogback.configurationFile=/etc/logging/config.xml"
```
matzew, line 105: any impact on removing these?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/Main.java
```diff
@@ -52,6 +52,10 @@
 
 public class Main {
 
+  static {
+    System.setProperty("logback.configurationFile", "/etc/logging/config.xml");
```
matzew, line 56: Ah, here we go!

hack/create-kind-cluster.sh
matzew: wanna give credit from there this is?

matzew (approved):
/lgtm
/approve

edude03:
Off topic, but I wanted to say thanks for this @pierDipi, I cloned this PR and tried it locally. It works great and you saved my knative spike - we (the engineering team) just switched to M1 macs and it's been a major headache, so I appreciate how smooth it was to `run.sh deploy` and get back to work.

pierDipi:
I rebased due to conflicts.

edude03:
Heh this e2e suite seems a bit brutal. 

Off topic as well - but I have this issue in real clusters as well

```
tracker.go:137: Failed to clean the resource "e2e-kafka-source-change-sink" : Internal error occurred: failed calling webhook "validation.webhook.kafka.eventing.knative.dev": Post "[https://kafka-webhook-eventing.knative-eventing.svc:443/resource-validation?timeout=2s](https://kafka-webhook-eventing.knative-eventing.svc/resource-validation?timeout=2s)": dial tcp 10.96.154.143:443: connect: connection refused
[3693](https://github.com/knative-sandbox/eventing-kafka-broker/runs/6302949741?check_suite_focus=true#step:12:3693)
```

If I install eventing then delete its namespace it gets stuck like this, should I open an issue for that?

haubenr:
@pierDipi Can you give an update on the status of this PR? Is there anything blocking you from getting it into a mergeable state? I need this PR to be merged soon as I want to piggyback on it's changes to enable the multi-arch image builds for s390x.

pierDipi:
I rebased the PR, PTAL @matzew @aliok @aavarghese @aslom

pierDipi:
Flake
/test integration-tests_eventing-kafka-broker_main

matzew:
LGTM

leaving for @aliok for final approval

aliok (approved):
/lgtm
/approve

colvint:
Trying to sort out whether the ARM64 changes are in release 1.5.3. It would seem so but applying with:

```bash
kubectl apply --filename https://github.com/knative-sandbox/eventing-kafka-broker/releases/download/knative-v1.5.3/eventing-kafka-source.yaml
```

But the `kafka-source-dispatcher` pod still goes into `CrashLoopBackOff` due to error:

```
 {"@timestamp":"2022-06-27T01:16:27.737Z","@version":"1","message":"Failed to startup the dispatcher","logger_name":"dev.knative.eventing.kafka.broker.dispatcher.main.Main","thread_name" │
│ :"main","level":"ERROR","level_value":40000,"stack_trace":"java.io.IOException: Function not implemented\n\tat java.base/sun.nio.fs.LinuxWatchService.<init>(Unknown Source)\n\tat java.b │
│ ase/sun.nio.fs.LinuxFileSystem.newWatchService(Unknown Source)\n\tat dev.knative.eventing.kafka.broker.core.file.FileWatcher.start(FileWatcher.java:88)\n\tat dev.knative.eventing.kafka. │
│ broker.dispatcher.main.Main.main(Main.java:122)\n"}                                                                                                                                       │
```

Interestingly, after 106 restarts, the crashes stopped.

colvint:
> Hi @colvint, this wasn't merged in 1.5.x, so it will be available in the next release 1.6.0.
> 
> In the meantime, you can use [nightly builds](https://console.cloud.google.com/storage/browser/knative-nightly/eventing-kafka-broker/latest?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false) (unreleased builds).

Ah -- perfect. Thanks for the quick response @pierDipi !

data-plane/config/broker/500-receiver.yaml
```diff
@@ -114,13 +114,8 @@ spec:
             # This should be set according to initial delay seconds
             - name: WAIT_STARTUP_SECONDS
               value: "8"
-          command:
```
matzew, line 117: Just curious, @pierDipi: why was this change needed for the introduction of ARM support

pierDipi: that's how Jib works, it injects the entrypoint for the container, so there is no need to call java directly, also because the way Jib works is that it creates a directory that is the classpath with all the dependencies jar and the actual jar is separate, so the actual `java ...` command is a little more complicated than we had (and it wouldn't work downstream as well)

PR #1983, opened 2022-03-09
Consolidate Kafka packages into one single package

pierDipi:
These 2 packages `pkg/reconciler/kafka` and `pkg/kafka` always
confuse me about their purpose, so I think it makes more sense to
keep one single package for "Kafka" related functions.


control-plane/pkg/apis/eventing/v1/broker_validation.go
```diff
@@ -29,7 +29,7 @@ import (
 	"knative.dev/pkg/webhook"
 	"knative.dev/pkg/webhook/resourcesemantics/validation"
 
-	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/kafka"
```
matzew, line 32: +1 that is confusing, these

matzew:
will take a :eye: 
/assign

matzew (approved):
/lgtm
/approve

thanks for doing the move!

PR #1971, opened 2022-03-07
Use broker status annotations to stamp ConfigMap reference

pierDipi:
Instead of adding a finalizer to the ConfigMap, we can record
the ConfigMap data in the Broker status so that we can re-build
the ConfigMap even when the ConfigMap is deleted first.

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>

Fixes #1970

**Release Note**

```release-note
Support brokers with long namespace and name values.
```

PR #1987, opened 2022-03-09
Remove ConfigMap finalizer during reconcile

pierDipi:
We removed the "add finalizer to the ConfigMap logic" in
https://github.com/knative-sandbox/eventing-kafka-broker/pull/1971
and we kept the "remove the ConfigMap finalizer" in `FinalizeKind`.

However, in order to remove the ConfigMap finalizer from every
existing Broker that won't be deleted in the time span going from
1.3 to whatever release we remove the logic to remove the finalizer,
we should remove the finalizer in `ReconcileKind` too.


matzew (approved):
/lgtm
/approve

PR #1999, opened 2022-03-14
Add consumer group to logging output

jhill072:
While troubleshooting problematic consumers (for example, that return malformed event data or are lagging behind), it would be helpful to know which particular consumer a record fails on. With the current log structure, we only receive the partition/offset of a record, but the same record can be dispatched to many different consumers.

## Proposed Changes

- :broom: Add consumer group as a key to logs in the RecordDispatcher

**Release Note**

```release-note
Log output from the record dispatcher now contains the consumer group as a key for better tracing of a record.
```

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
@@ -225,7 +225,7 @@ public class RecordDispatcherImpl implements RecordDispatcher {
   }
 
   private void onFilterMatching(final ConsumerRecordContext recordContext, final Promise<Void> finalProm) {
-    logDebug("Record matched filtering", recordContext.getRecord());
+    logTrace("Record matched filtering", recordContext.getRecord());
     subscriberSender.apply(recordContext.getRecord())
       .onSuccess(response -> onSubscriberSuccess(response, recordContext, finalProm))
       .onFailure(ex -> onSubscriberFailure(ex, recordContext, finalProm));
@@ -233,7 +233,7 @@ public class RecordDispatcherImpl implements RecordDispatcher {
 
   private void onFilterNotMatching(final ConsumerRecordContext recordContext,
                                    final Promise<Void> finalProm) {
-    logDebug("Record did not match filtering", recordContext.getRecord());
+    logTrace("Record did not match filtering", recordContext.getRecord());
     recordDispatcherListener.recordDiscarded(recordContext.getRecord());
     finalProm.complete();
   }
@@ -241,7 +241,7 @@ public class RecordDispatcherImpl implements RecordDispatcher {
   private void onSubscriberSuccess(final HttpResponse<?> response,
                                    final ConsumerRecordContext recordContext,
                                    final Promise<Void> finalProm) {
-    logDebug("Successfully sent event to subscriber", recordContext.getRecord());
+    logTrace("Successfully sent event to subscriber", recordContext.getRecord());
 
     incrementEventCount(response, recordContext);
     recordDispatchLatency(response, recordContext);
@@ -265,7 +265,7 @@ public class RecordDispatcherImpl implements RecordDispatcher {
 
   private void onDeadLetterSinkSuccess(final ConsumerRecordContext recordContext,
                                        final Promise<Void> finalProm) {
-    logDebug("Successfully sent event to the dead letter sink", recordContext.getRecord());
+    logTrace("Successfully sent event to the dead letter sink", recordContext.getRecord());
     recordDispatcherListener.successfullySentToDeadLetterSink(recordContext.getRecord());
     finalProm.complete();
   }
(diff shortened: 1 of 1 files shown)


pierDipi (commented):
Thank you!

There are compilation errors:
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.0:compile (default-compile) on project dispatcher: Compilation failure: Compilation failure: 
Error:  /home/runner/work/eventing-kafka-broker/eventing-kafka-broker/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java:[268,5] non-static method logDebug(java.lang.String,io.vertx.kafka.client.consumer.KafkaConsumerRecord<java.lang.Object,io.cloudevents.CloudEvent>) cannot be referenced from a static context
Error:  /home/runner/work/eventing-kafka-broker/eventing-kafka-broker/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java:[277,24] non-static method logError(java.lang.String,io.vertx.kafka.client.consumer.KafkaConsumerRecord<java.lang.Object,io.cloudevents.CloudEvent>,java.lang.Throwable) cannot be referenced from a static context
Error:  /home/runner/work/eventing-kafka-broker/eventing-kafka-broker/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java:[280,28] non-static method logError(java.lang.String,io.vertx.kafka.client.consumer.KafkaConsumerRecord<java.lang.Object,io.cloudevents.CloudEvent>,java.lang.Throwable) cannot be referenced from a static context
```

jhill072:
Sorry about that, my STS is bugging out and thought I could get this visually 😅

jhill072:
Seems like there is a linting issue from the control plane? Is that something I should be concerned with?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
```diff
@@ -333,41 +333,44 @@ private Tags getTags(final ConsumerRecordContext recordContext) {
...
 
-    logger.debug(msg + " {} {} {} {} {} {}",
+    logger.debug(msg + " {} {} {} {} {} {} {}",
       keyValue("topic", record.topic()),
       keyValue("partition", record.partition()),
       keyValue("headers", record.headers()),
       keyValue("offset", record.offset()),
       keyValue("key", record.key()),
-      keyValue("event", record.value())
+      keyValue("event", record.value()),
```
matzew, line 372: not sure on `debug` for the actual value.

I think recently @devguyio had a good point about doing this at `trace` level (e.g. security audits may not like the value to be flying around on `debug`)

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
```diff
@@ -333,41 +333,44 @@ private Tags getTags(final ConsumerRecordContext recordContext) {
...
     final Throwable cause) {
 
     if (logger.isDebugEnabled()) {
-      logger.error(msg + " {} {} {} {} {}",
+      logger.error(msg + " {} {} {} {} {} {}",
         keyValue("topic", record.topic()),
         keyValue("partition", record.partition()),
         keyValue("headers", record.headers()),
         keyValue("offset", record.offset()),
         keyValue("event", record.value()),
```
matzew, line 347: I am supportive on adding the consumer group.

but perhaps we can set the above to `trace` as well, since we do already log the value. I know - not introduced by you :sweat_smile:

matzew:
@jhill072 thanks for the PR! :heart_eyes_cat:

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
pierDipi: isn't it a different issue?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
jhill072: I don't mind, I'd rather make the updates if it is something that I am touching on. Rather see less tech debt if we can make the improvements while we are in the area!

I'll look to make the debug->trace updates later this afternoon, thanks for the feedback!

jhill072: Took longer than I wanted, but debug -> trace is added now :+1:

matzew (approved):
/lgtm
/approve

@jhill072 I have rebased your branch

matzew:
@jhill072 do you have a CLA? https://github.com/knative-sandbox/eventing-kafka-broker/pull/1999#issuecomment-1223618988

matzew:
Again... namespaced ... /cc @aliok 

```
timed out waiting for the condition
```

/test reconciler-tests-namespaced-broker_eventing-kafka-broker_main

aliok:
I think Tide (https://prow.knative.dev/tide) is messed up :(

This PR seems to be the first item in the merge pool for the past 2 days. And other PRs can't be merged before this one.

I am gonna put a hold on this to see what happens.

aliok:
Ok, hold didn't work. Gonna close and reopen to see what happens.

pierDipi:
I've reworked the logging part to include the context (consumer group, resource name, resource namespace, etc) in every log statement, let us know if that addresses your issue.

PR #2003, opened 2022-03-15
Poll for new records when `max.poll.records` limit isn't reached

pierDipi:
By only polling for new records when the aggregate future for the
entire batch completes we are waiting too much on the responses
(which is slow).

Since the goal is to avoid having too many in-flight requests
we can poll for new records as soon as the number of in-flight requests
is under `max.poll.records`.

We are not forcing the dispatcher to send less than `max.poll.records`
requests because we don't want to keep records in memory by waiting
for responses, which means that we might have no more than
`2 * max.poll.records` requests in flight.

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/UnorderedConsumerVerticle.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/UnorderedConsumerVerticle.java
@@ -21,6 +21,7 @@ import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.kafka.client.consumer.KafkaConsumerRecords;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -41,10 +42,20 @@ public final class UnorderedConsumerVerticle extends BaseConsumerVerticle {
   // This shouldn't be more than 2000, which is the default max time allowed
   // to block a verticle thread.
   private static final Duration POLL_TIMEOUT = Duration.ofMillis(1000);
+  private final int maxPollRecords;
 
-  public UnorderedConsumerVerticle(Initializer initializer,
-                                   Set<String> topics) {
+  private int inFlightRecords;
+
+  public UnorderedConsumerVerticle(final Initializer initializer,
+                                   final Set<String> topics,
+                                   final int maxPollRecords) {
     super(initializer, topics);
+    if (maxPollRecords <= 0) {
+      this.maxPollRecords = 500;
+    } else {
+      this.maxPollRecords = maxPollRecords;
+    }
+    this.inFlightRecords = 0;
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/UnorderedConsumerVerticle.java
```diff
@@ -41,10 +42,20 @@
   // This shouldn't be more than 2000, which is the default max time allowed
   // to block a verticle thread.
   private static final Duration POLL_TIMEOUT = Duration.ofMillis(1000);
+  private final int maxPollRecords;
 
-  public UnorderedConsumerVerticle(Initializer initializer,
-                                   Set<String> topics) {
+  private int inFlightRecords;
+
+  public UnorderedConsumerVerticle(final Initializer initializer,
```
matzew, line 49: perhaps we add some javadoc on the arguments?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/UnorderedConsumerVerticle.java
```diff
@@ -81,11 +102,17 @@ private void poll() {
...
+      this.inFlightRecords++;
+      this.recordDispatcher.dispatch(records.recordAt(i))
+        .onComplete(v -> {
+          this.inFlightRecords--;
+          poll();
+        });
     }
-    CompositeFuture.join(futures)
-      .onComplete(v -> poll());
+    poll();
```
matzew, line 116: so we call the blocking `poll` now inner and outer loop, to ensure the inflight and max are in balance, before we call the actual consumer's poll?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/UnorderedConsumerVerticle.java
pierDipi: Sure

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/UnorderedConsumerVerticle.java
pierDipi: Yes, `poll` itself decides when to poll or wait for another call done by the handler in `onComplete()`.

nit: the `poll` is non-blocking.

matzew: > itself decides when to poll or wait for 
ok, should we perhaps reflect that in the name of the method? `poll` seems very generic - but the method does more.

Sorry for the bikesheeding :sweat_smile:

pierDipi: `maybePoll`? :smile:

matzew (approved):
/lgtm
/approve

PR #2006, opened 2022-03-16
Config retries for KafkaSource scheduler based

pierDipi:
As per title.

These values are the same we use in the existing
KafkaSource in this repository

### DIFF
--- a/control-plane/pkg/reconciler/source/v2/sourcev2.go
+++ b/control-plane/pkg/reconciler/source/v2/sourcev2.go
@@ -126,8 +126,8 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, ks *sources.Kafk
 						DeliverySpec: &eventingduck.DeliverySpec{
 							Retry:         pointer.Int32(10),
 							BackoffPolicy: &backoffPolicy,
-							BackoffDelay:  pointer.String("PT10S"), // 10 seconds
-							Timeout:       pointer.String("PT500S"),
+							BackoffDelay:  pointer.String("PT10S"),
+							Timeout:       pointer.String("PT600S"),
 						},
 						Ordering: DefaultDeliveryOrder,
 					},
--- a/control-plane/pkg/reconciler/source/v2/sourcev2.go
+++ b/control-plane/pkg/reconciler/source/v2/sourcev2.go
@@ -24,7 +24,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 	sources "knative.dev/eventing-kafka/pkg/apis/sources/v1beta1"
+	eventingduck "knative.dev/eventing/pkg/apis/duck/v1"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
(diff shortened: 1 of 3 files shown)


control-plane/pkg/reconciler/source/v2/sourcev2.go
```diff
@@ -119,6 +123,12 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, ks *sources.Kafk
 						NetSpec: &ks.Spec.KafkaAuthSpec.Net,
 					},
 					Delivery: &internalscg.DeliverySpec{
+						DeliverySpec: &eventingduck.DeliverySpec{
+							Retry:         pointer.Int32(10),
+							BackoffPolicy: &backoffPolicy,
+							BackoffDelay:  pointer.String("PT10S"), // 10 seconds
+							Timeout:       pointer.String("PT500S"),
```
aavarghese, line 130: Is this too long of a `Timeout`?

aavarghese: Ah current `KafkaSource` has a timeout 0s. We should change that too?

pierDipi: Hey Ansu, I was talking with @matzew  and the idea is to set timeouts to 600s, I'll update the PR accordingly.

I think 0s (is that no timeout ?) is really bad.

pierDipi: Why is that 600s good?

We want that most of the people using this timeout shouldn't need to update this value (and KafkaSource has no config to change it) and, also, because Knative services on big and slow clusters, just to start take ~30 seconds.

pierDipi: cc @devguyio in case you want to add your opinion.

matzew: yes, I agree with a loooooonger timeout - thanks, for bringing this up.

and yes, 600 seems to be a good number, and also: yes, no itention to have users change that a lot.

And yes, `0` is really really bad.

Thanks @aavarghese and @pierDipi

pierDipi: Done!

aslom: Network timeout for TCP/IP are in that range so it seems very reasonable to me to have similiar timeout for eventing delivery for Linux for example is 924 seconds: https://pracucci.com/linux-tcp-rto-min-max-and-tcp-retries2.html
It would be good idea to check and set that default for all Knative Eventing components?

aavarghese (approved):
/lgtm

control-plane/pkg/reconciler/source/v2/sourcev2.go
pierDipi: > It would be good idea to check and set that default for all Knative Eventing components?

+1, can we keep track of this in an issue in knative/eventing?

pierDipi:
@aavarghese sorry, I had to fix unit tests :upside_down_face: would you mind reviewing the last commit?

control-plane/pkg/reconciler/source/v2/sourcev2.go
aslom: Opened issue: https://github.com/knative/eventing/issues/6278

aavarghese (approved):
Unit tests passed.

/lgtm

PR #2047, opened 2022-03-31
HA Scheduler config map update

aavarghese:

`EvenPodSpread` is specified to schedule vreplicas evenly across dispatcher pods which is very important for an equal event consumption from Kafka topic. 

`EvenPodSpread` can both be specified as a filter (predicate) OR as a scorer (priority). When specified as a predicate, the scheduler is highly constrained by the skew factor and requires a large number of pods (in some scenarios) which is not very friendly for small clusters (pods that have low capacity are filtered out and not used). Instead, we can use this plugin to score pods and schedule onto higher score pods. 

Changed `EvenPodSpread` to a priority instead of a predicate for the default config scheduler in this repo. 
Either can work depending on the cluster and the requirements.

Part of #1537 

/cc @pierDipi


pierDipi (approved):
/lgtm

PR #2048, opened 2022-04-01
Remove `vertx-stack-depchain`

pierDipi:
We don't need this additional dependency.


matzew (approved):
/lgtm
/approve

PR #2057, opened 2022-04-06
Increase default timeout to 10 minutes

pierDipi:
As discussed previously, the default timeout of 10 seconds isn't
sometimes enough to even start a Knative Service.

Using 10 minutes will avoid overriding the timeout with
`spec.delivery.timeout` in most of the use cases.

## Proposed Changes

- Increase default timeout to 10 minutes

**Release Note**

```release-note
The default request timeout is 10 minutes now, it was previously set to 10 seconds.
It can be overridden using `spec.delivery.timeout` on `Broker`, `Trigger`, `KafkaChannel` and `Subscription`.
```

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
@@ -85,6 +85,8 @@ public class ConsumerVerticleFactoryImpl implements ConsumerVerticleFactory {
 
   private static final Logger logger = LoggerFactory.getLogger(ConsumerVerticleFactoryImpl.class);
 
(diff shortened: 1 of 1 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
```diff
@@ -85,6 +85,8 @@
 
   private static final Logger logger = LoggerFactory.getLogger(ConsumerVerticleFactoryImpl.class);
 
+  private static final long DEFAULT_TIMEOUT_MS = 600_000L;
```
matzew, line 88: would something like `TimeUnit.MILLISECONDS.convert(10L, TimeUnit.MINUTES)` be more readable?

matzew (approved):
/lgtm
/approve

PR #2058, opened 2022-04-06
Dynamically set `max.poll.interval.ms` based on the delivery spec

pierDipi:
`max.poll.interval.ms` is used to bound the processing time at
the consumer level, however, we have application-level timeouts
that are dinamically controlled.

In this patch, I'm performing the maximum processing time (
see how in `maxProcessingTimeMs` method) and set
`max.poll.interval.ms` to that value.

## Proposed Changes

- Dynamically set `max.poll.interval.ms` based on the delivery spec

**Release Note**

```release-note
Dynamically set `max.poll.interval.ms` based on the delivery spec
```


matzew (approved):
This does make sense to bind this more dynamic

/lgtm
/approve

matzew:
For 1.2 I did https://github.com/knative-sandbox/eventing-kafka-broker/pull/2064

PR #2067, opened 2022-04-08
Handle cases when there is no committed offset

pierDipi:
Calling `committed()` on a fresh consumer produces no
`OffsetAndMetadata`, so we shouldn't rely on this to get the
committed offset.


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/OffsetManager.java
```diff
@@ -141,7 +139,9 @@ private void commit(final KafkaConsumerRecord<?, ?> record) {
       // Reset the state
       tracker.setCommitted(newOffset);
 
-      logger.debug("Committing offset for {} offset {}", topicPartition, newOffset);
+      logger.info("Committing offset for {} offset {}",
```
matzew, line 142: do we really want this to be `info` ?

matzew: Otherwise

LGTM

pierDipi: Done

matzew (approved):
/lgtm
/approve

PR #2091, opened 2022-04-13
Initialize offset in ConsumerGroup reconciler

pierDipi:
When the offset strategy is set to latest, we need to initialize
the offset in Kafka for that particular consumer group.

Part of #1537


aavarghese (approved):
/lgtm

PR #2092, opened 2022-04-13
Topology spread constraints on zones and anti-affinity for receivers and dispatchers

pierDipi:
To spread receivers and dispatchers across zones for HA.

Part of #1537 and Broker HA

### DIFF
--- a/data-plane/config/broker/500-dispatcher.yaml
+++ b/data-plane/config/broker/500-dispatcher.yaml
@@ -33,6 +33,23 @@ spec:
         app: kafka-broker-dispatcher
         kafka.eventing.knative.dev/release: devel
     spec:
+      # To avoid node becoming SPOF, spread our replicas to different nodes and zones.
+      topologySpreadConstraints:
+        - maxSkew: 2
+          topologyKey: topology.kubernetes.io/zone
+          whenUnsatisfiable: ScheduleAnyway
+          labelSelector:
+            matchLabels:
+              app: kafka-broker-dispatcher
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - podAffinityTerm:
+                labelSelector:
+                  matchLabels:
+                    app: kafka-broker-dispatcher
+                topologyKey: kubernetes.io/hostname
+              weight: 100
       serviceAccountName: knative-kafka-broker-data-plane
       securityContext:
         runAsNonRoot: true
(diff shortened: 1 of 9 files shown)


aavarghese (commented):
Thanks!!

One more file needs similar changes...[sourcev2](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/data-plane/config/sourcev2/500-dispatcher.yaml)

pierDipi:
Done

data-plane/config/broker/500-dispatcher.yaml
```diff
@@ -33,6 +33,23 @@ spec:
...
+              app: kafka-broker-dispatcher
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - podAffinityTerm:
+                labelSelector:
+                  matchLabels:
+                    app: kafka-broker-dispatcher
+                topologyKey: kubernetes.io/hostname
+              weight: 100
```
aavarghese, line 52: Do you know if this combination is better than having two pod topology spread constraints - one for zone and one for node? (Was trying to read about this...)

aavarghese: Since we don't have any specific anti-affinity rules for nodes, we could instead have multiple pod topology spread constraints?

```
     # To avoid node becoming SPOF, spread our replicas to different nodes and zones.
      topologySpreadConstraints:
        - maxSkew: 2
          topologyKey: topology.kubernetes.io/zone
          whenUnsatisfiable: ScheduleAnyway
          labelSelector:
            matchLabels:
              app: kafka-broker-dispatcher
        - maxSkew: 2
          topologyKey: kubernetes.io/hostname
          whenUnsatisfiable: ScheduleAnyway
          labelSelector:
            matchLabels:
              app: kafka-broker-dispatcher
```
This should also satisfy the existing anti-affinity rule for `preferredDuringSchedulingIgnoredDuringExecution`.

On the other hand, our existing dispatchers already had `podAntiAffinity` for nodes, so if we want to make minimal changes, that is understandable
cc: @pierDipi was this your thinking too?

pierDipi: Yes, afaik, `topologySpreadConstraints` + `maxSkew = 1` + `ScheduleAnyway` should be equivalent to our existing antiAffinity rule, I'm ok to migrate antiAffinity rules to use 
`topologySpreadConstraints` + `maxSkew = 1` + `ScheduleAnyway`

aavarghese: Ok, i think making this equivalent change separately another time/another PR is fine too.

aavarghese (approved):
/lgtm

PR #2094, opened 2022-04-13
Properly configure tracing in upgrade test toml configurations

mgencur:
* also remove the deprecated usage of .Namespace in the toml file and
use ForwarderTarget

If tracing is not properly configured in toml then the default (no-op) tracing is used.

**Release Note**

```release-note

```

**Docs**


pierDipi (approved):
/lgtm
/approve

pierDipi:
I think there is something wrong with this patch, perhaps the target change is the problem? :thinking:

PR #2095, opened 2022-04-13
Limit the number of poll calls for unordered delivery

pierDipi:
Multiple async `poll` calls can lead to a large number of records
loaded into memory and dispatched to subscribers.

In this PR, I'm limiting the number of in-flight poll requests to 1.


matzew (approved):
/lgtm
/approve

PR #2100, opened 2022-04-14
Clean up Zipkin tracing only once in upgrade tests

mgencur:
This is similar to https://github.com/knative/eventing/pull/6331 . When the Eventing changes propagate to this repo we'll need to ensure cleanup of Zipkin port-forwarding. This PR fixes that.

## Proposed Changes

- Clean up Zipkin port-forwarding only, when the upgrade test suite ends

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/upgrade/upgrade_test.go
+++ b/test/upgrade/upgrade_test.go
@@ -20,12 +20,18 @@
 package upgrade
 
 import (
+	"log"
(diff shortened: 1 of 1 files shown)


pierDipi (approved):
/lgtm

PR #2116, opened 2022-04-21
Add config-kafka-source-defaults ConfigMap for KEDA integration

pierDipi:
As per title.


aavarghese:
@pierDipi We have this file in https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/config/eventing-kafka-broker/100-source/config-kafka-source-defaults.yaml. Why do we need it here additionally?

pierDipi:
Good catch, closing

PR #2146, opened 2022-05-03
Add unit test for AsyncCloseable's compose utility

pierDipi:
As per title.


matzew (approved):
/lgtm
/approve

any specific case leading to adding this test?

pierDipi:
> /lgtm /approve
> 
> any specific case leading to adding this test?

I didn't trust the execution of a sequence of futures with "success", "failed", and "success" results, so I added a test to see the expected behavior.

PR #2153, opened 2022-05-04
Track secrets in consumer reconciler 

### LINKED ISSUE #2144: Kafka source should pick up missing secret
**Describe the bug**
When a Kafka source is created referencing a missing secret resource, when someone creates the secret the Kafka source references, Kafka source should be able to pick up the secret and becomes ready. 

 **Expected behavior**
Kafka source should be able to pick up the missing secret and becomes ready. 

 **To Reproduce**
1. Create a Kafka source referencing a secret that does not exist. 
2. Create the missing secret and observe that Kafka source does not pick up the secret and never gets ready. 

 **Knative release version**

 **Additional context**

pierDipi:
Fixes #2144 

## Proposed Changes

- Track secrets in consumer reconciler

### DIFF
--- a/control-plane/pkg/reconciler/consumer/consumer.go
+++ b/control-plane/pkg/reconciler/consumer/consumer.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/utils/pointer"
-	bindings "knative.dev/eventing-kafka/pkg/apis/bindings/v1beta1"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/reconciler"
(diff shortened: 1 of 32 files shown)


aavarghese (approved):
/lgtm

pierDipi:
@aavarghese there were conflict, I rebased now

control-plane/pkg/reconciler/consumer/consumer.go
```diff
@@ -204,15 +209,6 @@ func (r Reconciler) reconcileAuth(ctx context.Context, c *kafkainternals.Consume
 				Version:   secret.ResourceVersion,
 			},
 		}
-		ref := tracker.Reference{
-			APIVersion: c.APIVersion,
-			Kind:       c.Kind,
-			Namespace:  c.GetNamespace(),
-			Name:       c.GetName(),
```
aavarghese, line 211: hmm why was this Reference tracker even here before?

matzew (approved):
/lgtm
/approve

matzew (approved):
/lgtm
/approve

control-plane/pkg/reconciler/consumer/consumer.go
pierDipi: previously, we were only tracking the case where there is only one secret reference (all cases except the KafkaSource one), we're now tracking properly both cases.

PR #2164, opened 2022-05-09
Use host-based routing in KafkaChannel

pierDipi:
The previous implementation of the KafkaChannel uses host-based
routing.
To upgrade from the previous implementation without a big down-time
we need to use host-based routing in this implementation as well.

Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


pierDipi:
Blocked by https://github.com/knative-sandbox/eventing-kafka/pull/1172

control-plane/pkg/reconciler/channel/channel.go
```diff
@@ -62,7 +64,7 @@ const (
 	// TopicPrefix is the Kafka Channel topic prefix - (topic name: knative-messaging-kafka.<channel-namespace>.<channel-name>).
 	TopicPrefix                  = "knative-messaging-kafka"
 	DefaultDeliveryOrder         = contract.DeliveryOrder_ORDERED
-	NewChannelIngressServiceName = "kafka-channel-dispatcher"
```
matzew, line 65: +1

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/IngressProducerReconcilableStore.java
```diff
@@ -89,9 +89,27 @@ public IngressProducer resolve(String host, String path) {
...
+    // different host headers:
+    // - my-kafka-channel-kn-channel.default
+    // - my-kafka-channel-kn-channel.default:80
+    // - my-kafka-channel-kn-channel.default.svc
+    // - my-kafka-channel-kn-channel.default.svc:80
+    // - my-kafka-channel-kn-channel.default.svc.cluster.local [1]
+    // - my-kafka-channel-kn-channel.default.svc.cluster.local:80
+    //
+    // Current implementations expects only the long form that contains the cluster
+    // domain without the port [1], however, this is clearly a non-robust implementation,
```
matzew, line 109: PErhaps worth to create an issue for a different impl... but the names should, on the cluster, resolve to same, regardless if long form or short

pierDipi: right, that's what I mean with this comment, but for routing, a simple hash map doesn't work!

matzew (approved):
/lgtm
/approve

matzew:
I tested a channel, and got this `address`:

* `http://<channel-name>-kn-channel.default.svc.cluster.local`

and I could curl a CE against it (got 202)

matzew:
Is this the way to go forward? Using HOST based path?

I guess it has benefits (expect for the bizzare routing table @pierDipi mentioned on the java code), as we have a _service_ for it (e.g. `<channel-name>-kn-channel`)

I think that maybe... nice for broker too ? 

(e.g. I think the serving domain mapper makes things easier with a "proper" service, per obj)

pierDipi:
> Is this the way to go forward? Using HOST based path?
> 
> I guess it has benefits (expect for the bizzare routing table @pierDipi mentioned on the java code), as we have a _service_ for it (e.g. `<channel-name>-kn-channel`)
> 
> I think that maybe... nice for broker too ?
> 

Yes, it might be nice but for now we don't an easy migration paths for changing Broker addresses (and integrations with service meshes or ingresses are gonna break with these address changes)

> (e.g. I think the serving domain mapper makes things easier with a "proper" service, per obj)

matzew:
> Yes, it might be nice but for now we don't an easy migration path for changing Broker addresses (and integrations with service meshes or ingresses are gonna break with these address changes)

fair point

aliok:
```
prober.go:106: event #3304 had unavailable period 1m42.26562673s which is over down time limit of 1m0s
```

This is the error.

@matzew I remember you were increasing some timeout somewhere. Was it the same thing?

matzew:
@aliok see: https://github.com/knative/eventing/pull/6381

/cc @mgencur

aliok:
This PR blocked by https://github.com/knative/eventing/pull/6381

mgencur:
Why is it blocked by https://github.com/knative/eventing/pull/6381 ? It will only have a little better reporting in tests. Should not be blocking anything.

matzew:
@pierDipi could you rebase yours, and squeeze in this https://github.com/knative-sandbox/eventing-kafka-broker/pull/2281/commits/26a6593f2a911bada33ec94d463e8afc3d9f5393 ?

matzew:
@pierDipi I was allowing myself to do the rebase and the cherry-pick :cherries:  :sweat_smile:

matzew:
@pierDipi I hope to get https://github.com/knative-sandbox/eventing-kafka-broker/pull/2282  merged, than this

PR #2184, opened 2022-05-11
Add -XX:+CrashOnOutOfMemoryError flag

pierDipi:
Sometimes there are OOM errors when memory limits are set,
so by specifying this flag we get:
- a readable report of what happened
- the pod is restarted instead of leaving the JVM in a broken
  state


aliok (approved):
/lgtm
/approve

PR #2193, opened 2022-05-12
Use local instead of readonly

pierDipi:
Readonly variables inside functions are globally readonly, so
we cannot execute the function twice.

Fixes maybe nightly release errors.


aavarghese (approved):
Thanks @pierDipi - looks like the right fix for nightly.

/lgtm

PR #2215, opened 2022-05-23
Use already built image for Sacura tests

pierDipi:
sacura needed some patches to vegeta libraries in 
https://github.com/pierDipi/sacura/pull/16 and since 
sacura itself and eventing core need different versions
of vegeta we need to either remove the dependency
from eventing or from sacura (of course, the former isn't possible).
In this patch, I'm using the Sacura image directly.

This also removes:
- a lot of unnecessary dependencies
- the time to build the sacura image in CI

To be reviewed changes are in:
- test/config/sacura/300-job.yaml
- hack/tools.go


test/config/sacura/300-job.yaml
```diff
@@ -37,7 +37,7 @@ spec:
 
       containers:
         - name: sacura
-          image: ko://github.com/pierdipi/sacura/cmd/sacura
```
matzew, line 40: sound good to use an externally produced image

:+1:

matzew (approved):
/lgtm
/approve

PR #2221, opened 2022-05-25
Commit offsets on partition revoked

### LINKED ISSUE #2202: Unacceptable duplicate events from dispatcher to sink
**Describe the bug**
When running Kafka source and sink receiving events, we encountered unacceptable number of duplicate events. 

 **Expected behavior**
Although dups are unavoidable but the number of duplicate events is too many. See my results below:

Testing with no delay, 3000 events result 
source stevensource received: 3000 messages, 0 are dups and 0 non 200s.
Testing with no delay, 5000 events result 
source stevensource received: 5000 messages, 0 are dups and 0 non 200s.
**Testing with no delay, 20000 events result 
source stevensource received: 28000 messages, 8081 are dups and 0 non 200s.
Testing with 3 seconds delay with 2000 events result 
source stevensource received: 2000 messages, 1045 are dups and 0 non 200s.
Testing with 3 seconds delay with 10000 events result 
source stevensource received: 10000 messages, 8106 are dups and 0 non 200s.**

 **To Reproduce**
Create a Kafka source and make a sink that delays for 3-5 seconds, send large number of events to keep Kafka client consumer running for more than 15-20 mins and you should see dups in sink. 

 **Knative release version**

 **Additional context**

pierDipi:

Fixes #2202

## Proposed Changes

- Commit offsets on partition revoked

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/OffsetManager.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/OffsetManager.java
@@ -32,6 +32,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/OffsetManager.java
```diff
@@ -72,12 +73,29 @@ public OffsetManager(final Vertx vertx,
...
+      } finally {
+        // Remove revoked partitions in any case.
+        partitions.forEach(offsetTrackers::remove);
+      }
       logPartitions("revoked", partitions);
-      partitions.forEach(offsetTrackers::remove);
     });
 
     this.consumer.partitionsAssignedHandler(partitions ->
-      partitions.forEach(tp -> logPartitions("assigned", partitions))
```
aavarghese, line 80: @pierDipi wondering if on partition assignment, do we need to set this consumer back to the last committed offset if partition exists in offsettracker (i realize we've removed everything on revoke)....or the logic here is that when this consumer receives a new record, the offset tracker will get the new offset anyways?

pierDipi: Yes, I'm trying to find a good design for this because we want to consume this event in multiple components so we need a pub/sub here

pierDipi: On revoke, we remove only the offset trackers associated with the revoked partitions.

pierDipi: Once we will handle new partitions, we will re-introduce new offset trackers as we see records

pierDipi:
(I'll force push due to the cla/google and easyCLA problem)

PR #2244, opened 2022-06-01
Handle key-type in scheduler based source

pierDipi:
I saw this problem in https://github.com/knative-sandbox/eventing-kafka-broker/pull/2219

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>

Part of #1537 

## Proposed Changes

- Handle key-type in scheduler based source

PR #2247, opened 2022-06-02
Minor fix for missing allowed key type

aavarghese:

Related to #2244 

Fixes a problem in #2219

Part of #1537

### DIFF
--- a/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_types.go
+++ b/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_types.go
@@ -151,7 +151,7 @@ var (
 		"byte-array",
 		"string",
 		"int",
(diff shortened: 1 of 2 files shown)


matzew (approved):
/lgtm
/approve

matzew:
/hold

actually, can't we just use https://github.com/knative-sandbox/eventing-kafka/blob/main/pkg/apis/sources/v1beta1/kafka_types.go#L103

aavarghese:
> actually, can't we just use https://github.com/knative-sandbox/eventing-kafka/blob/main/pkg/apis/sources/v1beta1/kafka_types.go#L103

@matzew yeah, we should. Made that change in a new commit. Thanks!

pierDipi (approved):
Thank you!

/lgtm
/approve

PR #2299, opened 2022-06-13
New sticky partition strategy for Kafka rebalancing (works for all kafka clients versions)

### LINKED ISSUE #2202: Unacceptable duplicate events from dispatcher to sink

aavarghese:
Fixes #2202 

See https://github.com/knative-sandbox/eventing-kafka-broker/issues/2202#issuecomment-1142333099

## Proposed Changes

- Changes partition strategy for rebalancing from default (`RangeAssignor`) to `StickyAssignor` which causes less number of duplicate messages to be received in sink due to less disrupting partition assignments to consumers. Customers on newer clusters with Kafka v2.4+ can use a sticky + incremental protocol `CooperativeStickyAssignor` instead.

**Release Note**

```release-note
Changed partition strategy for rebalancing from default (`RangeAssignor`) to `StickyAssignor` . 
```

**Docs**

`StickyAssignor` is supported on Kafka client versions [0.11.0](https://javadoc.io/static/org.apache.kafka/kafka-clients/0.11.0.0/org/apache/kafka/clients/consumer/StickyAssignor.html) ([support ended in 2019](https://docs.confluent.io/platform/current/installation/versions-interoperability.html)) to current [3.2.x](https://javadoc.io/static/org.apache.kafka/kafka-clients/2.8.0/org/apache/kafka/clients/consumer/StickyAssignor.html). So this setting should be compatible with older Kafka clusters and newer ones. And provides better rebalancing than `RangeAssignor` (Default)
Users with newer clusters must use newer less disruptive protocols like [`CooperativeStickyAssignor`](https://javadoc.io/static/org.apache.kafka/kafka-clients/2.8.0/org/apache/kafka/clients/consumer/CooperativeStickyAssignor.html)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/OrderedConsumerVerticle.java
```diff
@@ -176,6 +177,9 @@ private void recordsHandler(KafkaConsumerRecords<Object, CloudEvent> records) {
       bucket.forceAddTokens(records.size());
     }
 
+    Timestamp timestamp1 = new Timestamp(System.currentTimeMillis());
+    System.out.println("Time is: " + timestamp1.toString() + ". received records count: " + records.size());
```
pierDipi, line 181: FYI, tf we create a [ConsumerRecordContext](https://github.com/knative-sandbox/eventing-kafka-broker/blob/935d918bdf0ecc693fa24c6a010f4fd7ef9a2b9f/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java#L153) here and pass it to the RecordDispatcher we get this latency included in the processing latency metric, that would definitely make it easier to monitor across multiple replicas.

aavarghese: Yes - that's great. Thanks I'll add this in a separate PR in conjunction with new metric changes and remove it from here...

aavarghese:
My `duplicates` scenario tests do show decrease in duplicates with this new assignor. @aslom and IBM Code Engine tests in CE shards @cr22rc can confirm

ok to merge? @pierDipi

data-plane/config/broker/100-config-kafka-broker-data-plane.yaml
```diff
@@ -133,7 +133,7 @@ data:
     isolation.level=read_uncommitted
     max.poll.interval.ms=300000
     max.poll.records=500
-    # partition.assignment.strategy=org.apache.kafka.clients.consumer.CooperativeStickyAssignor
+    partition.assignment.strategy=org.apache.kafka.clients.consumer.CooperativeStickyAssignor
```
pierDipi, line 136: The problem with this is that it's supported only on Kafka 2.4+, so shipping this config by default isn't optimal since there are lots of people on old Kafka clusters.

I think we can get the same benefits by using the StickyAssignor with a smarter logic that records the last assignment as suggested in the docs here https://kafka.apache.org/26/javadoc/org/apache/kafka/clients/consumer/StickyAssignor.html:

```
 class TheNewRebalanceListener implements ConsumerRebalanceListener {
   Collection<TopicPartition> lastAssignment = Collections.emptyList();

   void onPartitionsRevoked(Collection<TopicPartition> partitions) {
     for (TopicPartition partition: partitions)
       commitOffsets(partition);
   }

   void onPartitionsAssigned(Collection<TopicPartition> assignment) {
     for (TopicPartition partition: difference(lastAssignment, assignment))
       cleanupState(partition);

     for (TopicPartition partition: difference(assignment, lastAssignment))
       initializeState(partition);

     for (TopicPartition partition: assignment)
       initializeOffset(partition);

     this.lastAssignment = assignment;
   }
 }
```

what do you think?

matzew: > The problem with this is that it's supported only on Kafka 2.4+,

what version of Kafka do we generally support? (as in a range)

aavarghese: @pierDipi @matzew  I think we'll need to experiment with this logic a bit to see if it gives the same low number of duplicate results as `CooperativeStickyAssignor` especially since this sticky+smart logic may be still be a semi-eager protocol. What do you think about changing the current `Range` default to `Sticky` anyways for release `1.6.0`? And then customizing it a bit more after we've done some testing with results..

IBM Event Streams is on Kafka v2.8 so they should probably use Cooperative...

data-plane/dispatcher/src/test/java/dev/knative/eventing/kafka/broker/dispatcher/integration/UnorderedConsumerTest.java
```diff
@@ -75,6 +76,7 @@ public void testUnorderedConsumer(final Vertx vertx) throws Exception {
     final var producerConfigs = new Properties();
     final var consumerConfigs = new Properties();
     consumerConfigs.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, 100L);
+    consumerConfigs.put(ConsumerConfig.PARTITION_ASSIGNMENT_STRATEGY_CONFIG, StickyAssignor.class.getName());
```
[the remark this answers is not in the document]
pierDipi: Ah, nvm, it's a test

aavarghese:
@pierDipi @matzew why do these two specific CLI tests fail most of the time (not just in this PR) but in some cases pass?

```
[ERROR] Errors: 
[ERROR]   RecordDispatcherTest.shouldCallFailedToSendToDeadLetterSinkIfValueMatchesAndSubscriberSenderFailsAndNoDeadLetterSinkSender:274->assertEventProcessingLatency:418 ? ConditionTimeout
[ERROR]   RecordDispatcherTest.shouldSendToDeadLetterSinkIfValueMatchesAndSubscriberSenderFails:193->assertEventProcessingLatency:418 ? ConditionTimeout
```

pierDipi (approved):
/lgtm
/approve

PR #2304, opened 2022-06-14
Detect subscribers changes in KafkaChannel reconciler

pierDipi:
When a subscriber is added or removed, the reconciler won't detect
this change and it won't update the contract generation, leading
to the data plane to skip the reconciliation since a contract
with the same generation is considered the same contract.


aliok (approved):
/lgtm
/approve

PR #2305, opened 2022-06-14
Reduce default consumer and producer timeouts

pierDipi:
Big timeouts as before end up parking threads unnecessarily, this is
especially evident when deleting a resource end up having consumers
polling topic/partitions that don't exist anymore until timeouts
are reached.


aliok (approved):
/lgtm
/approve

PR #2333, opened 2022-06-15
Fallback mechanism for the old external topic annotation for broker

aliok:
Fixes #

## Proposed Changes

- Fallback mechanism for the old external topic annotation for broker
-
-

**Release Note**

```release-note
As some users of the KafkaBroker use `x-kafka.eventing.knative.dev/external.topic` annotation for specifying an external topic instead of the new annotation (`kafka.eventing.knative.dev/external.topic`), a fallback mechanism is added to support it. New annotation has precedence over the old one.
```

**Docs**


aliok:
/close

We only need this at downstream at Red Hat

PR #2342, opened 2022-06-20
[release-1.4] Adding annotation to drive the name of the backing topic

matzew:

Manual backport of #1023 to 1.4 branch


control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -319,27 +338,33 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker)
 	}
 
 	topicConfig, brokerConfig, err := r.topicConfig(logger, broker)
```
matzew, line 340: due to the `r.removeFinalizerCM(...)` call inside the `finalizeKind`, I keep the `topicConfig()` call outside of the "is not external topic" block...

matzew:
/close 

we have this on 1.5, which is good enough

PR #2356, opened 2022-06-24
[Prefactoring] Rename `reconciler.SystemNamespace` to `reconciler.DataPlaneNamespace`

aliok:
Fixes #

## Proposed Changes

- Better semantics, as we're not going to talk about system namespace anymore but talk about the namespace that the reconciler instance is going to take care of
- Prefactoring for KafkaBroker with dataplane-per-namespace

**Release Note**

```release-note

```

**Docs**


pierDipi:
LGTM, there is a build error

pierDipi (approved):
/lgtm
/approve

PR #2360, opened 2022-06-25
Do not use `system.Namespace()` in reconciler unit tests

aliok:
Fixes #

## Proposed Changes

- Do not use `system.Namespace()` in reconciler unit tests
- Unit tests on GH actions failed because of that
- Introduced by https://github.com/knative-sandbox/eventing-kafka-broker/pull/2356

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/channel/v2/channelv2_test.go
+++ b/control-plane/pkg/reconciler/channel/v2/channelv2_test.go
@@ -29,14 +29,12 @@ import (
 	"k8s.io/utils/pointer"
 	eventingduck "knative.dev/eventing/pkg/apis/duck/v1"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
+	kubeclient "knative.dev/pkg/client/injection/kube/client/fake"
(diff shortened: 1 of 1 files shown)


control-plane/pkg/reconciler/channel/v2/channelv2_test.go
```diff
@@ -110,7 +108,7 @@ func TestReconcileKind(t *testing.T) {
 				NewChannel(
 					WithInitKafkaChannelConditions,
 					WithDeletedTimeStamp),
-				NewConfigMapWithTextData(system.Namespace(), DefaultEnv.GeneralConfigMapName, map[string]string{
+				NewConfigMapWithTextData(env.SystemNamespace, DefaultEnv.GeneralConfigMapName, map[string]string{
```
matzew, line 111: was that only on GH action failing?

aliok: yeah

PR #2361, opened 2022-06-27
Namespaced broker implementation - initial implementation with disabled controllers

aliok:
Fixes #

## Proposed Changes

- Namespaced broker implementation

Notes:
- We wrap the existing reconcilers to make them work in user namespaces
- Per-namespace dataplane manifest is stored in `control-plane/cmd/kafka-controller/kodata/dataplane/config/broker`. I wanted to reuse the existing manifest in `data-plane/config/broker` but there were some differences
  - The `ClusterRoleBinding` for the ClusterRole `knative-kafka-broker-data-plane` cannot be used as we don't want the dataplane in user namespace to have permissions across cluster. I created a RoleBinding for the same ClusterRole so that same permissions are given within the user namespace.
  - `config-tracing` and `config-logging` is in `knative-eventing` namespace and cannot mounted. Removed these mounts,
  - We don't want to recreate the `ClusterRole`.
- New RBAC permissions are given to the controller so that it can manage resources in user namespace
- When creating the dataplane in user namespace by the controller, we need to set the image of the receiver and dispatcher deployments. To accomplish that,
  - Controller reads env vars called `BROKER_RECEIVER_IMAGE` and `BROKER_DISPATCHER_IMAGE`.
  - It changes the  `${KNATIVE_KAFKA_RECEIVER_IMAGE}` and `${KNATIVE_KAFKA_DISPATCHER_IMAGE}` placeholders in the manifest with those.
  - `BROKER_RECEIVER_IMAGE` and `BROKER_DISPATCHER_IMAGE` env vars are defined in the controller deployment YAML. There are 2 placeholders for them: `${KNATIVE_KAFKA_RECEIVER_IMAGE}` and `${KNATIVE_KAFKA_DISPATCHER_IMAGE}`. These are filled by scripts when you do `hack/run.sh deploy`; just like the placeholders in receiver and dispatcher deployment YAMLs are filled when you do `hack/run.sh deploy`.
  - So, the point I am trying to make is: scripting don't touch the YAML in `control-plane/cmd/kafka-controller/kodata/dataplane/config/broker`. If we go that route, we would need to make sure the dataplane images are built and pushed before the control plane is built and that we need to operate on in-memory/temporary YAML files instead of the files in the source tree.
- Dataplane manifest applying is done via using Manifestival.
- We inject ownerRefs to the things created in user namespace. For multiple brokers in the same namespace, we set multiple ownerRefs. Thanks to the Kubernetes GC, when there are no more ownerRefs remaining in the resources, dataplane is removed automatically.
- To use the same prober for multiple broker instances, a new IPLister called `IPListerWithMapping` is created. This struct holds a mapping of broker-to-ingressService and prober asks the ingressService for a probing request for every broker.
- Watches for the contract configmap in usernamespace are done by watching configmaps with label `eventing.knative.dev/namespaced=true`. Broker reconciler needs it, for the reconciliation of the configmap is deleted or something. Trigger reconciler needs this watch as well, to see the changes in the contract and adjust things.

Issues:
- E2E tests are missing
- Some solution to be able to read `knative-eventing/config-tracing` and `knative-eventing/config-logging`
- `config-kafka-broker-data-plane` dataplane configmap is created by the reconciler. This results in 2 issues: a) changes might be reverted b) users cannot specify the dataplane config in advance. One solution would be adding a watch but not reverting changes AND not adding any ownerRef when the configmap is already created by the user (to prevent GC)
- There are duplicate YAML content for the global dataplane and the namespaced dataplane. One solution would be using Kustomize to statically build these during the build time
- Add special label for namespaced dataplane resources for future migrations and things like that
- (optional) Read dataplane configmap in user namespace, then fallback to system namespace
- (optional) `IPListerWithMapping` struct can leak resources in case of a leader change. We might need to create a TTL mechanism

Testing instructions for broker reconciliation:
- Create a broker instance:
```yaml
kind: Broker
metadata:
  annotations:
    # case-sensitive
    eventing.knative.dev/broker.class: KafkaNamespaced
    hello: world
  name: default
  namespace: default
spec:
  # Configuration specific to this broker.
  config:
    apiVersion: v1
    kind: ConfigMap
    name: kafka-broker-config
    namespace: knative-eventing
```
- See dataplane created in `default` namespace
- Change the `hello:world` annotation to see reconciliation
- Create another broker in the same namespace and see that the same dataplane is used
- Delete one of the brokers, dataplane is still there
- Delete the other broker, dataplane will be deleted
- TODO: Trigger testing instructions

**Release Note**

```release-note

```

**Docs**


control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -274,15 +276,15 @@ func (r *Reconciler) reconcileBrokerTopic(broker *eventing.Broker, securityOptio
 
 func (r *Reconciler) FinalizeKind(ctx context.Context, broker *eventing.Broker) reconciler.Event {
 	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		return r.finalizeKind(ctx, broker)
+		return r.DoFinalizeKind(ctx, broker)
 	})
 }
 
-func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker) reconciler.Event {
+func (r *Reconciler) DoFinalizeKind(ctx context.Context, broker *eventing.Broker) reconciler.Event {
```
pierDipi, line 283: I don't think the `RetryOnConfict` in the other reconciler is needed, we should just call `ReconciledKind` of this reconciler.

control-plane/pkg/reconciler/broker/namespaced_broker.go
```diff
@@ -0,0 +1,278 @@
...
+
+	Prober prober.Prober
+
+	IPsLister prober.IPListerWithMapping
+}
+
+func (r *NamespacedReconciler) ReconcileKind(ctx context.Context, broker *eventing.Broker) reconciler.Event {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		return r.reconcileKind(ctx, broker)
+	})
```
pierDipi, line 91: This shouldn't be needed if we call `ReconcileKind` of the other reconciler.

control-plane/cmd/kafka-controller/kodata/dataplane/config/broker/100-config-kafka-broker-data-plane.yaml
```diff
@@ -0,0 +1,179 @@
+# Copyright 2020 The Knative Authors
```
pierDipi, line 1: Receiver, dispatcher and this CM are exactly the same as the other in `data-plane/config/...` (the CM is actually already out of date ...), can we just add to `./hack/update-codegen.sh` `cp <dispatcher_path> <ko_data_dispatcher_path>` so that we don't need to keep them in sync manually?

Other thought, since we need to support customizations, I believe creating manifests on the fly based on the ones in the system namespace would provide the customizations we need and it would also avoid injecting the images into the controller deployment.

Example:
- get receiver, dispatcher and CM from `system.Namespace()`
- get SA and role from the kodata filesystem
- create manifests with all the transformations that we're doing

In this way, if I update, for example, the receiver with an addition env variable, I get that propagated to all other receivers in my cluster.

Example patch for CM:
```
diff --git a/control-plane/pkg/reconciler/broker/namespaced_broker.go b/control-plane/pkg/reconciler/broker/namespaced_broker.go
index 7fa566f1..32bf0354 100644
--- a/control-plane/pkg/reconciler/broker/namespaced_broker.go
+++ b/control-plane/pkg/reconciler/broker/namespaced_broker.go
@@ -27,18 +27,20 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/pointer"
+	eventing "knative.dev/eventing/pkg/apis/eventing/v1"
+	"knative.dev/pkg/reconciler"
+	"knative.dev/pkg/resolver"
+
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/config"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/kafka"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/prober"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/base"
-	eventing "knative.dev/eventing/pkg/apis/eventing/v1"
-	"knative.dev/pkg/reconciler"
-	"knative.dev/pkg/resolver"
 )
 
 // TODO: 1.
@@ -164,6 +166,11 @@ func (r *NamespacedReconciler) createReconcilerForBrokerInstance(broker *eventin
 func (r *NamespacedReconciler) getManifest(broker *eventing.Broker) (mf.Manifest, error) {
 	manifest := r.BaseDataPlaneManifest
 
+	cmManifest := r.dataPlaneConfigMapManifest()
+	// TODO do the same for receiver and dispatcher deployments/statefulsets
+
+	manifest.Append(cmManifest)
+
 	manifest, err := manifest.Transform(mf.InjectNamespace(broker.Namespace))
 	if err != nil {
 		return mf.Manifest{}, fmt.Errorf("unable to transform base dataplane manifest with namespace injection. namespace: %s, error: %v", broker.Namespace, err)
@@ -172,6 +179,15 @@ func (r *NamespacedReconciler) getManifest(broker *eventing.Broker) (mf.Manifest
 	return manifest.Transform(appendNewOwnerRefsToPersisted(manifest.Client, broker))
 }
 
+// TODO handle errors
+func (r *NamespacedReconciler) dataPlaneConfigMapManifest() mf.Manifest {
+	cm, _ := r.ConfigMapLister.ConfigMaps(r.Env.DataPlaneConfigMapNamespace).Get(r.Env.DataPlaneConfigMapName)
+	cmUnstr, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(cm)
+	u := unstructured.Unstructured{Object: cmUnstr}
+	cmManifest, _ := mf.ManifestFrom(mf.Slice{u})
+	return cmManifest
+}
+
 func appendNewOwnerRefsToPersisted(client mf.Client, broker *eventing.Broker) mf.Transformer {
 	return func(resource *unstructured.Unstructured) error {
 		existingRefs, err := getPersistedOwnerRefs(client, resource)
```

pierDipi: The same approach can be used for config-tracing, logging, etc (that we're removing)

control-plane/pkg/reconciler/trigger/namespaced_controller.go
```diff
@@ -0,0 +1,132 @@
...
+		func(name string, value interface{}) {
+			flags, ok := value.(feature.Flags)
+			if !ok {
+				logger.Warn("Features ConfigMap " + name + " updated but we didn't get expected flags. Skipping updating cached features")
+			}
+			logger.Debug("Features ConfigMap " + name + " updated. Updating cached features.")
+			reconciler.FlagsLock.Lock()
+			defer reconciler.FlagsLock.Unlock()
+			reconciler.Flags = flags
+			impl.GlobalResync(triggerInformer.Informer())
```
pierDipi, line 97: Can we extract common functions with common logic?

control-plane/pkg/reconciler/trigger/namespaced_controller.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+	)
+	featureStore.WatchConfigs(watcher)
+
+	reconciler.Resolver = resolver.NewURIResolverFromTracker(ctx, impl.Tracker)
+
+	triggerInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
```
pierDipi, line 104: we need additional handlers for all additional resources we're creating

pierDipi: wrong place for this comment, it is meant to go for the namespace_controller.go in broker package

control-plane/pkg/reconciler/trigger/trigger.go
```diff
@@ -117,7 +119,7 @@ func (r *Reconciler) reconcileKind(ctx context.Context, trigger *eventing.Trigge
 	}
 
 	// Get data plane config map.
-	contractConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx)
+	contractConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx, r.DataPlaneConfigMapLabeler)
```
pierDipi, line 122: This additional parameter should not be necessary.

control-plane/pkg/reconciler/trigger/trigger.go
```diff
@@ -203,7 +205,7 @@ func (r *Reconciler) finalizeKind(ctx context.Context, trigger *eventing.Trigger
 	}
 
 	// Get data plane config map.
-	dataPlaneConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx)
+	dataPlaneConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx, r.DataPlaneConfigMapLabeler)
```
pierDipi, line 208: ditto.

control-plane/pkg/reconciler/base/reconciler.go
```diff
@@ -108,25 +108,27 @@ func isAtLeastOneRunning(pods []*corev1.Pod) bool {
...
+
 func (r *Reconciler) GetOrCreateDataPlaneConfigMap(ctx context.Context, options ...ConfigMapOption) (*corev1.ConfigMap, error) {
 
 	cm, err := r.KubeClient.CoreV1().
 		ConfigMaps(r.DataPlaneConfigMapNamespace).
 		Get(ctx, r.DataPlaneConfigMapName, metav1.GetOptions{})
 
 	if apierrors.IsNotFound(err) {
-		cm, err = r.createDataPlaneConfigMap(ctx, options)
+		cm, err = CreateDataPlaneConfigMap(ctx, r.KubeClient, r.DataPlaneConfigMapNamespace, r.DataPlaneConfigMapName, options...)
```
pierDipi, line 120: I think we should transform existing ConfigMaps with the options for consistency.

control-plane/cmd/kafka-controller/kodata/dataplane/config/broker/201-broker-data-plane-role-binding.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: knative-kafka-broker-data-plane
+  labels:
+    kafka.eventing.knative.dev/release: devel
+subjects:
+  - kind: ServiceAccount
+    name: knative-kafka-broker-data-plane
+roleRef:
+  kind: ClusterRole
+  name: knative-kafka-broker-data-plane
+  apiGroup: rbac.authorization.k8s.io
```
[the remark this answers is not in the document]
aliok: Yes it does.

control-plane/pkg/reconciler/broker/broker.go
aliok: LOL good point :)

control-plane/cmd/kafka-controller/kodata/dataplane/config/broker/100-config-kafka-broker-data-plane.yaml
aliok: IMO let's discuss this separately.

control-plane/pkg/reconciler/trigger/trigger.go
aliok: Did not get what you mean...

The `GetOrCreateDataPlaneConfigMap` is defined in the `base.Reconciler` and is used by various reconciler implementations. In some we pass something, in another we pass nothing.

Can you elaborate?

control-plane/pkg/reconciler/broker/broker.go
aliok: Done

control-plane/pkg/reconciler/trigger/trigger.go
pierDipi: This is the idea to remove duplicated code that is in `reconcileDataPlaneConfigMap` (this is bad also because is based on how current the existing reconciler works internally) and also to have `GetOrCreateDataPlaneConfigMap` to have a reconciler semantic:

```
pierdipi@pierdipi eventing-kafka-broker (2022-06-27-namespaced-broker-impl) $ cat remove_duplicate_code.patch 
diff --git a/control-plane/pkg/reconciler/base/reconciler.go b/control-plane/pkg/reconciler/base/reconciler.go
index 1565f6dd..62b05fbd 100644
--- a/control-plane/pkg/reconciler/base/reconciler.go
+++ b/control-plane/pkg/reconciler/base/reconciler.go
@@ -85,6 +85,8 @@ type Reconciler struct {
 
        DispatcherLabel string
        ReceiverLabel   string
+
+       DataPlaneConfigMapTransformer ConfigMapOption
 }
 
 func (r *Reconciler) IsReceiverRunning() bool {
@@ -110,36 +112,36 @@ type ConfigMapOption func(cm *corev1.ConfigMap)
 
 func NoopConfigmapOption(cm *corev1.ConfigMap) {}
 
-func (r *Reconciler) GetOrCreateDataPlaneConfigMap(ctx context.Context, options ...ConfigMapOption) (*corev1.ConfigMap, error) {
+func (r *Reconciler) GetOrCreateDataPlaneConfigMap(ctx context.Context) (*corev1.ConfigMap, error) {
 
        cm, err := r.KubeClient.CoreV1().
                ConfigMaps(r.DataPlaneConfigMapNamespace).
                Get(ctx, r.DataPlaneConfigMapName, metav1.GetOptions{})
 
        if apierrors.IsNotFound(err) {
-               cm, err = CreateDataPlaneConfigMap(ctx, r.KubeClient, r.DataPlaneConfigMapNamespace, r.DataPlaneConfigMapName, options...)
+               cm, err = r.createDataPlaneConfigMap(ctx)
+       }
+
+       if r.DataPlaneConfigMapTransformer != nil {
+               r.DataPlaneConfigMapTransformer(cm)
        }
 
        return cm, err
 }
 
-func CreateDataPlaneConfigMap(ctx context.Context, kubeClient kubernetes.Interface, namespace string, name string, options ...ConfigMapOption) (*corev1.ConfigMap, error) {
+func (r *Reconciler) createDataPlaneConfigMap(ctx context.Context) (*corev1.ConfigMap, error) {
        cm := &corev1.ConfigMap{
                TypeMeta: metav1.TypeMeta{},
                ObjectMeta: metav1.ObjectMeta{
-                       Name:      name,
-                       Namespace: namespace,
+                       Name:      r.DataPlaneConfigMapName,
+                       Namespace: r.DataPlaneConfigMapNamespace,
                },
                BinaryData: map[string][]byte{
                        ConfigMapDataKey: []byte(""),
                },
        }
 
-       for _, opt := range options {
-               opt(cm)
-       }
-
-       return kubeClient.CoreV1().ConfigMaps(namespace).Create(ctx, cm, metav1.CreateOptions{})
+       return r.KubeClient.CoreV1().ConfigMaps(r.DataPlaneConfigMapNamespace).Create(ctx, cm, metav1.CreateOptions{})
 }
 
 // GetDataPlaneConfigMapData extracts contract from the given config map.
diff --git a/control-plane/pkg/reconciler/broker/broker.go b/control-plane/pkg/reconciler/broker/broker.go
index 230ae987..922c9f84 100644
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -69,8 +69,6 @@ type Reconciler struct {
        BootstrapServers string
 
        Prober prober.Prober
-
-       DataPlaneConfigMapLabeler base.ConfigMapOption
 }
 
 func (r *Reconciler) ReconcileKind(ctx context.Context, broker *eventing.Broker) reconciler.Event {
@@ -136,7 +134,7 @@ func (r *Reconciler) DoReconcileKind(ctx context.Context, broker *eventing.Broke
        }
 
        // Get contract config map.
-       contractConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx, r.DataPlaneConfigMapLabeler)
+       contractConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx)
        if err != nil {
                return statusConditionManager.FailedToGetConfigMap(err)
        }
@@ -284,7 +282,7 @@ func (r *Reconciler) DoFinalizeKind(ctx context.Context, broker *eventing.Broker
        logger := kafkalogging.CreateFinalizeMethodLogger(ctx, broker)
 
        // Get contract config map.
-       contractConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx, r.DataPlaneConfigMapLabeler)
+       contractConfigMap, err := r.GetOrCreateDataPlaneConfigMap(ctx)
        if err != nil {
                return fmt.Errorf("failed to get contract config map %s: %w", r.DataPlaneConfigMapAsString(), err)
        }
diff --git a/control-plane/pkg/reconciler/broker/namespaced_broker.go b/control-plane/pkg/reconciler/broker/namespaced_broker.go
index 1965ca00..5729afe1 100644
--- a/control-plane/pkg/reconciler/broker/namespaced_broker.go
+++ b/control-plane/pkg/reconciler/broker/namespaced_broker.go
@@ -24,20 +24,20 @@ import (
        appsv1 "k8s.io/api/apps/v1"
        corev1 "k8s.io/api/core/v1"
        "k8s.io/apimachinery/pkg/api/errors"
-       apierrors "k8s.io/apimachinery/pkg/api/errors"
        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
        "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
        "k8s.io/apimachinery/pkg/types"
        "k8s.io/client-go/kubernetes/scheme"
        corelisters "k8s.io/client-go/listers/core/v1"
        "k8s.io/utils/pointer"
+       eventing "knative.dev/eventing/pkg/apis/eventing/v1"
+       "knative.dev/pkg/reconciler"
+       "knative.dev/pkg/resolver"
+
        "knative.dev/eventing-kafka-broker/control-plane/pkg/config"
        "knative.dev/eventing-kafka-broker/control-plane/pkg/kafka"
        "knative.dev/eventing-kafka-broker/control-plane/pkg/prober"
        "knative.dev/eventing-kafka-broker/control-plane/pkg/reconciler/base"
-       eventing "knative.dev/eventing/pkg/apis/eventing/v1"
-       "knative.dev/pkg/reconciler"
-       "knative.dev/pkg/resolver"
 )
 
 // TODO: 1.
@@ -92,11 +92,6 @@ func (r *NamespacedReconciler) ReconcileKind(ctx context.Context, broker *eventi
 
        br := r.createReconcilerForBrokerInstance(broker)
 
-       err := r.reconcileDataPlaneConfigMap(ctx, broker)
-       if err != nil {
-               return fmt.Errorf("failed to create/update data plane config map: %s, error: %v", fmt.Sprintf("%s/%s", broker.Namespace, r.Env.DataPlaneConfigMapName), err)
-       }
-
        manifest, err := r.getManifest(broker)
        if err != nil {
                return fmt.Errorf("unable to transform dataplane manifest. namespace: %s, broker: %v, error: %v", broker.Namespace, broker, err)
@@ -134,6 +129,11 @@ func (r *NamespacedReconciler) createReconcilerForBrokerInstance(broker *eventin
 
                        DataPlaneNamespace:          broker.Namespace,
                        DataPlaneConfigMapNamespace: broker.Namespace,
+
+                       DataPlaneConfigMapTransformer: func(cm *corev1.ConfigMap) {
+                               kafka.NamespacedDataplaneLabelConfigmapOption(cm)
+                               appendBrokerAsOwnerRef(broker)
+                       },
                },
                Env:                        r.Env,
                Resolver:                   r.Resolver,
@@ -141,8 +141,6 @@ func (r *NamespacedReconciler) createReconcilerForBrokerInstance(broker *eventin
                NewKafkaClusterAdminClient: r.NewKafkaClusterAdminClient,
                BootstrapServers:           r.BootstrapServers,
                Prober:                     r.Prober,
-
-               DataPlaneConfigMapLabeler: kafka.NamespacedDataplaneLabelConfigmapOption,
        }
 }
 
@@ -236,30 +234,3 @@ func setImagesForDeployments(imageMap map[string]string) mf.Transformer {
                return scheme.Scheme.Convert(deployment, resource, nil)
        }
 }
-
-func (r *NamespacedReconciler) reconcileDataPlaneConfigMap(ctx context.Context, broker *eventing.Broker) error {
-       cm, err := r.KubeClient.CoreV1().
-               ConfigMaps(broker.Namespace).
-               Get(ctx, r.Env.DataPlaneConfigMapName, metav1.GetOptions{})
-
-       if err != nil && !apierrors.IsNotFound(err) {
-               return err
-       }
-
-       if apierrors.IsNotFound(err) {
-               // not found, create
-               _, err = base.CreateDataPlaneConfigMap(ctx, r.KubeClient, broker.Namespace, r.Reconciler.DataPlaneConfigMapName, appendBrokerAsOwnerRef(broker), kafka.NamespacedDataplaneLabelConfigmapOption)
-               return err
-       }
-
-       // found, update
-
-       if refs, appended := appendOwnerRef(cm.GetOwnerReferences(), broker); appended {
-               cm.SetOwnerReferences(refs)
-               // only update if we actually appended an ownerRef
-               _, err = r.KubeClient.CoreV1().ConfigMaps(broker.Namespace).Update(ctx, cm, metav1.UpdateOptions{})
-               return err
-       }
-
-       return nil
-}
```

pierDipi: with this we also avoid unnecessary API server calls in `reconcileDataPlaneConfigMap`

control-plane/pkg/reconciler/base/reconciler.go
pierDipi: I think this captured the general idea: https://github.com/knative-sandbox/eventing-kafka-broker/pull/2361#discussion_r928702545

control-plane/cmd/kafka-controller/kodata/dataplane/config/broker/100-config-kafka-broker-data-plane.yaml
pierDipi: Ok

pierDipi: but we need to update manifests ...

control-plane/pkg/reconciler/broker/broker.go
pierDipi: I think this is still a problem because we're still calling `Do*`

control-plane/pkg/reconciler/trigger/trigger.go
aliok: > This additional parameter should not be necessary.

Well, it was impossible for me to get all of these from 1 generic sentence.

I wrote that `reconcileDataPlaneConfigMap` because the existing `GetOrCreateDataPlaneConfigMap` didn't work like I wanted. Specifically, it was checking for existence and if CM was not there it was creating it. However, it is not updating the CM when it already existed.

I guess this approach would work. Except, it won't update the CM immediately and update it later with a label and owner. I need to think more about this.

control-plane/pkg/reconciler/broker/broker.go
aliok: Oh, nice catch. I messed up

control-plane/cmd/kafka-controller/kodata/dataplane/config/broker/100-config-kafka-broker-data-plane.yaml
aliok: I am gonna merge this in a disabled way. Things work now w/o configmaps mounted.
Manifests can be updated later when we're ready with the whole thing.

Or, do you mean something else?

control-plane/pkg/reconciler/trigger/trigger.go
aliok: > I guess this approach would work. Except, it won't update the CM immediately and update it later with a label and owner. I need to think more about this.

No, there are some cases we need to think about...

It is safer to apply the label and the owner immediately. Otherwise, if there's something wrong until the rest of the reconciliation, we could lose the label and the ownerRef information until the update is happening.

I am talking about the case that the CM already exists. In that case, the label would be there anyway and that's good, but we might lose the ownerRef in case of a reconciliation problem.

What do you think?

pierDipi: We can still call update here [1] but it's not necessary, however, at least the logic is encapsulated in one single place without duplication

[1]

```
+       if r.DataPlaneConfigMapTransformer != nil {
+               r.DataPlaneConfigMapTransformer(cm)
+               // update
        }
```

control-plane/cmd/kafka-controller/kodata/dataplane/config/broker/100-config-kafka-broker-data-plane.yaml
pierDipi: > I am gonna merge this in a disabled way. Things work now w/o configmaps mounted.
Manifests can be updated later when we're ready with the whole thing.

This CM config-kafka-broker-data-plane is still mounted in the new data plane and it's different from the one in `data-plane/config/...`, feel free to ignore this comment for now but the current situation is not optimal, given the CM config-kafka-broker-data-plane is already out of date ...

control-plane/pkg/reconciler/trigger/trigger.go
pierDipi: > I am talking about the case that the CM already exists. In that case, the label would be there anyway and that's good, but we might lose the ownerRef in case of a reconciliation problem.

There shouldn't be a problem with this since, in the worst case, the ConfigMap will be deleted and then recreated when the Broker is reconciled again.

aliok:
Addressed comments, disabled the controllers.

cc @pierDipi

control-plane/pkg/reconciler/broker/namespaced_broker.go
```diff
@@ -0,0 +1,236 @@
...
+		resource.SetOwnerReferences(refs)
+		return nil
+	}
+}
+
+func getPersistedOwnerRefs(mfc mf.Client, res *unstructured.Unstructured) ([]metav1.OwnerReference, error) {
+	retrieved, err := mfc.Get(res)
+	if err != nil && !errors.IsNotFound(err) {
+		// actual problem
+		return nil, err
```
pierDipi, line 180: A bit of info on the error:
```suggestion
		return nil, fmt.Errorf("failed to get resource %s/%s: %w", res.GetNamespace(), res.GetName(), err)
```

pierDipi (approved):
/lgtm
/approve

pierDipi:
/hold

There are lint errors

PR #2395, opened 2022-07-08
Only update finalizers on secrets

matzew:
Follow up on #2346 

## Proposed Changes

- narrow scope for `update` verb to `secrets/finalizers`

**Release Note**

```release-note

```

**Docs**


matzew:
/retest
(secret not found)

control-plane/config/eventing-kafka-broker/200-controller/200-controller-cluster-role.yaml
```diff
@@ -80,6 +80,12 @@ rules:
       - get
       - update
```
pierDipi, line 81: should we remove this?

matzew: exactly similar to the others, and with that removed: does not work - I tried

pierDipi: > @pierDipi we can also just keep as is... w/o this PR

yeah, then let's keep it as is

PR #2402, opened 2022-07-11
Ignore artifacts files only in the root directory

pierDipi:
Without the `/` it will ignore files with the same name in all
directories


matzew (approved):
/lgtm
/approve

PR #2412, opened 2022-07-19
Force include all Kafka clients classes

pierDipi:
I'm getting StikyAssignor ClassNotFound when building the data plane
without `Jib` by running `./mvnw package` and using the jar
`data-plane/dispatcher/target/dispatcher-1.0-SNAPSHOT.jar`.

### DIFF
--- a/data-plane/dispatcher/pom.xml
+++ b/data-plane/dispatcher/pom.xml
@@ -156,6 +156,12 @@
                     <include>**</include>
                   </includes>
                 </filter>
+                <filter>
+                  <artifact>org.apache.kafka:kafka-clients</artifact>
+                  <includes>
+                    <include>**</include>
+                  </includes>
+                </filter>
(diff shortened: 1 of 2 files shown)


data-plane/dispatcher/pom.xml
```diff
@@ -156,6 +156,12 @@
                     <include>**</include>
                   </includes>
                 </filter>
+                <filter>
```
aavarghese, line 159: @pierDipi Why is it a `filter` rather than a project `dependency`?

pierDipi: it is a dependency but any class that isn't actually referenced in the final uber Jar will be removed, so when Kafka consumer dynamically asks for a class (uses reflection) it won't find it.

aavarghese (approved):
/lgtm

PR #2429, opened 2022-07-28
Compare all egress fields in data plane reconciler for triggering updates

pierDipi:
We are missing some recently added fields to compare.


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
```diff
@@ -240,8 +240,12 @@ private boolean egressEquals(DataPlaneContract.Egress e1, DataPlaneContract.Egre
     return Objects.equals(e1.getUid(), e2.getUid())
       && Objects.equals(e1.getConsumerGroup(), e2.getConsumerGroup())
       && Objects.equals(e1.getDestination(), e2.getDestination())
+      && Objects.equals(e1.hasDiscardReply(), e2.hasDiscardReply())
       && Objects.equals(e1.getReplyUrl(), e2.getReplyUrl())
       && Objects.equals(e1.getReplyToOriginalTopic(), e2.getReplyToOriginalTopic())
+      && Objects.equals(e1.getDeliveryOrder(), e2.getDeliveryOrder())
+      && Objects.equals(e1.getVReplicas(), e2.getVReplicas())
+      && Objects.equals(e1.getFeatureFlags(), e2.getFeatureFlags())
```
aavarghese, line 248: Should we compare `Reference` field too?

pierDipi: it's not needed, for now, it is immutable and the fact that there is a version (resourceVersion) in the field might trigger unexpected side effects.

aavarghese: ok

aavarghese (approved):
Tests in-progress...

/lgtm

PR #2431, opened 2022-07-28
Add pause/resume lifecyecle to ordered consumer

pierDipi:
By adding the pause/resume lifecycle, we avoid polling events
from partitions that would need to be kept in-memory since
the per-partition queue isn't empty.


aavarghese (approved):
/lgtm

PR #2432, opened 2022-07-28
Perform `max.poll.interval.ms` considering `max.poll.records`

pierDipi:
Without considering `max.poll.records` in the calculation of
the `max.poll.interval.ms` we might cause unnecessary rebalances
since without this patch slow consumers might trigger a rebalance
at every handled record (when their processing time ~= timeout).

```release-note
When using ordered delivery with slow consumers, unnecessary consumer group rebalances might be triggered, which causes higher e2e latency and increased duplicates.
```

/kind bug

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java
@@ -165,7 +165,8 @@ public class ConsumerVerticleFactoryImpl implements ConsumerVerticleFactory {
            egress.getEgressConfig() :
            resource.getEgressConfig();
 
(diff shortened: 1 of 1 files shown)


aavarghese:
@pierDipi @aslom since we're computing the `max.poll.interval.ms` value for both Ordered and Unordered, that implies  [this](https://github.com/aavarghese/eventing-kafka-broker/blob/main/data-plane/config/source/100-config-kafka-source-data-plane.yaml#L116) parameter in consumer properties will get overridden. Should we add this info as comment or remove that parameter?

pierDipi:
> @pierDipi @aslom since we're computing the `max.poll.interval.ms` value for both Ordered and Unordered, that implies [this](https://github.com/aavarghese/eventing-kafka-broker/blob/main/data-plane/config/source/100-config-kafka-source-data-plane.yaml#L116) parameter in consumer properties will get overridden. Should we add this info as comment or remove that parameter?

+1, let's work on that in a separate issue since it's an existing behavior.

aavarghese (approved):
Tests in-progress....

/lgtm

PR #2433, opened 2022-07-28
Fix profiler job

pierDipi:
As per title, flamegraphs can be downloaded from
https://github.com/knative-sandbox/eventing-kafka-broker/actions/runs/2753216718

### DIFF
--- a/data-plane/profiler/run.sh
+++ b/data-plane/profiler/run.sh
@@ -40,7 +40,7 @@ echo "Profiling event ${EVENT}"
 
 PROJECT_ROOT_DIR=$(dirname $0)/..
 RESOURCES_DIR="$(dirname $0)"/resources
-ASYNC_PROFILER_URL="https://github.com/jvm-profiling-tools/async-profiler/releases/download/v2.6/async-profiler-2.6-linux-x64.tar.gz"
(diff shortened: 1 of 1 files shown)


data-plane/profiler/run.sh
```diff
@@ -101,7 +101,7 @@ export WAIT_STARTUP_SECONDS="8"
 export SERVICE_NAME="kafka-broker-receiver"
 export SERVICE_NAMESPACE="knative-eventing"
 export INGRESS_PORT="8080"
-export METRICS_PORT="9090"
+export METRICS_PORT="9098"
 export INSTANCE_ID="receiver"
```
aavarghese, line 105: Why the port # change?

pierDipi: Because sacura started to expose metrics on 9090, so that broke this job :)

matzew: :thinking:

pierDipi: I started to expose metrics on 9090 for perf testing on sacura side, so the receiver and sacura are trying to use the same local port 

https://github.com/pierDipi/sacura/commit/8d35ca8ae2beb65c7e5764d5d25efa0a018f30b4

aavarghese:
/test upgrade-tests_eventing-kafka-broker_main
--


<br class="Apple-interchange-newline">

aavarghese (approved):
/lgtm

PR #2434, opened 2022-07-28
Increase default HTTP client max pool size

pierDipi:
During some perf-testing for KafkaSource, we've observed
that the Vert.x's default for `maxPoolSize` is too low
for parallel event dispatch.

This is just changing the default values to improve the out
of the box experience and the common use case, people can
still change it.

### DIFF
--- a/data-plane/config/broker/100-config-kafka-broker-data-plane.yaml
+++ b/data-plane/config/broker/100-config-kafka-broker-data-plane.yaml
@@ -175,5 +175,6 @@ data:
     # ssl.trustmanager.algorithm
   config-kafka-broker-webclient.properties: |
     idleTimeout=10000
+    maxPoolSize=100
   config-kafka-broker-httpserver.properties: |
     idleTimeout=0
--- a/data-plane/config/channel/100-config-kafka-channel-data-plane.yaml
+++ b/data-plane/config/channel/100-config-kafka-channel-data-plane.yaml
@@ -174,5 +174,6 @@ data:
     # ssl.trustmanager.algorithm
   config-kafka-channel-webclient.properties: |
     idleTimeout=10000
+    maxPoolSize=100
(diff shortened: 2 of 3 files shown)


aavarghese:
@pierDipi @aslom Similar to `max.poll.interval.ms` , this value gets overridden for atleast Source V2 (with scheduler) implementation [here](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java#L344) . Inform that via a comment? Todo: need to change that size to #ofpartitions instead of vreps

pierDipi:
> @pierDipi @aslom Similar to `max.poll.interval.ms` , this value gets overridden for atleast Source V2 (with scheduler) implementation [here](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java#L344) . Inform that via a comment? Todo: need to change that size to #ofpartitions instead of vreps

This makes can we capture it in a different issue?

aavarghese:
> > @pierDipi @aslom Similar to `max.poll.interval.ms` , this value gets overridden for atleast Source V2 (with scheduler) implementation [here](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleFactoryImpl.java#L344) . Inform that via a comment? Todo: need to change that size to #ofpartitions instead of vreps
> 
> This makes can we capture it in a different issue?

Sure, one of us will remember now. :)

aavarghese (approved):
/lgtm

PR #2442, opened 2022-07-29
Pass broker class to e2e tests

aliok:
Fixes #

## Proposed Changes

- Pass broker class to e2e tests
- For now, only `Kafka` broker class and not the namespaced broker class
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -116,7 +116,7 @@ jobs:
       NODE_SHA: ${{ matrix.kind-image-sha }}
 
       EVENTING_KAFKA_BROKER_CHANNEL_AUTH_SCENARIO: ${{ matrix.channel-auth-scenario }}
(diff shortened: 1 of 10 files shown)


.github/workflows/kind-e2e.yaml
```diff
@@ -87,6 +116,7 @@ jobs:
       NODE_SHA: ${{ matrix.kind-image-sha }}
 
       EVENTING_KAFKA_BROKER_CHANNEL_AUTH_SCENARIO: ${{ matrix.channel-auth-scenario }}
+      EVENTING_KAFKA_BROKER_CLASS: ${{ matrix.broker-class }}
```
pierDipi, line 119: Eventing uses "BROKER_CLASS" can we use the same?

https://github.com/knative/eventing/blob/15dd7ca8c81146cd4ff0f5b5ad1d5fe3820df9ec/test/rekt/resources/broker/broker.go#L48-L51

pierDipi (approved):
/lgtm
/approve

PR #2443, opened 2022-07-29
Separate GH actions definition files for components

aliok:
Fixes #

## Proposed Changes

- Separate GH actions definition files for components
-
-

**Release Note**

```release-note

```

**Docs**


pierDipi (commented):
Can we somehow reduce duplication?

Adding a new k8s version, and upgrading ko or go would require lots of changes with this.

aliok:
> Can we somehow reduce duplication?
> Adding a new k8s version, and upgrading ko or go would require lots of changes with this.

I reduce code duplication with this PR by eliminating the strange "exclude" cases.
Now that we have more cases with namespaced broker, we will add even more things there if we keep this structure.

I understand your point though. There is a lot of things that can be shared. I did some reading here already: https://docs.github.com/en/actions/using-workflows/reusing-workflows

However, my preference would be doing this as a copy paste and then extracting a reusable workflow because:
1. Reusing the workflows is a lot of work and is complex. I don't want to waste time on that optimization part of the namespaced broker work.
2. It is nicer to have a copy-paste solution working, then extracting common parts out of that. That way we would have the chance to actually compare if the reused workflow is doing what we actually want.

pierDipi:
Since GH actions are mostly a signal to see if we're still working with all k8s supported versions, given that the only annoying component is the channel auth scenarios and given the plans to support auth options on a per channel basis, I wonder if should just drop channel auth scenarios (maybe leaving only the popular SASL_SSL) from GH actions [1]. We will still test those in prow, so we should be good with coverage.

https://github.com/knative-sandbox/eventing-kafka-broker/blob/89b4bb7ac22a1baa62eab29ed03286de0404ccbe/.github/workflows/kind-e2e.yaml#L27-L31

aliok:
> Since GH actions are mostly a signal to see if we're still working with all k8s supported versions, given that the only annoying component is the channel auth scenarios and given the plans to support auth options on a per channel basis, I wonder if should just drop channel auth scenarios (maybe leaving only the popular SASL_SSL) from GH actions [1]. We will still test those in prow, so we should be good with coverage.

If we're cool skipping those, then I would be happy to just chop things off.

I reverted my changes and pushed a new commit removing the auth tests for channel.

pierDipi (approved):
/lgtm

for GH action tests
/hold

aliok:
/unhold

Getting irrelevant error:
```
Internal error occurred: failed calling webhook "defaulting.webhook.kafka.eventing.knative.dev": failed to call webhook: Post "[https://kafka-webhook-eventing.knative-eventing.svc:443/defaulting?timeout=2s](https://kafka-webhook-eventing.knative-eventing.svc/defaulting?timeout=2s)": dial tcp 10.96.120.39:443: connect: connection refused
```

PR #2452, opened 2022-08-02
Align backoff delay to old source

pierDipi:
As per title


aavarghese (approved):
/lgtm

PR #2453, opened 2022-08-02
Add TODO for max pool size override

pierDipi:


aavarghese (approved):
/lgtm

So today, both deliveries will scale http clients upto 100 till TODO is implemented.

PR #2456, opened 2022-08-02
Reduce default max.poll.records and max.partition.fetch.bytes

pierDipi:
Having very high values by default requires using too much memory
when there are many resources and might also cause a higher chance
to have duplicates.

```release-note
The default max.partition.fetch.bytes is `65536` and `max.poll.records` is 50.
```


aavarghese (approved):
/lgtm

Test failing due to timing out...

matzew:
@pierDipi I like reasoning about this. Thanks.

Was there any report or so hinting to do so?

PR #2495, opened 2022-08-10
Fix ownerRef cannot be set on resources you cannot delete

pierDipi:
In OpenShift, I'm getting the above error when creating a
KafkaSource without the patch.


matzew (approved):
/lgtm
/approve

PR #2497, opened 2022-08-11
Propagate non-ready consumers top level condition in ConsumerGroup

pierDipi:
When a Consumer wasn't ready, we would get a generic error
on the user-facing API which isn't optimal to debug the issue.


matzew (approved):
/lgtm
/approve

PR #2539, opened 2022-08-22
[WIP] Register webhook for KafkaBroker

### LINKED ISSUE #2537: Broker validation webhook is not triggered
**Describe the bug**
Broker validation webhook is not triggered

 **Expected behavior**
Broker validation webhook is to be triggered

 **To Reproduce**
Create a broker with invalid configuration (`spec.config.kind` is not a `ConfigMap`):

```
kind: Broker
metadata:
  annotations:
  name: default3
  namespace: default
spec:
  config:
    apiVersion: v1
    kind: Fuk
    name: kafka-broker-config
    namespace: knative-eventing
```

Broker is not rejected by the webhook.

I added some log statements to the validation code (https://github.com/knative-sandbox/eventing-kafka-broker/blob/2133e8c7068c935aeb07304770f0c3962755d564/control-plane/pkg/apis/eventing/v1/broker_validation.go#L71-L73) and I saw nothing is printed.

```
kubectl get brokers -A
NAMESPACE   NAME       URL   AGE    READY   REASON
default     default3         106s   False   supported config Kind: ConfigMap - got Fuk
```

 **Knative release version**

 **Additional context**

aliok:
Fixes #2537

## Proposed Changes

- Register webhook for KafkaBroker
-
-

TODO:
* Verified locally but I want to make sure about the implications of having multiple webhooks (for other broker implementations and for the knative/eventing broker api type)
* I want to write an e2e test

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/cmd/webhook-kafka/main.go
+++ b/control-plane/cmd/webhook-kafka/main.go
@@ -49,6 +49,7 @@ var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	eventingv1alpha1.SchemeGroupVersion.WithKind("KafkaSink"):    &eventingv1alpha1.KafkaSink{},
 	sourcesv1beta1.SchemeGroupVersion.WithKind("KafkaSource"):    &sourcesv1beta1.KafkaSource{},
 	messagingv1beta1.SchemeGroupVersion.WithKind("KafkaChannel"): &messagingv1beta1.KafkaChannel{},
+	eventingcorev1beta1.SchemeGroupVersion.WithKind("Broker"):    &eventingcorev1beta1.Broker{},
 }
 
 var validationCallbacks = map[schema.GroupVersionKind]validation.Callback{


control-plane/cmd/webhook-kafka/main.go
```diff
@@ -49,6 +49,7 @@ var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	eventingv1alpha1.SchemeGroupVersion.WithKind("KafkaSink"):    &eventingv1alpha1.KafkaSink{},
 	sourcesv1beta1.SchemeGroupVersion.WithKind("KafkaSource"):    &sourcesv1beta1.KafkaSource{},
 	messagingv1beta1.SchemeGroupVersion.WithKind("KafkaChannel"): &messagingv1beta1.KafkaChannel{},
+	eventingcorev1beta1.SchemeGroupVersion.WithKind("Broker"):    &eventingcorev1beta1.Broker{},
```
matzew, line 52: why `v1beta1`?

aliok:
/hold

Waiting for @dprotaso 's input (and perhaps others' as well)

control-plane/cmd/webhook-kafka/main.go
matzew: wait wait, we don't have `beta1` anymore

matzew: Look: the import is just bad named: https://github.com/knative-sandbox/eventing-kafka-broker/blob/5887163fa2ae513db4a598278184ce7fc5538c84/control-plane/cmd/webhook-kafka/main.go#L35

matzew: Also, isn't already here: https://github.com/knative/eventing/blob/main/cmd/webhook/main.go#L63 ?

aliok: Thanks for the heads up. Fixed it now.

dprotaso:
> I thought the Extra validating callback concept is working ([here](https://github.com/knative-sandbox/eventing-kafka-broker/blob/5887163fa2ae513db4a598278184ce7fc5538c84/control-plane/cmd/webhook-kafka/main.go#L136)) but when I don't register the type, they are not called at all.

This is probably a bug in webhook pkg code

aliok:
> I thought the Extra validating callback concept is working ([here](https://github.com/knative-sandbox/eventing-kafka-broker/blob/5887163fa2ae513db4a598278184ce7fc5538c84/control-plane/cmd/webhook-kafka/main.go#L136)) but when I don't register the type, they are not called at all.

I have been thinking more about this... When I don't register the type, then there's no `Broker` resource registered at the `validatingwebhookconfiguration` and of course it won't be triggered. So, this actually makes sense.

@dprotaso what is the use case for the `Extra validating callback` concept ([here](https://github.com/knative-sandbox/eventing-kafka-broker/blob/5887163fa2ae513db4a598278184ce7fc5538c84/control-plane/cmd/webhook-kafka/main.go#L136)) then? Any idea?

dprotaso:
The original intent was to support callbacks without a concrete go type - ie. using `unstructured.Unstructured`

It was added to support PodSpec DryRun

PR #2557, opened 2022-08-24
Move creation of secrets for tests to test/e2e-common.sh

mgencur:
## Proposed Changes

- The secrets are created for tests (they belong to them) so moving their creation into test_setup.
- This also makes it possible to "source" e2e-common.sh and call `create_tls_secrets` or `create_sasl_secrets` directly which is beneficial downstream or when Kafka cluster is already installed and we don't want to run kafka_setup.sh just to install the secrets (which would change the Kafka installation as well). This is basically a better separation of installation and tests.

**Release Note**

```release-note

```

**Docs**


matzew (approved):
/lgtm
/approve

mgencur:
@matzew Thanks!

PR #2562, opened 2022-08-25
Proper webhook registration for KafkaBroker

### LINKED ISSUE #2537: Broker validation webhook is not triggered

aliok:
Fixes #2537

## Proposed Changes

- Register the webhook for `Broker` resource in our webhook. We only had a validation callback before but it was never called because we didn't register the type and that was not added in the admissioncontroller.
- When I simply register the `Broker` resource with the `eventingv1.Broker` type, the webhook automatically calls `Validate()` of that type. That's not good because in that case our webhook would also validate the brokers from other classes.
- Introduced a new stub, `BrokerStub` that doesn't call the `Validate()` on `eventingv1.Broker` type. I only added the validations that are specific to our classes. We don't care about the general things (like missing config, etc) here and just leave the upstream webhook to its job. This way we would properly see the `admission webhook "validation.webhook.kafka.eventing.knative.dev" denied the request` OR `admission webhook "validation.webhook.eventing.knative.dev" denied the request` messages.

For example, when the `spec.config.name` is not specified, we would see this:
```
admission webhook "validation.webhook.eventing.knative.dev" denied the request: validation failed: missing field(s): spec.config.name
```
(not `validation.webhook.kafka.eventing.knative.dev` but `validation.webhook.eventing.knative.dev`)

But, when `spec.config.namespace` is not same as `broker.namespace` (namespaced broker)
```
admission webhook "validation.webhook.kafka.eventing.knative.dev" denied the request: validation failed: invalid value: ConfigMap: spec.config.namespace
Expected ConfigMap in same namespace with broker resource
```

I cared about this and worked on this because I don't want for example try to create a MTChannelBroker with invalid stuff and see "kafka webhook denied request".

As discussed in https://github.com/knative-sandbox/eventing-kafka-broker/pull/2539, I copied the approach from https://github.com/knative/serving/pull/7081

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/cmd/webhook-kafka/main.go
+++ b/control-plane/cmd/webhook-kafka/main.go
@@ -32,7 +32,7 @@ import (
 	"knative.dev/pkg/webhook/resourcesemantics/validation"
 
 	sourcesv1beta1 "knative.dev/eventing-kafka/pkg/apis/sources/v1beta1"
-	eventingcorev1beta1 "knative.dev/eventing/pkg/apis/eventing/v1"
+	eventingcorev1 "knative.dev/eventing/pkg/apis/eventing/v1"
 
 	messagingv1beta1 "knative.dev/eventing-kafka/pkg/apis/messaging/v1beta1"
 
@@ -49,11 +49,11 @@ var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	eventingv1alpha1.SchemeGroupVersion.WithKind("KafkaSink"):    &eventingv1alpha1.KafkaSink{},
 	sourcesv1beta1.SchemeGroupVersion.WithKind("KafkaSource"):    &sourcesv1beta1.KafkaSource{},
 	messagingv1beta1.SchemeGroupVersion.WithKind("KafkaChannel"): &messagingv1beta1.KafkaChannel{},
-	eventingcorev1beta1.SchemeGroupVersion.WithKind("Broker"):    &eventingcorev1beta1.Broker{},
+	eventingcorev1.SchemeGroupVersion.WithKind("Broker"):    &eventingcorev1.Broker{},
 }
 
 var validationCallbacks = map[schema.GroupVersionKind]validation.Callback{
-	eventingcorev1beta1.SchemeGroupVersion.WithKind("Broker"): eventingv1.BrokerValidationCallback(),
+	eventingcorev1.SchemeGroupVersion.WithKind("Broker"): eventingv1.BrokerValidationCallback(),
 }
 
 var defaultingCallbacks = map[schema.GroupVersionKind]defaulting.Callback{
--- a/control-plane/cmd/webhook-kafka/main.go
+++ b/control-plane/cmd/webhook-kafka/main.go
@@ -49,6 +49,7 @@ var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	eventingv1alpha1.SchemeGroupVersion.WithKind("KafkaSink"):    &eventingv1alpha1.KafkaSink{},
 	sourcesv1beta1.SchemeGroupVersion.WithKind("KafkaSource"):    &sourcesv1beta1.KafkaSource{},
 	messagingv1beta1.SchemeGroupVersion.WithKind("KafkaChannel"): &messagingv1beta1.KafkaChannel{},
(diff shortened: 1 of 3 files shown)


aliok:
cc @dprotaso 
Mind having a look?

aliok:
Note: There might be some tests failing now as the webhook was never running and things slipping before are now not slipping.

aliok:
/retest-required

Interesting error:
```
CONT  TestBrokerRedeliveryBrokerV1BackoffLinear/0/fibonacci
    creation.go:550: Failed to create role binding "eventing-e2e9-eventwatcher": rolebindings.rbac.authorization.k8s.io "eventing-e2e9-eventwatcher" is forbidden: user "prow-job@knative-tests.iam.gserviceaccount.com" (groups=["system:authenticated"]) is attempting to grant RBAC permissions not currently held:
```
and the regular `failed to sync configmap cache: timed out waiting for the condition`

aliok:
/hold

I want to have @dprotaso eyeball on this

control-plane/cmd/webhook-kafka/main.go
```diff
@@ -32,7 +32,7 @@ import (
 	"knative.dev/pkg/webhook/resourcesemantics/validation"
 
 	sourcesv1beta1 "knative.dev/eventing-kafka/pkg/apis/sources/v1beta1"
-	eventingcorev1beta1 "knative.dev/eventing/pkg/apis/eventing/v1"
+	eventingcorev1 "knative.dev/eventing/pkg/apis/eventing/v1"
```
aliok, line 35: The import was named badly.

control-plane/pkg/apis/eventing/v1/broker_validation.go
```diff
@@ -21,53 +21,92 @@ import (
...
 	duckv1 "knative.dev/pkg/apis/duck/v1"
-	"knative.dev/pkg/webhook"
-	"knative.dev/pkg/webhook/resourcesemantics/validation"
-
-	"knative.dev/eventing-kafka-broker/control-plane/pkg/kafka"
+	"knative.dev/pkg/webhook/resourcesemantics"
 )
 
-type BrokerPartial struct {
+type BrokerStub struct {
```
aliok, line 35: Renamed existing BrokerPartial (was not used at all) to BrokerStub

control-plane/cmd/webhook-kafka/main.go
```diff
@@ -131,9 +128,6 @@ func NewValidationAdmissionController(ctx context.Context, _ configmap.Watcher)
 
 		// Whether to disallow unknown fields.
 		true,
-
-		// Extra validating callbacks to be applied to resources.
-		validationCallbacks,
```
matzew, line 136: why remove?

aliok: Because it was added before thinking that adding this callback here would automatically make the webhook validate the broker resource. It doesn't work that way.

To make that callback getting triggered, we need to register the resource. And when we do register the resource, we don't need this callback as we can use the proper validation in `BrokerStub`.

Use case for this callback is something different: being able to validate an object that's not implementing `apis.Validatable`, such as a Pod. Callback treats everything as `unstructured` so, that's how it works.

pierDipi (approved):
/lgtm
/approve

PR #2570, opened 2022-08-30
Check for kafka/network errors on finalizeKind, to unblock the borker deletion if the cluster is not reachable

### LINKED ISSUE #2569: Unreachable Kafka cluster prevents broker deletion
Sub issue from https://github.com/knative-sandbox/eventing-kafka-broker/issues/2536

having a broker with a config for a Kafka cluster that does not exist, or is not reachable:

```yaml
kind: Broker
metadata:
  annotations:
  name: my-broken-kafka-broker
spec:
  config:
    apiVersion: v1
    kind: ConfigMap
    name: my-broken-kafka-broker-config
    namespace: default
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-broken-kafka-broker-config
data:
  default.topic.partitions: "10"
  default.topic.replication.factor: "3"
  bootstrap.servers: "matzew-test.kafka.not.existing:443"

```

Will go to a `false` `READY` state:

```
➜  ~ k get brokers.eventing.knative.dev -A 
NAMESPACE   NAME                     URL   AGE   READY   REASON
default     my-broken-kafka-broker         13m   False   cannot obtain Kafka cluster admin, kafka: client has run out of available brokers to talk to: dial tcp: lookup matzew-test.kafka.not.existing:443 on 10.xx.xx:xx: no such host
```

But that none-ready broker is than stuck for deleting, and finalizeKind does not exit correctly

matzew:

Fixes #2569

## Proposed Changes

- If apache kafka is not reachble, we do not longer block the finalization of the broker 

**Release Note**

```release-note
If apache kafka is not reachble, we do not longer block the finalization/deletion of the knative kafka broker 
```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -386,10 +386,17 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker) 
 
 		// get security option for Sarama with secret info in it
 		securityOption := security.NewSaramaSecurityOptionFromSecret(secret)
-		if err := r.finalizeNonExternalBrokerTopic(broker, securityOption, topicConfig, logger); err != nil {
+		err = r.finalizeNonExternalBrokerTopic(broker, securityOption, topicConfig, logger)
+
+		// if finalizeNonExternalBrokerTopic returns error that kafka is not reachable,
+		// we return nil, and clean up the broker
+		if err != nil {
+			if strings.Contains(err.Error(), "cannot obtain Kafka") {
+				logger.Error("Error reaching Kafka cluster", zap.Error(err))
+				return nil
+			}
 			return err
 		}
-
 	}
 
 	if err := r.removeFinalizerSecret(ctx, finalizerSecret(broker), secret); err != nil {
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"go.uber.org/zap"
@@ -57,6 +58,32 @@ const (
 	ConsumerConfigKey = "config-kafka-broker-consumer.properties"
 )
 
+type Counter struct {
+	counterLock sync.RWMutex
+	counterMap  map[string]int
+}
+
+func NewCounter() *Counter {
+	return &Counter{
+		counterLock: sync.RWMutex{},
+		counterMap:  make(map[string]int),
+	}
+}
+
+func (c Counter) Inc(brokerUUID string) int {
+	c.counterLock.Lock()
+	defer c.counterLock.Unlock()
+	c.counterMap[brokerUUID]++
+
+	return c.counterMap[brokerUUID]
+}
+
+func (c Counter) Del(brokerUUID string) {
+	c.counterLock.Lock()
+	defer c.counterLock.Unlock()
+	delete(c.counterMap, brokerUUID)
+}
+
 type Reconciler struct {
 	*base.Reconciler
 	*config.Env
@@ -71,7 +98,8 @@ type Reconciler struct {
 
 	BootstrapServers string
 
-	Prober prober.Prober
+	Prober  prober.Prober
+	Counter *Counter
 }
 
 func (r *Reconciler) ReconcileKind(ctx context.Context, broker *eventing.Broker) reconciler.Event {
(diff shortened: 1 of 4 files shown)


aliok:
Short version:

LGTM, but let's improve in a separate ticket/PR.

Long version:

The bad case that can happen is that one started having network issues and tried deleting the broker. Then we leak the topic. I don't think it is not a good idea to block the deletion for the possibility of network is coming back. We cannot know when it will come back.

However, for short network hiccups that live a short period of time, it would be a shame to not wait and not delete the topic. 

Can we add a retry mechanism? Like, try 5 finalization attempts (preferably with exponential backoff) before giving up. We can store the attempt counts in the resource status or in an annotation. We should never wait for the brokers that were never ready (never created a topic) though. I am 100% ok to merge this PR to unblock namespaced broker usecases, but we can do this improvements in a separate PR.

matzew:
> However, for short network hiccups that live a short period of time, it would be a shame to not wait and not delete the topic.
> Can we add a retry mechanism? 

that is actually a good idea to explore

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -373,7 +373,14 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker)
 
 		// get security option for Sarama with secret info in it
 		securityOption := security.NewSaramaSecurityOptionFromSecret(secret)
-		if err := r.finalizeNonExternalBrokerTopic(broker, securityOption, topicConfig, logger); err != nil {
+		err = r.finalizeNonExternalBrokerTopic(broker, securityOption, topicConfig, logger)
+
+		// if finalizeNonExternalBrokerTopic returns error that kafka is not reachable,
+		// we return nil, and clean up the broker
+		if err != nil {
+			if strings.Contains(err.Error(), "cannot obtain Kafka") {
```
matzew, line 381: Is this error we check for enough?

aliok: When I check what `finalizeNonExternalBrokerTopic` does, I see it returns an error when:
- It cannot build Sarama config (e.g. because of corrupt configuration, deleted configmap, etc)
- It cannot connect to Kafka
- It cannot delete the topic

IMO, we should simply continue with the deletion of the broker resource in all of those 3 cases. There's not much we can do in none of these cases.

Or, if you want to treat for example case#1 differently, then move that case in this method and return the error.

aliok: @pierDipi https://github.com/knative-sandbox/eventing-kafka-broker/pull/2570#issuecomment-1234859223

How to track how many retries we made? In status?

My suggestion is to do this retries in a separate PR though.

matzew:
> I know you're not there with this PR yet

perfect timing :smile:  I was about to update this PR and include the test

aliok:
/retest

Strimzi operator issues

matzew:
@aliok this is ready. Retry is done separately...

aliok:
@matzew Thanks. I see all jobs are green, which is awesome. It proves the issue is fixed.

However, I added some comments here: https://github.com/knative-sandbox/eventing-kafka-broker/pull/2570#discussion_r964565706

control-plane/pkg/reconciler/broker/broker.go
matzew: ok, I can change that - I wanted to be more specific, but I see your point. I will update later

thanks

test/e2e_new/features/broker_deleted_recreated.go
```diff
@@ -82,3 +83,24 @@ func BrokerConfigMapDoesNotExist() *feature.Feature {
 
 	return f
 }
+
+// BrokerCannotReachKafkaCluster tests that a broker can be deleted even when KafkaCluster is unreachable.
+func BrokerCannotReachKafkaCluster() *feature.Feature {
+	f := feature.NewFeatureNamed("delete broker ConfigMap first")
```
pierDipi, line 89: copy-pasted from above?
```suggestion
	f := feature.NewFeatureNamed("delete broker with unreachable Kafka cluster")
```

aliok: LOL my bad

control-plane/pkg/reconciler/broker/broker.go
matzew: I think I disagree. 

I think we should do the retires if we return `nil` on any error. The remote kafka not reachable is IMO a different beast than topic not being able to deleted (see the test-failures when I changed)

matzew:
`Kafka Source ` flakes:

matzew:
/test integration-tests-namespaced-broker_eventing-kafka-broker_main


`Kafka Source` flakes..

control-plane/pkg/reconciler/broker/broker.go
aliok: Ok, sounds good to me.

As we talked in the videochat, we should create multiple functions out of this function and then things would make sense.

aliok (approved):
/lgtm
/approve

PR #2437, opened 2022-07-28
WIP Set triggerv2 and channelv2 as default controller and dispatcher for KEDA scaling plus migration logic for deployment to statefulset

aavarghese:
Signed-off-by: aavarghese <avarghese@us.ibm.com>

## Proposed Changes

- Use [`eventing-autoscaler-keda`](https://github.com/knative-sandbox/eventing-autoscaler-keda) as a library to KEDA scale Kafka resources in `eventing-kafka-broker`. 
- **TODO** (separate PR?): Refactor `ConsumerGroup` resource name to avoid confusion from reusing a known Kafka terminology
- **TODO** Expose only certain KEDA annotations through user facing `KafkaSource` API
- Removed old implementation of KafkaChannel and Trigger Controller and Dispatcher Deployment
- Set KafkaChannel and Trigger (v2) with KEDA Scaling and  HA Scheduling as the new implementation for control-plane and data-plane
- Added migration logic in post-install for deleting old dispatcher deployment

**Release Note**

```release-note
Set KafkaChannel and Trigger (v2) with KEDA Scaling and HA Scheduling as the new implementation for control-plane and data-plane. Added migration logic in post-install for deleting old dispatcher deployment.
```

PR #2595, opened 2022-09-01
Setting KEDA specific annotations from Broker

aavarghese:

## Proposed Changes

- Setting KEDA specific annotations from `Broker` into ConsumerGroup API
- **If KEDA is installed**, `ConsumerGroup` replicas field is set by KEDA controller (no initial value)

### DIFF
--- a/control-plane/pkg/reconciler/trigger/v2/triggerv2.go
+++ b/control-plane/pkg/reconciler/trigger/v2/triggerv2.go
@@ -26,7 +26,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corelisters "k8s.io/client-go/listers/core/v1"
-	"knative.dev/eventing-autoscaler-keda/pkg/reconciler/keda"
 	sources "knative.dev/eventing-kafka/pkg/apis/sources/v1beta1"
 	eventingduck "knative.dev/eventing/pkg/apis/duck/v1"
 	eventing "knative.dev/eventing/pkg/apis/eventing/v1"
@@ -46,6 +45,8 @@ import (
 
 	kedaclientset "knative.dev/eventing-autoscaler-keda/third_party/pkg/client/clientset/versioned"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/apis/eventing/v1alpha1"
+
+	kedafunc "knative.dev/eventing-kafka-broker/control-plane/pkg/keda"
 )
 
 const (
@@ -182,28 +183,8 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, broker *eventing
 
 	//expectedCg.Spec.Replicas = ptr.Int32(2) //to be removed
 
-	// TODO: make these parameters below configurable and maybe unexposed
-	if broker.Annotations != nil {
-		expectedCg.Annotations = map[string]string{}
-		if broker.GetAnnotations()[keda.AutoscalingClassAnnotation] != "" {
-			expectedCg.Annotations[keda.AutoscalingClassAnnotation] = broker.GetAnnotations()[keda.AutoscalingClassAnnotation]
-		}
-		if broker.GetAnnotations()[keda.AutoscalingMinScaleAnnotation] != "" {
-			expectedCg.Annotations[keda.AutoscalingMinScaleAnnotation] = broker.GetAnnotations()[keda.AutoscalingMinScaleAnnotation]
-		}
-		if broker.GetAnnotations()[keda.AutoscalingMaxScaleAnnotation] != "" {
-			expectedCg.Annotations[keda.AutoscalingMaxScaleAnnotation] = broker.GetAnnotations()[keda.AutoscalingMaxScaleAnnotation]
-		}
-		if broker.GetAnnotations()[keda.KedaAutoscalingPollingIntervalAnnotation] != "" {
-			expectedCg.Annotations[keda.KedaAutoscalingPollingIntervalAnnotation] = broker.GetAnnotations()[keda.KedaAutoscalingPollingIntervalAnnotation]
-		}
-		if broker.GetAnnotations()[keda.KedaAutoscalingCooldownPeriodAnnotation] != "" {
-			expectedCg.Annotations[keda.KedaAutoscalingCooldownPeriodAnnotation] = broker.GetAnnotations()[keda.KedaAutoscalingCooldownPeriodAnnotation]
-		}
-		if broker.GetAnnotations()[keda.KedaAutoscalingKafkaLagThreshold] != "" {
-			expectedCg.Annotations[keda.KedaAutoscalingKafkaLagThreshold] = broker.GetAnnotations()[keda.KedaAutoscalingKafkaLagThreshold]
-		}
-	}
+	// TODO: make keda annotation values configurable and maybe unexposed
+	expectedCg.Annotations = kedafunc.SetAutoscalingAnnotations(broker.Annotations)
 
 	if secretName != "" {
 		expectedCg.Spec.Template.Spec.Auth = &internalscg.Auth{
@@ -229,7 +210,7 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, broker *eventing
 		return cg, nil
 	}
 
-	if equality.Semantic.DeepDerivative(expectedCg.Spec, cg.Spec) {
+	if equality.Semantic.DeepDerivative(expectedCg.Spec, cg.Spec) && equality.Semantic.DeepDerivative(expectedCg.Annotations, cg.Annotations) {
 		return cg, nil
 	}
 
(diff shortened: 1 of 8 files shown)


matzew:
RE: *TODO*: I personally like the idea of separate, smaller PRs for work done on-top

control-plane/pkg/reconciler/channel/v2/channelv2.go
```diff
@@ -576,30 +582,65 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, channel *messagi
 		},
 	}
 
+	//expectedCg.Spec.Replicas = ptr.Int32(2) //to be removed
```
matzew, line 585: was this for hacking? :sweat_smile:

data-plane/config/channel/500-dispatcher.yaml
```diff
@@ -114,8 +117,17 @@ spec:
...
+
+          resources:
+            requests:
+              cpu: 1000m
+              # 600Mi for virtual replicas + 100Mi overhead
+              memory: 700Mi
+            limits:
+              cpu: 2000m
+              memory: 1000Mi
+
```
matzew, line 130: got these from tests?

control-plane/pkg/reconciler/channel/v2/channelv2.go
aavarghese: Yes :)

data-plane/config/channel/500-dispatcher.yaml
aavarghese: I believe but passed on from `eventing-kafka` settings.

aavarghese:
Waiting for https://github.com/knative-sandbox/eventing-kafka-broker/pull/2607 and https://github.com/knative-sandbox/eventing-kafka-broker/pull/2608 to merge first.

Part of breaking down https://github.com/knative-sandbox/eventing-kafka-broker/pull/2437 into smaller ones.
This **fourth** one is ready for review and merge.

/cc @pierDipi @matzew

aavarghese:
@pierDipi this ready too for your review

data-plane/config/brokerv2/500-dispatcher.yaml
```diff
@@ -155,6 +159,9 @@ spec:
         - name: config-kafka-broker-data-plane
           configMap:
             name: config-kafka-broker-data-plane
+        - name: kafka-broker-brokers-triggers
+          configMap:
+            name: kafka-broker-brokers-triggers
```
pierDipi, line 164: This volume shouldn't be needed here.

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -118,7 +143,21 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, broker *eventing
...
+	}
+	if err := r.TrackSecret(secret, broker); err != nil {
+		return nil, fmt.Errorf("failed to track secret: %w", err)
+	}
+
+	bootstrapServers := broker.Status.Annotations[kafka.BootstrapServersConfigMapKey]
+	topicName, ok := broker.Annotations[ExternalTopicAnnotation]
+	if !ok {
+		topicName = kafka.BrokerTopic(TopicPrefix, broker)
+	}
```
pierDipi, line 158: There is a status annotation to read the topic name, see the other reconciler here: https://github.com/knative-sandbox/eventing-kafka-broker/blob/d787a6ebb77e26e8ef759fb1774048ed41655e84/control-plane/pkg/reconciler/trigger/trigger.go#L390-L393

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -155,32 +197,47 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, broker *eventing
 		},
 	}
 
+	expectedCg.Spec.Replicas = ptr.Int32(1) //Must be set for Scaled Object ref
```
pierDipi, line 200: Same concern as the other PR, in this way we will override KEDA changes when it changes the replicas.

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -118,7 +143,21 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, broker *eventing
...
 	}
 
-	newcg := &internalscg.ConsumerGroup{
+	secret, err := security.Secret(ctx, &security.AnnotationsSecretLocator{Annotations: broker.Status.Annotations, Namespace: trigger.Namespace}, r.SecretProviderFunc())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get secret: %w", err)
+	}
+	if err := r.TrackSecret(secret, broker); err != nil {
+		return nil, fmt.Errorf("failed to track secret: %w", err)
+	}
```
pierDipi, line 152: Is the tracking needed here since we're not acting on the secret at all?

Also, if we need to track the secret, the parent object should be the one that we're reconciling, so in this case is trigger.

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -41,22 +43,42 @@ import (
...
 	Env                 *config.Env
 	ConsumerGroupLister internalslst.ConsumerGroupLister
 	InternalsClient     internalsclient.Interface
+	KedaClient          kedaclientset.Interface
 }
 
 func (r *Reconciler) ReconcileKind(ctx context.Context, trigger *eventing.Trigger) reconciler.Event {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		return r.reconcileKind(ctx, trigger)
+	})
```
pierDipi, line 78: I don't think this retry loop is needed in this case because it's less likely to be triggered given the different operations.

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -41,22 +43,42 @@ import (
...
 	deliveryOrderAnnotation = "kafka.eventing.knative.dev/delivery.order"
+	// TopicPrefix is the Kafka Broker topic prefix - (topic name: knative-broker-<broker-namespace>-<broker-name>).
+	TopicPrefix = "knative-broker-"
+
+	// ExternalTopicAnnotation for using external kafka topic for the broker
+	ExternalTopicAnnotation = "kafka.eventing.knative.dev/external.topic"
 )
 
 type Reconciler struct {
+	*base.Reconciler
```
pierDipi, line 65: Why do we need it?

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -41,22 +43,42 @@ import (
...
 	BrokerLister        eventinglisters.BrokerLister
 	ConfigMapLister     corelisters.ConfigMapLister
 	EventingClient      eventingclientset.Interface
 	Env                 *config.Env
 	ConsumerGroupLister internalslst.ConsumerGroupLister
 	InternalsClient     internalsclient.Interface
+	KedaClient          kedaclientset.Interface
```
pierDipi, line 72: I think KedaClient isn't used anywhere, can we remove it?

data-plane/config/brokerv2/500-dispatcher.yaml
aavarghese: Same as https://github.com/knative-sandbox/eventing-kafka-broker/pull/2613#discussion_r984801145 @pierDipi

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
aavarghese: You're right - we don't need to track it.

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
aavarghese: It was for the [secretProviderFunc](https://github.com/knative-sandbox/eventing-kafka-broker/pull/2595/files#diff-17a28f62dd8ddd363016236484edbadd40e76c69068b777b11030d93e0c94578R146) but redid that one..

data-plane/config/brokerv2/500-dispatcher.yaml
pierDipi: I see, so we need to migrate them to use the generic values as the source is using:
https://github.com/knative-sandbox/eventing-kafka-broker/blob/d0d8aeb62f813ba69018656d2ab92475ac864cb8/data-plane/config/source/500-dispatcher.yaml#L95-L96

pierDipi: but we should still remove the volume here, see https://github.com/knative-sandbox/eventing-kafka-broker/blob/d0d8aeb62f813ba69018656d2ab92475ac864cb8/data-plane/config/source/500-dispatcher.yaml#L155-L166

pierDipi: see also how it works here https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/pkg/apis/core/pod_defaulting.go, in particular these lines:
https://github.com/knative-sandbox/eventing-kafka-broker/blob/d0d8aeb62f813ba69018656d2ab92475ac864cb8/control-plane/pkg/apis/core/pod_defaulting.go#L85-L87

control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -155,32 +184,45 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, broker *eventing
 		},
 	}
 
+	// TODO: make keda annotation values configurable and maybe unexposed
+	expectedCg.Annotations = kedafunc.SetAutoscalingAnnotations(broker.Annotations)
```
pierDipi, line 188: We should read annotations from the Trigger object since each trigger might want to have a different configs based on the subscriber service ability to process events.

pierDipi (approved):
/lgtm
/approve

Issue #1412, opened 2021-11-04
Allow overriding of Kafka configurations per producer/consumer


jhill072:
**Problem**
From my understanding, configurations for producers and consumers are determined by a single config map, plus or minus a couple of specific properties such as _group.id_. I would like to be able to control some more properties on a case-by-case basis, and I'll focus on a specific example: _auto.offset.reset_. Some of my applications may be fine with receiving the same event multiple times, while others may not be able to (or may not want to) understand past events, but I have to choose exactly one per dispatcher. Since the general operation for updating a Trigger is to recreate the Trigger, this results in a new consumer group (and therefore a new offset).

Another similar option on the producer side would be the acks, where we may want higher throughput at the risk of losing some data.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for? Event producers and event consumers.

**Exit Criteria**
I can set a property on a Trigger, SinkBinding, etc. that can change the default behavior of a single producer or consumer.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve? Based on my quick look into the code, there are events for new egresses and a map for consumer configs that gets instantiated on each new consumer. I would hope to be able to reference labels, annotations, etc. on the Trigger that can override the properties for _that particular consumer_. That doesn't seem too difficult to iterate over an incoming egress, but naming conventions and ensuring this doesn't break existing users could mean more discussion points/analysis.

**Additional context (optional)**

pierDipi:
Hi @jhill072 !

This is definitely useful.

What I was thinking is rather than give event producers and event consumers the ability to set every possible producer and consumer configuration directly, what if we have a set of named profiles (with the ability to define your own) where you can group configurations together and every trigger, broker, etc can choose the profile that best fits that use case?

For example:
```yaml
t1-profile:
  auto.offset.reset=latest
  # ...
t2-profile:
  auto.offset.reset=earliest
  # ...
```

The output would be the global set of configurations merged with the profile-specific configurations that the resource chooses.

jhill072:
That would be fantastic, as I was planning on using this in basically the same way!

pierDipi:
/triage accepted

This issue depends on https://github.com/knative-sandbox/eventing-kafka-broker/issues/1537 since we're changing the way consumers work and the new model facilitates the implementation of this feature.

wSedlacek:
I would be interested in this feature being revisited.
[...]

PR #2601, opened 2022-09-05
Kafka Broker Triggers receive events from after they are Ready by default

pierDipi:
So far, Kafka Broker Triggers receive events from the time before
they were ready by default, contrary to the eventing spec intent.

https://github.com/knative/specs/blob/main/specs/eventing/control-plane.md#trigger-lifecycle

> When a Trigger becomes associated with a Broker (either due to creating the Trigger or the Broker), the Trigger MUST only set the Ready condition to true after the Broker has been configured to send all future events matching the spec.filter to the Trigger's spec.subscriber. The Broker MAY send some events to the Trigger's spec.subscriber prior to the Trigger's Readycondition being set to true. When a Trigger is deleted, the Broker MAY send some additional events to the Trigger's spec.subscriber after the deletion.

(notice the terms "all future events" and "MAY send some events
... prior to the Trigger's Ready condition"), the spec clearly mean 
only future events should be sent, (allowing only "some prior events" 
because of the possible race between Readiness and actual 
broker setup... )

This behavior differs from the MTBroker behavior, so now
it's more consistent by default.

Users can still switch back to the previous behavior by changing the 
`auto.offset.reset=latest` to `earliest` in config-kafka-broker-data-plane

## Proposed Changes

- [Support initializing offsets in Trigger reconciler](https://github.com/knative-sandbox/eventing-kafka-broker/commit/4471293345dbf4149b7a9d42d0f3b939f4338b70)
- [Set auto.offset.reset to latest by default](https://github.com/knative-sandbox/eventing-kafka-broker/commit/113a804a87dcab674fc2c993b5e5143231257e51)
- [Add E2E test](https://github.com/knative-sandbox/eventing-kafka-broker/commit/db14c34d682a004f1d9950cf4e9c623bddefeae2)

**Release Note**

```release-note
Previously, triggers were receiving events that were sent to a Kafka Broker before they were created.
Now triggers receive events that are sent to the broker after the trigger is ready by default, skipping events that were sent previously.
To switch back to the previous behavior, you can change the `auto.offset.reset=latest` configuration to `earliest` in the `config-kafka-broker-data-plane` ConfigMap.
```


control-plane/config/eventing-kafka-broker/200-controller/500-controller.yaml
```diff
@@ -69,15 +69,24 @@ spec:
             - name: SOURCE_DATA_PLANE_CONFIG_MAP_NAMESPACE
               value: knative-eventing
 
-            - name: BROKER_DATA_PLANE_CONFIG_MAP_NAME
+            - name: BROKER_CONTRACT_CONFIG_MAP_NAME
```
aliok, line 72: Can we do this change in a separate PR? I don't think it is a good idea have this refactoring/renaming in this PR.

aliok:
Can you improve the release note? 

Something like:
```
Triggers were doing ...bla... which was not compliant with the Knative Eventing specification.
Now triggers will receive events that are sent to the broker after the trigger is ready by default, skipping events that were sent previously.
To switch back to the previous behavior, you can change the `auto.offset.reset=latest` configuration to `earliest` in the `config-kafka-broker-data-plane` ConfigMap.
```

aliok (changes requested):

aliok:
Is this a fair statement for what this PR is doing: When `auto.offset.reset=latest`, we set the offset to latest for the consumer group in **control-plane** before consumption starts. 

I couldn't really grasp how this would fix our issue here. We set the offset in control plane when trigger is still not ready. We might set the offset and then wait 5 mins before the trigger is ready.

Maybe instead of writing me a reply here, can you update the PR description?

/assign

control-plane/config/eventing-kafka-broker/200-controller/500-controller.yaml
pierDipi: Done in https://github.com/knative-sandbox/eventing-kafka-broker/pull/2602

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -271,6 +274,9 @@ func (r *Reconciler) reconcileBrokerTopic(broker *eventing.Broker, securityOptio
 
 	statusConditionManager.TopicReady(topicName)
 	logger.Debug("Topic created", zap.Any("topic", topicName))
+
+	broker.Status.Annotations[kafka.TopicAnnotation] = topicName
```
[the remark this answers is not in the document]
pierDipi: to read it in the trigger reconciler instead of assuming or duplicating the logic for internal or external topics (or whatever other logic we come up next).

pierDipi:
Waiting for https://github.com/knative-sandbox/eventing-kafka-broker/pull/2602 to go in.

control-plane/pkg/reconciler/broker/broker.go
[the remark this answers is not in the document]
matzew: If we stash more content on the `status annotation`, you think this is still "OK" ? 

https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/pkg/reconciler/broker/broker.go#L503

as the topic name is in theory not on that `kafka-broker-config` CM

pierDipi:
Maven:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.3.0:shade (default) on project receiver: Execution default of goal org.apache.maven.plugins:maven-shade-plugin:3.3.0:shade failed: Plugin org.apache.maven.plugins:maven-shade-plugin:3.3.0 or one of its dependencies could not be resolved: Could not transfer artifact org.apache.maven:maven-artifact:jar:3.0 from/to central (https://repo.maven.apache.org/maven2): transfer failed for https://repo.maven.apache.org/maven2/org/apache/maven/maven-artifact/3.0/maven-artifact-3.0.jar: Connection reset -> [Help 1]
```
/test unit-tests_eventing-kafka-broker_main

control-plane/config/eventing-kafka-broker/200-controller/500-controller.yaml
```diff
@@ -78,6 +78,15 @@ spec:
...
 
+            - name: BROKER_DATA_PLANE_CONFIG_CONFIG_MAP_NAME
+              value: config-kafka-broker-data-plane
+            - name: SOURCE_DATA_PLANE_CONFIG_CONFIG_MAP_NAME
+              value: config-kafka-source-data-plane
+            - name: SINK_DATA_PLANE_CONFIG_CONFIG_MAP_NAME
+              value: config-kafka-source-data-plane
+            - name: CHANNEL_DATA_PLANE_CONFIG_CONFIG_MAP_NAME
+              value: config-kafka-source-data-plane
+
```
aliok, line 89: I just realized that we need to make adjustments to this approach when we have dataplane configurations per namespace in the future.

With this PR, we always read the dataplaneConfigmap from the system namespace right?

pierDipi: I'm not sure this is just the name, I think we can still require to have this exact name but in their namespace.

afaik, it's not a requirement that we have so far.

pierDipi: however, "delivery profiles" feature is the way to go IMHO https://github.com/knative-sandbox/eventing-kafka-broker/issues/1412

control-plane/pkg/kafka/topic.go
```diff
@@ -31,6 +31,8 @@ const (
 	DefaultTopicNumPartitionConfigMapKey      = "default.topic.partitions"
 	DefaultTopicReplicationFactorConfigMapKey = "default.topic.replication.factor"
 	BootstrapServersConfigMapKey              = "bootstrap.servers"
+
+	TopicAnnotation = "default.topic"
```
aliok, line 35: Can you explain why we need to store the topic in the annotation?

pierDipi: https://github.com/knative-sandbox/eventing-kafka-broker/pull/2601#discussion_r963534251

control-plane/pkg/reconciler/trigger/controller.go
```diff
@@ -119,6 +126,9 @@ func NewController(ctx context.Context, watcher configmap.Watcher, configs *conf
 		},
 	})
 
+	reconciler.SecretTracker = impl.Tracker
+	secretinformer.Get(ctx).Informer().AddEventHandler(controller.HandleAll(reconciler.SecretTracker.OnChanged))
```
aliok, line 130: Did we simply forget this OR we just need this tracking because of the changes here?

control-plane/pkg/reconciler/broker/broker.go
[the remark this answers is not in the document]
pierDipi: I think it's ok, it's mostly to decouple concerns, broker owns the topic

 (Btw, I also need this for the scaling feature)

control-plane/pkg/reconciler/trigger/controller.go
pierDipi: because of the changes here

pierDipi:
Failed https://github.com/knative-sandbox/eventing-kafka-broker/issues/2612 for unrelated error

/test upgrade-tests_eventing-kafka-broker_main

pierDipi:
I rebased

aliok (approved):
/lgtm
/approve

PR #2604, opened 2022-09-06
Doing base 64 encoding of the data

### LINKED ISSUE #2594: kafka dispatcher "SEVERE: Unhandled exception" when a sink returns a non-cloudevent response with an error status
**Describe the bug**
Having a sink returning a 404 status with a body

```
<!doctype html>
<html>
<head>
    <title>Error Page(tm)</title></head>
<body>
<p>Quoth the server, 404!
</body></html>
```

causes an Unhandled exception in the kafka dispatcher

```
SEVERE: Unhandled exception
java.lang.IllegalArgumentException: only ' ' and '\t' are allowed after '\n'
```

Furthermore, the event that was sent to the sink is not processed further (there is no attempt at sending it to a deadLetterSink when this error occured)

 **Expected behavior**
Dispatcher should not log an "SEVERE: Unhandled exception" and should retry a retry-able status code from a sink even if it contains any random response body

 **To Reproduce**
With an existing Kafka Broker config, it can be reproduce by something like
```
kind: Broker
metadata:
  name: broker
  annotations:
spec:
  config:
    apiVersion: v1
    kind: ConfigMap
    name: broker
  delivery:
    backoffDelay: PT1S
    backoffPolicy: linear
    deadLetterSink:
      ref:
        apiVersion: eventing.knative.dev/v1alpha1
        kind: KafkaSink
        name: dlq-sink
    retry: 5
---
kind: Trigger
metadata:
  name: subscription
spec:
  broker: broker
  subscriber:
    uri: http://example.com/doesnotexist
```

 **Knative release version**
1.4

 **Additional context**
Using OpenShift Serverless 1.25 from current `main` https://github.com/openshift-knative/serverless-operator

matzew:

Fixes #2594 

## Proposed Changes

- Avoiding "invalid" HTTP header values for CE extension, we encode the actual data of the `knativeerrordata` extension.
- updated unit and reconciler tests to reflect the base64 encoding
- added a new case for the returning http error and webpage...

**Release Note**

```release-note
We are doing base64 encoding for the value of the `knativeerrordata`, to avoid issues with adding the data to the HTTP header, when the message is delivered to the deadlettersink. Consumers of the `knativeerrordata` must decode it in case they are interested in it. 
```

**Docs**

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -295,7 +296,8 @@ public class RecordDispatcherImpl implements RecordDispatcher {
       if (data.length() > KN_ERROR_DATA_MAX_BYTES) {
         data = data.substring(0, KN_ERROR_DATA_MAX_BYTES);
       }
-      extensions.put(KN_ERROR_DATA_EXT_NAME, data);
+
+      extensions.put(KN_ERROR_DATA_EXT_NAME, Base64.getEncoder().encodeToString(data.getBytes()));
     }
 
     return addExtensions(recordContext, extensions);
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
@@ -34,6 +34,7 @@ import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.ext.web.client.HttpResponse;
 import io.vertx.kafka.client.common.tracing.ConsumerTracer;
 import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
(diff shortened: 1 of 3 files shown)


matzew:
Need to write a test for this.

Instead of doing:
```java
extensions.put(KN_ERROR_DATA_EXT_NAME, Base64.getEncoder().encodeToString(data.getBytes()));
```

we could also sanitize the data, by using `replaceAll` on the data `string`, but this has an an impact on the actual data, since we manipulate it.

I think we need to update the documentation that, that `knativeerrordata` extension is _always_ base 64 encoded...

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/RecordDispatcherImpl.java
```diff
@@ -295,7 +296,8 @@ private ConsumerRecordContext errorTransform(final ConsumerRecordContext recordC
       if (data.length() > KN_ERROR_DATA_MAX_BYTES) {
         data = data.substring(0, KN_ERROR_DATA_MAX_BYTES);
       }
-      extensions.put(KN_ERROR_DATA_EXT_NAME, data);
+
+      extensions.put(KN_ERROR_DATA_EXT_NAME, Base64.getEncoder().encodeToString(data.getBytes()));
```
pierDipi, line 300: One problem with transforming the response body to string and then base64'ing it is that it might produce unserializable bytes on the other side.

For example, when the body is a protobuf message, converting bytes to string and then base64'ing it will cause the message to be unserializable, is there any way to base64 the response body as bytes?

pierDipi: To clarify:

we do:
```java
data = response.bodyAsString();
Base64.getEncoder().encodeToString(data.getBytes())
```

matzew: Good point, I think there is some vertx byte buffer/array

pierDipi (approved):
/lgtm
/approve

pierDipi:
@matzew please adjust release notes

pierDipi:
@matzew I think we should do the same in eventing core

matzew:
> I  think we should do the same in eventing core

we dont have that on the broker, but I think the channels.
Will add soon-ish

PR #2608, opened 2022-09-06
Setting KEDA specific annotations from KafkaSource

aavarghese:
Signed-off-by: aavarghese <avarghese@us.ibm.com>

## Proposed Changes

- Setting KEDA specific annotations from `KafkaSource` into ConsumerGroup API
- **If KEDA is installed**, `ConsumerGroup` replicas field is set by KEDA controller (initial value == KafkaSource consumers)
- **If KEDA is not installed**, `ConsumerGroup` replicas field is determined by KafkaSource controller

PR #2640, opened 2022-09-09
Keda enabled check function for KafkaSource

aavarghese:
As per https://github.com/knative-sandbox/eventing-kafka-broker/pull/2608#discussion_r966788144


control-plane/pkg/reconciler/source/source.go
```diff
@@ -223,3 +217,15 @@ func propagateConsumerGroupStatus(cg *internalscg.ConsumerGroup, ks *sources.Kaf
...
+func (r *Reconciler) isKEDAEnabled(ctx context.Context, namespace string) bool {
+	// TODO: code below failing unit tests with err: "panic: interface conversion: testing.ActionImpl is not testing.GetAction: missing method GetName"
+	/*if err := discovery.ServerSupportsVersion(r.KubeClient.Discovery(), keda.KedaSchemeGroupVersion); err == nil {
+		return true
+	}*/
+
+	if _, err := r.KedaClient.KedaV1alpha1().ScaledObjects(namespace).List(ctx, metav1.ListOptions{}); err == nil || !apierrors.IsNotFound(err) {
+		return true
+	}
+	return false
```
pierDipi, line 230: should we stay on `false` when there is a generic error `err != nil && !apierrors.IsNotFound(err)`?

pierDipi (approved):
Thank you!

/lgtm
/approve

PR #2641, opened 2022-09-12
NIT: Reflect that the ContentMode for KafkaSink is binary

matzew:

## Proposed Changes

- In #2391 we changed the default of the `ContentMode` for the `KafkaSink` CRD, but we did not change the doc on the actual `types` file

**Release Note**

```release-note

```

**Docs**


pierDipi (approved):
/lgtm
/approve

PR #2642, opened 2022-09-12
KEDA as scaling library minor fix in TriggerAuth gen

aavarghese:

Fix for new code in #2607 

/cc @pierDipi


pierDipi (approved):
/lgtm
/approve

PR #2652, opened 2022-09-15
Remove noisy log lines "Nothing changed" at info level

pierDipi:
When egress changes are detected log the egress info
at debug level instead of logging egresses that weren't
changed at info.


matzew (approved):
/lgtm
/approve

PR #2653, opened 2022-09-15
KEDA Trigger Authentication SASL updates for secret spec

aavarghese:

### DIFF
--- a/control-plane/pkg/reconciler/consumergroup/consumergroup.go
+++ b/control-plane/pkg/reconciler/consumergroup/consumergroup.go
@@ -455,7 +455,12 @@ func (r Reconciler) reconcileKedaObjects(ctx context.Context, cg *kafkainternals
 			return err
 		}
 
-		triggerAuthentication, secret, err = kedafunc.GenerateTriggerAuthentication(cg, saslType)
+		protocol, err := r.retrieveProtocolIfPresent(ctx, cg)
+		if err != nil {
+			return err
+		}
+
+		triggerAuthentication, secret, err = kedafunc.GenerateTriggerAuthentication(cg, saslType, protocol)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


aavarghese:
KEDA integrations works for Channels and Triggers now with SASL types in SecretSpec. Will add support for TLS in a following PR...

control-plane/pkg/reconciler/consumergroup/consumergroup.go
```diff
@@ -489,17 +494,39 @@ func (r Reconciler) reconcileKedaObjects(ctx context.Context, cg *kafkainternals
...
+
+func (r *Reconciler) retrieveProtocolIfPresent(ctx context.Context, cg *kafkainternals.ConsumerGroup) (*string, error) {
+	if hasSecretSpecConfig(cg.Spec.Template.Spec.Auth) && cg.Spec.Template.Spec.Auth.SecretSpec.Ref != nil {
+		secretKeyRefName := cg.Spec.Template.Spec.Auth.SecretSpec.Ref.Name
+		secretKeyRefNamespace := cg.Spec.Template.Spec.Auth.SecretSpec.Ref.Namespace
+		secret, err := r.KubeClient.CoreV1().Secrets(secretKeyRefNamespace).Get(ctx, secretKeyRefName, metav1.GetOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("unable to get Protocol from secret: \"%s/%s\", %w", secretKeyRefNamespace, secretKeyRefName, err)
 		}
+		protocolValue := string(secret.Data["protocol"])
```
pierDipi, line 528: Should we check if the protocol key is actually there to avoid panics?
```
protocolValue, ok := secret.Data["protocol"]
// ...
```

pierDipi: We can also use the constants in security package for keys `protocol`, etc.

pierDipi: sounds good!

/lgtm

PR #2667, opened 2022-09-23
Wait for sts to exists before deleting dispatcher during migration

pierDipi:
This reduces downtime and latency during migration.

### DIFF
--- a/control-plane/cmd/post-install/kafka_source_dispatcher_deleter.go
+++ b/control-plane/cmd/post-install/kafka_source_dispatcher_deleter.go
@@ -19,13 +19,19 @@ package main
 import (
 	"context"
 	"fmt"
+	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"knative.dev/pkg/system"
 )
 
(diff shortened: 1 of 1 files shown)


control-plane/cmd/post-install/kafka_source_dispatcher_deleter.go
```diff
@@ -38,3 +44,17 @@ func (d *kafkaSourceDeleter) DeleteDispatcher(ctx context.Context) error {
 
 	return nil
 }
+
+func (d *kafkaSourceDeleter) waitStatefulSetExists(ctx context.Context) error {
+	const sourceDispatcherStatefulSetName = "kafka-source-dispatcher"
```
aavarghese, line 49: There's also a declaration of the same constant on [line 36](https://github.com/knative-sandbox/eventing-kafka-broker/pull/2667/files#diff-60bca8e7937b5364cdc18d6afe44c8e11e5df784e6a155014a1e8574f63a8b8bR36). Maybe we declare one outside or pass name as func input?

pierDipi: I separated them because, in theory, they can have different names since one is the statefulset and one is the deployment.

aavarghese: Ok.

aavarghese (approved):
/lgtm

PR #2672, opened 2022-09-26
Spec.Config.Namespace is not required

pierDipi:
`Spec.Config.Namespace` is not required since the logical default
is the broker's namespace.


matzew (approved):
/lgtm
/approve

PR #2684, opened 2022-09-27
Fix misleading log message

pierDipi:
It's not here that we check for retry-able status codes.


matzew (approved):
/lgtm
/approve

PR #2702, opened 2022-10-05
Add owner reference only when it is not present

pierDipi:
I noticed that sometimes we try to add the same owner
reference multiple times.

### DIFF
--- a/control-plane/pkg/reconciler/consumer/consumer.go
+++ b/control-plane/pkg/reconciler/consumer/consumer.go
@@ -401,11 +401,12 @@ func cmNameFromPod(p *corev1.Pod, c *kafkainternals.Consumer) (string, error) {
 func podOwnerReference(p *corev1.Pod) base.ConfigMapOption {
 
 	expected := metav1.OwnerReference{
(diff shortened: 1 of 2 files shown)


control-plane/pkg/reconciler/consumer/consumer.go
```diff
@@ -398,15 +399,21 @@ func cmNameFromPod(p *corev1.Pod, c *kafkainternals.Consumer) (string, error) {
...
+		Controller: pointer.Bool(true),
+	}
 	return func(cm *corev1.ConfigMap) {
-		cm.OwnerReferences = append(cm.OwnerReferences, metav1.OwnerReference{
-			APIVersion:         corev1.SchemeGroupVersion.String(),
-			Kind:               "Pod",
-			Name:               p.Name,
-			UID:                p.UID,
-			Controller:         pointer.Bool(true),
-			BlockOwnerDeletion: pointer.Bool(true),
```
aavarghese, line 408: @pierDipi why the removing of `BlockOwnerDeletion` in this PR?

aavarghese (approved):
/lgtm

PR #2703, opened 2022-10-05
Adding consumergroup and consumer to defaulting webhook plus validation logic updates

aavarghese:

As per title

/cc @pierDipi

### DIFF
--- a/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_validation_test.go
+++ b/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_validation_test.go
@@ -51,35 +51,13 @@ func TestConsumerGroup_Validate(t *testing.T) {
 			wantErr: true,
 		},
 		{
-			name: "no topics",
+			name: "no delivery/subscriber",
 			ctx:  context.Background(),
 			given: &ConsumerGroup{
 				Spec: ConsumerGroupSpec{
 					Replicas: pointer.Int32Ptr(1),
 					Selector: map[string]string{"app": "app"},
 					Template: ConsumerTemplateSpec{},
-				},
-			},
-			wantErr: true,
-		},
-		{
-			name: "missing group.id",
-			ctx:  context.Background(),
-			given: &ConsumerGroup{
-				Spec: ConsumerGroupSpec{
-					Replicas: pointer.Int32Ptr(1),
-					Selector: map[string]string{"app": "app"},
-					Template: ConsumerTemplateSpec{
-						Spec: ConsumerSpec{
-							Topics: []string{"t1"},
-							Subscriber: duckv1.Destination{
-								URI: &apis.URL{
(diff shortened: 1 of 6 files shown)


control-plane/pkg/reconciler/source/source.go
```diff
@@ -131,7 +131,7 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, ks *sources.Kafk
 							Retry:         pointer.Int32(10),
 							BackoffPolicy: &backoffPolicy,
 							BackoffDelay:  pointer.String("PT0.3S"),
-							Timeout:       pointer.String("PT600S"),
+							//Timeout:       pointer.String("PT600S"), //experimental feature that needs to be enabled
```
pierDipi, line 134: I think we should keep the timeout regardless of the experimental feature enabled/disabled because we want to set timeouts regardless for reliability.

control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_validation_test.go
```diff
@@ -62,28 +62,6 @@ func TestConsumerGroup_Validate(t *testing.T) {
 			},
 			wantErr: true,
 		},
-		{
```
pierDipi, line 65: Can we keep at lease one test that verifies that the ConsumerTemplateSpec is validated as part of the ConsumerGroup validation?

control-plane/pkg/reconciler/source/source.go
aavarghese: @pierDipi Wondering if there is a way around this. We should have the feature enabled in our EKB setup I assumed but its not. Consumer validation code calls for delivery validation and [this](https://github.com/knative-sandbox/eventing-kafka-broker/pull/2703#issuecomment-1268939993) error is thrown in our controller: https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/vendor/knative.dev/eventing/pkg/apis/duck/v1/delivery_types.go#L106 for timeout

pierDipi: The timeout feature is enabled https://github.com/knative-sandbox/eventing-kafka-broker/blob/77e7b96c7a030ee5921f6470a086218a0d2bbe6e/third_party/eventing-latest/eventing-core.yaml#L451-L453, I think the missing part is to inject the feature flags in the context of our webhook.

pierDipi: This is how eventing does it: https://github.com/knative/eventing/blob/fdacfa352be704cbf00187ecfbb0b115e28a4752/cmd/webhook/main.go#L96-L97

aavarghese: Thank you Pierangelo for sharing the link - did not realize the webhook ctx was missing it.

pierDipi (approved):
Thanks!

/lgtm
/approve

PR #2716, opened 2022-10-07
Source HA cleanup minor changes

aavarghese:

Dispatches replicas setting initially to 0 so statefulset can autoscale up as needed by scheduler and do the eventual necessary pod binding plus config map creation, etc...

Removing old contract config map env var...

/cc @pierDipi

### DIFF
--- a/control-plane/config/eventing-kafka-broker/200-controller/500-controller.yaml
+++ b/control-plane/config/eventing-kafka-broker/200-controller/500-controller.yaml
@@ -75,8 +75,6 @@ spec:
               value: kafka-channel-channels-subscriptions
             - name: SINK_CONTRACT_CONFIG_MAP_NAME
               value: kafka-sink-sinks
-            - name: SOURCE_CONTRACT_CONFIG_MAP_NAME
-              value: kafka-source-sources
 
             - name: BROKER_DATA_PLANE_CONFIG_CONFIG_MAP_NAME
               value: config-kafka-broker-data-plane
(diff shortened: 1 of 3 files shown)


data-plane/config/source/500-dispatcher.yaml
```diff
@@ -23,6 +23,7 @@ metadata:
     app: kafka-source-dispatcher
     kafka.eventing.knative.dev/release: devel
 spec:
+  replicas: 0
```
pierDipi, line 26: This won't work with the downstream operator or any GitOps setup because these systems reconcile these manifests so they will revert any changes done by the autoscaler.

control-plane/config/eventing-kafka-broker/200-controller/500-controller.yaml
```diff
@@ -75,8 +75,6 @@ spec:
               value: kafka-channel-channels-subscriptions
             - name: SINK_CONTRACT_CONFIG_MAP_NAME
               value: kafka-sink-sinks
-            - name: SOURCE_CONTRACT_CONFIG_MAP_NAME
-              value: kafka-source-sources
```
pierDipi, line 79: I think it is there because it's a required configuration in the struct but we don't use it?

aavarghese: We don't use the Env struct for Source anywhere so this is not required, I believe

data-plane/config/source/500-dispatcher.yaml
aavarghese: Oh boy - yeah, the operator world. Thanks for catching that!

pierDipi (approved):
/lgtm
/approve

PR #2725, opened 2022-10-14
Removing dependence on eventing-autoscaler-keda

aavarghese:

Depending on `eventing-autoscaler-keda` causes complication when pulling in dependencies...We only need kedacore APIs and generated codegen files for KEDA integration with Kafka components here.


pierDipi (approved):
Thanks!

/lgtm
/approve

PR #2741, opened 2022-10-24
Namespaced broker upgrade tests

aliok:
Fixes #

## Proposed Changes

- Read broker class env var passed by Prow in upgrade tests
- We will later use this to run upgrade tests for namespaced broker
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/pkg/broker/broker.go
+++ b/test/pkg/broker/broker.go
@@ -54,6 +54,12 @@ func Creator(client *eventingtestlib.Client, version string) string {
 	return CreatorWithConfigOptions(client, version)
 }
 
+func CreatorForClass(class string) func(*eventingtestlib.Client, string) string {
+	return func(client *eventingtestlib.Client, version string) string {
+		return CreatorForClassWithOptions(class, client, version, []resources.BrokerOption{}, []ConfigOptions{})
+	}
+}
+
 type ConfigOptions func(data map[string]string)
 
 func CreatorWithConfigOptions(client *eventingtestlib.Client, version string, configOptions ...ConfigOptions) string {
@@ -70,6 +76,10 @@ func CreatorWithOptions(client *eventingtestlib.Client, version string, brokerOp
 		panic(fmt.Sprintf("error getting KafkaBroker class from env '%v'", err))
 	}
 
+	return CreatorForClassWithOptions(class, client, version, brokerOptions, configOptions)
+}
+
+func CreatorForClassWithOptions(class string, client *eventingtestlib.Client, version string, brokerOptions []resources.BrokerOption, configOptions []ConfigOptions) string {
 	name := "broker"
 
 	version = strings.ToLower(version)
--- a/test/pkg/broker/broker.go
+++ b/test/pkg/broker/broker.go
@@ -58,6 +58,7 @@ func WithBrokerClassFromEnvVar(b *eventingv1.Broker) {
 	}
 	resources.WithBrokerClassForBroker(class)(b)
 }
+
 // Creator creates a Broker with the class that is specified in an environment variable.
 func Creator(client *eventingtestlib.Client, version string) string {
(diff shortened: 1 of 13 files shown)


matzew (approved):
/lgtm
/approve

matzew:
Do we know, why on the release branch we had to do this:
https://github.com/knative-sandbox/eventing-kafka-broker/pull/2736/files

I don't know if the "branch" job are configured differently, or where to change those /cc @aliok

aliok:
@matzew https://github.com/knative/test-infra/pull/3587 should fix the release process issue for 1.8 and for new branches.

pierDipi:
Why a standalone job for upgrade test of this new class, IMHO the upgrade test should be for both classes, what if classes are upgradable in isolation but they don't work with each other?

pierDipi:
Downstream, we won't use this solution, so we need another one for downstream (which is not good) ...

matzew:
>  IMHO the upgrade test should be for both classes

isn't the there one test that reads the passed enviroments?  Or do you mean the one test is invoked separately, with different args passed in. 

Perhaps yes, we have tests in parallel, that are "namespaced" and "global"?

> what if classes are upgradable in isolation 
Are they  upgradable in isolation? 

>but they don't work with each other?
perhaps we have a test that tests them together - instead the two different execution?

pierDipi:
> > what if classes are upgradable in isolation
> 
> Are they upgradable in isolation?

Maybe yes, but that's not the way we ship our software so we're not testing how users would upgrade the software.

> 
> > but they don't work with each other?
> 
> perhaps we have a test that tests them together - instead the two different execution?

Yes, we should add upgrade tests for namespace broker as part of the upgrade test suite https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/test/upgrade/suite.go

aliok:
I am ok with testing both classes in one go. Let me craft a PR for that.

aliok:
@pierDipi @matzew Ready for initial review.
- Added new test functions for namespaced broker

test/pkg/broker/broker.go
```diff
@@ -70,6 +76,10 @@ func CreatorWithOptions(client *eventingtestlib.Client, version string, brokerOp
 		panic(fmt.Sprintf("error getting KafkaBroker class from env '%v'", err))
 	}
 
+	return CreatorForClassWithOptions(class, client, version, brokerOptions, configOptions)
+}
+
+func CreatorForClassWithOptions(class string, client *eventingtestlib.Client, version string, brokerOptions []resources.BrokerOption, configOptions []ConfigOptions) string {
```
pierDipi, line 82: We can reduce the class parameter to a BrokerOption

pierDipi (commented):
LGTM

left a comment to avoid the route of 100 functions with different parameters.

test/pkg/broker/broker.go
pierDipi: There could even be a function already with that purpose.

aliok: I thought about it, but I didn't want to do it. It is because, if I make the class a BrokerOption, then I need to create a separate function that reads the env var for class anyway. So, the number of functions won't change and we would have function that's very implicit about the class.

pierDipi: It's the parameters of this function that I don't think it's readable and maintainable long term, not the functions number, so I still think it's worth doing it now.

I had to deal with this bad pattern in the past for reconciler tests due to us passing raw inputs to function instead of proper arguments and objects and I wasted so much time refactoring code due to the quick and dirty additions of parameters without having a good arguments design.

This is exploding in parameters, like class (and, fwiw existing version) can and should be `BrokerOption`s.

pierDipi (approved):
Thank you!

/lgtm
/approve

Issue #2729, opened 2022-10-17
[e2e] [s390x]: sacura deployments fails intermittently due to prior webhook deployment scaling not having finished


haubenr:
**Describe the bug**
There are intemittent e2e test failures on the s390x platform where sacura fails to deploy due to the prior webhook deployment scaling not having finished in time, e.g.: https://prow.knative.dev/view/gs/knative-prow/logs/s390x-e2e-tests-ssl_eventing-kafka-broker_main_periodic/1581907930591006720

Introducing a waiting period after the webhook deployment (and all other deployments that are part of the scaling operation) should address the issue.

 **Expected behavior**
The sacura deployment works fine for every single test run.

 **To Reproduce**
Run e2e tests on s390x platform, see also: https://prow.knative.dev/view/gs/knative-prow/logs/s390x-e2e-tests-ssl_eventing-kafka-broker_main_periodic/1581907930591006720

 **Knative release version**
1.7
main

 **Additional context**
Affected code:
https://github.com/knative-sandbox/eventing-kafka-broker/blob/87bb6457d56e89402883c1f49ec0d924780404f4/test/e2e-common.sh#L247
https://github.com/knative-sandbox/eventing-kafka-broker/blob/87bb6457d56e89402883c1f49ec0d924780404f4/test/e2e-tests.sh#L11

PR #2744, opened 2022-10-25
Increased wait timeout, fail test if scaling times out

haubenr:
Signed-off-by: Dirk Haubenreisser <haubenr@de.ibm.com>

# Fixes

This PR is a follow-up to issue #2729 
While the initial PR for that issue did address the knative deployment scaling hiccups on the s390x platform to ,say, 95% there is still the occasional test run that fails due to the knative webhook deployment not being available at the time when sacura is to be deployed (see: https://prow.knative.dev/view/gs/knative-prow/logs/s390x-e2e-tests-sasl-plain_eventing-kafka-broker_release-1.8_periodic/1584595827215568896)

## Proposed Changes

The change in this PR attempts to fix the issue fully by
a) increasing the wait timeout for the knative deployment scaling to 10 minutes
b) failing the test run deliberately in case the deployment scaling timeout is reached


pierDipi (approved):
/lgtm
/approve

PR #2770, opened 2022-11-04
Run hack/update-codegen.sh with Go 1.19

pierDipi:
As per title.

Go 1.19 is now required to generate and verify generated code.


aavarghese (approved):
/lgtm

Thank you!!

PR #2775, opened 2022-11-08
Fix for using namespace of the broker instead of status annotation namespace

aavarghese:


pierDipi (approved):
Thanks!

/lgtm

PR #2793, opened 2022-11-17
Retry Only If the Status Code Is Retryable

VikramVuppla:
Signed-off-by: Vikram Vuppla naga.vicky@gmail.com

Fixes #

Why:
- Currently the retry is happening regardless of the response status code is retryable or not, remove the buggy retry logic

## Proposed Changes

- :bug: Fix bug

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
@@ -133,6 +133,11 @@ public final class WebClientCloudEventSender implements CloudEventSender {
           if (isRetryableStatusCode(response.statusCode()) && retryCounter < consumerVerticleContext.getEgressConfig().getRetry()) {
             return retry(retryCounter, event);
           }
+          return Future.failedFuture(cause);
+        }
+
+        if (retryCounter < consumerVerticleContext.getEgressConfig().getRetry()) {
+          return retry(retryCounter, event);
         }
 
         return Future.failedFuture(cause);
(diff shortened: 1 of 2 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
```diff
@@ -135,10 +135,6 @@ private Future<HttpResponse<Buffer>> send(final CloudEvent event, final int retr
           }
         }
 
-        if (retryCounter < consumerVerticleContext.getEgressConfig().getRetry()) {
-          return retry(retryCounter, event);
-        }
```
pierDipi, line 140: This is handling the case where we have a failure but without a response, I think we're missing a return at line 135 when this `if (isRetryableStatusCode(response.statusCode()) && retryCounter < consumerVerticleContext.getEgressConfig().getRetry())` is false

pierDipi (commented):
Thanks for pointing this out @VikramVuppla!

VikramVuppla:
thank you @pierDipi for the review, addressed the review comments, please re-review

pierDipi (commented):
Thanks!

Can we add a new unit test covering this case in WebClientCloudEventSenderTest (see example test `shouldRetry`?

Also, EasyCLA complains for the unsigned Linux Foundation CLA: https://github.com/knative-sandbox/eventing-kafka-broker/pull/2793#issuecomment-1319316385

VikramVuppla:
@pierDipi EasyCLA is now covered and added the unit test case, could you please re-review

pierDipi (approved):
Thanks @VikramVuppla !

/lgtm
/approve
/retest-required

PR #2797, opened 2022-11-21
Decouple autoscaling annotations from KEDA

pierDipi:
Our annotations expose KEDA, I think we should avoid doing that, so
I'm removing keda from the prefix of our annotations.

In addition, it is recommended that Knative annotations follow the kebab-case
convention instead of camel case


aavarghese (approved):
ah great - thanks @pierDipi !! All changes look appropriate.

/lgtm

PR #2811, opened 2022-11-28
Configurable default values for autoscaler

### LINKED ISSUE #2804: Add a ConfigMap `config-kafka-autoscaler` to configure autoscaler default values
**Problem**

Autoscaler default values are in https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/pkg/autoscaler/autoscaler_api.go, however, as administrator there is no way for me to override those defaults.

Proposed ConfigMap structure:

```
defaults.min-scale: <value>
defaults.max-scale: <value>
defaults.polling-interval: <value>
defaults.cooldown-period: <value>
defaults.lag-threshold: <value>
defaults.activation-lag-threshold: <value>
```

We should also add an example entry as usual like: https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/data-plane/config/broker/100-config-kafka-broker-data-plane.yaml#L24 to explain what each config does

Operator

**Exit Criteria**

**Time Estimate (optional):** 
1

/area api

aavarghese:
Fixes #2804

## Proposed Changes

- New configmap for specifying [default values](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/pkg/autoscaler/autoscaler_api.go#L52-L63) for Autoscaling Kafka resources

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/config/eventing-kafka-broker/200-controller/100-config-kafka-autoscaler.yaml
+++ b/control-plane/config/eventing-kafka-broker/200-controller/100-config-kafka-autoscaler.yaml
@@ -1,0 +1,29 @@
+# Copyright 2022 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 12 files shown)


pierDipi (commented):
Thanks!

/lgtm
/approve

PR #2829, opened 2022-12-08
Do not block finalization if secret for external topic is gone

### LINKED ISSUE #2828: Broker deletion is stuck when the referenced secret is gone
If using a `auth.secret.ref.name` and if that referenced secret is gone (regardless how it was removed), the deletion of the broker is stuck:

```
k delete -f ./015-broker-with-config.yaml 
broker.eventing.knative.dev "my-broken-kafka-broker" deleted
configmap "my-broken-kafka-broker-config" deleted
```

Especially in cases with _external_ topics, we should be more lenient, if the secret is not present, and not block the finalization

matzew:

Fixes #2828

## Proposed Changes

- If secret for external managed topic is gone / not present, we do not block the finalization. The broker does anyways not need to perform any sorts of clean-ups on these externally managed topics

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -401,7 +401,6 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker) 
 
 	// External topics are not managed by the broker,
 	// therefore we do not delete them
-	//	_, externalTopic := isExternalTopic(broker)
 	if !externalTopic {
 
 		// I do not like the use of `_`
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -387,14 +387,21 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker) 
 		return err
 	}
 
+	_, externalTopic := isExternalTopic(broker)
 	secret, err := security.Secret(ctx, &security.MTConfigMapSecretLocator{ConfigMap: brokerConfig, UseNamespaceInConfigmap: false}, r.SecretProviderFunc())
 	if err != nil {
-		return fmt.Errorf("failed to get secret: %w", err)
+
+		if externalTopic {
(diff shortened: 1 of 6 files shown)


control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -387,14 +387,21 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker)
...
+			return nil
+		} else {
+			return fmt.Errorf("failed to get secret: %w", err)
+		}
 	}
 
 	// External topics are not managed by the broker,
 	// therefore we do not delete them
-	_, externalTopic := isExternalTopic(broker)
+	//	_, externalTopic := isExternalTopic(broker)
```
creydr, line 404: nit: remove line
```suggestion
```

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -387,14 +387,21 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker)
 		return err
 	}
 
+	_, externalTopic := isExternalTopic(broker)
 	secret, err := security.Secret(ctx, &security.MTConfigMapSecretLocator{ConfigMap: brokerConfig, UseNamespaceInConfigmap: false}, r.SecretProviderFunc())
```
creydr, line 391: maybe we can move this to the block starting on line 405, as this is only needed in case of a internal topic  :shrug:

matzew: we still have 
```
	if err := r.removeFinalizerSecret(ctx, finalizerSecret(broker), secret); err != nil {
		return err
	}
```

on 448+

creydr: ah ok

control-plane/pkg/reconciler/broker/broker.go
matzew: yeah, saw that when looking at 405 :-)

PR #2853, opened 2022-12-20
Fix finalization problem when broker has incorrect topic config

### LINKED ISSUE #2815: Kafka Broker finalizer stuck if "unable to build topic config from configmap: error validating topic config from configmap invalid configuration"
Creating a Kafka broker with an existing ConfigMap that does not have valid configuration, .e.g accidentally pointing to a MTBroker ConfigMap from a Kafka-classed Broker, leads to a Broker that cannot be deleted - the finalizer becomes stuck

to reproduce:

Config:
```yaml
apiVersion: v1
data:
  default-br-config: |
    clusterDefault:
      brokerClass: MTChannelBasedBroker
      apiVersion: v1
      kind: ConfigMap
      name: config-br-default-channel
      namespace: knative-eventing
      delivery:
        retry: 10
        backoffPolicy: exponential
        backoffDelay: PT0.2S
kind: ConfigMap
```

Create a broker like:

```yaml
kind: Broker
metadata:
  name: broker
  annotations:
```

It is stuck as:

```bash
k get brokers.eventing.knative.dev -A 
NAMESPACE   NAME     URL   AGE     READY   REASON
default     broker         3m22s   False   unable to build topic config from configmap: error validating topic config from configmap invalid configuration - numPartitions: 0 - replicationFactor: 0 - bootstrapServers: [] - ConfigMap data: map[channel-template-spec:apiVersion: messaging.knative.dev/v1...
```

Deletion of that above broker is stuck at:

```
broker.eventing.knative.dev "broker" deleted
```

matzew:

Fixes #2815

## Proposed Changes

- If an invalid topic config is present for a broker we do not block finalization

**Release Note**

```release-note
If an invalid topic config is present for a broker we do not block finalization
```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -360,7 +360,15 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker) 
 	if !externalTopic {
 		topicConfig, err := r.topicConfig(logger, broker, brokerConfig)
 		if err != nil {
-			return fmt.Errorf("failed to resolve broker config: %w", err)
+
+			// On finalize, we fail to get a valid config, we can safely ignore and return nil
+			// no further actions are needed since we are also not putting the finalizer on given secret
+			// if we are not having a valid topic config
+			if strings.Contains(err.Error(), "validating topic config") {
+				return nil
+			} else {
+				return fmt.Errorf("failed to resolve broker config: %w", err)
+			}
 		}
 
 		// get security option for Sarama with secret info in it
--- a/control-plane/pkg/reconciler/broker/broker_test.go
+++ b/control-plane/pkg/reconciler/broker/broker_test.go
@@ -2201,7 +2201,50 @@ func brokerFinalization(t *testing.T, format string, env config.Env) {
 				testProber: probertesting.MockProber(prober.StatusNotReady),
 			},
 		},
-
+		{
+			Name: "Reconciled normal - with broken config",
+			Objects: []runtime.Object{
+				NewDeletedBroker(
+					WithBrokerConfig(KReference(BogusBrokerConfig())),
+				),
+				BogusBrokerConfig(),
+				NewConfigMapFromContract(&contract.Contract{
+					Resources: []*contract.Resource{
+						{
+							Uid:     BrokerUUID,
+							Topics:  []string{BrokerTopic()},
+							Ingress: &contract.Ingress{Path: receiver.Path(BrokerNamespace, BrokerName)},
+						},
+					},
+					Generation: 1,
+				}, env.DataPlaneConfigMapNamespace, env.ContractConfigMapName, env.ContractConfigMapFormat),
+				NewService(),
+				BrokerReceiverPod(env.SystemNamespace, map[string]string{
+					"annotation_to_preserve": "value_to_preserve",
+				}),
+				BrokerDispatcherPod(env.SystemNamespace, map[string]string{
+					"annotation_to_preserve": "value_to_preserve",
+				}),
+			},
+			Key: testKey,
+			WantUpdates: []clientgotesting.UpdateActionImpl{
+				ConfigMapUpdate(env.DataPlaneConfigMapNamespace, env.ContractConfigMapName, env.ContractConfigMapFormat, &contract.Contract{
+					Resources:  []*contract.Resource{},
+					Generation: 2,
+				}),
+				BrokerReceiverPodUpdate(env.SystemNamespace, map[string]string{
+					base.VolumeGenerationAnnotationKey: "2",
+					"annotation_to_preserve":           "value_to_preserve",
+				}),
+				BrokerDispatcherPodUpdate(env.SystemNamespace, map[string]string{
+					base.VolumeGenerationAnnotationKey: "2",
+					"annotation_to_preserve":           "value_to_preserve",
+				}),
+			},
+			OtherTestData: map[string]interface{}{
+				testProber: probertesting.MockProber(prober.StatusNotReady),
+			},
+		},
 		{
 			Name: "Failed to delete topic",
 			Objects: []runtime.Object{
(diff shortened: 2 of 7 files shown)


control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -407,7 +407,13 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker)
 		// TODO: refactor
 		topicConfig, _, err := r.topicConfig(logger, broker)
 		if err != nil {
-			return fmt.Errorf("failed to resolve broker config: %w", err)
+
+			// On finalize, we fail to get a valid config, we can safely ignore and return nil
+			if strings.Contains(err.Error(), "validating topic config") {
+				return nil
```
aliok, line 413: I think in this case we shouldn't return immediately but remove the finalizer on the secret.

The secret is reachable (compared to https://github.com/knative-sandbox/eventing-kafka-broker/pull/2829) and we are able to remove the finalizer.

In https://github.com/knative-sandbox/eventing-kafka-broker/pull/2829, the secret was not reachable and we were not able to remove the finalizer and it was ok to just return immediately.

matzew: you mean, the configmap, containing bogus stuff, instead of real kafka config (we are not touching secrets here)

matzew:
```
    broker.go:140: admission webhook "validation.webhook.kafka.eventing.knative.dev" denied the request: validation failed: invalid value: knative-eventing: spec.config.namespace
        Expected ConfigMap in same namespace with broker resource
    --- FAIL: TestBrokerWithBogusConfig/delete_broker_with_non_existing_Secret/Setup/install_broker (0.13s)
```

This, is an interesting issue w/ the "namespaced" broker /cc @aliok

control-plane/pkg/reconciler/broker/broker.go
aliok: So, I think this code in this line should still run: https://github.com/knative-sandbox/eventing-kafka-broker/blob/9e166c9458ed7502c686b35fb541bd9d06ca0901/control-plane/pkg/reconciler/broker/broker.go#L453

```
r.removeFinalizerSecret(ctx, finalizerSecret(broker), secret)
```


That is, as you said, when the topic config in the configmap is incorrect.

However, there's nothing preventing us removing the finalizer from the secret.

aliok:
> This, is an interesting issue w/ the "namespaced" broker /cc @aliok

The webhook allows the admission for the namespaced broker:
- Config namespace is empty (assumed same as the broker namespace)
- Config namespace is same as broker namespace

See https://github.com/knative-sandbox/eventing-kafka-broker/blob/c42f668edd46227a848eb49f777d1fa087300592/control-plane/pkg/apis/eventing/v1/broker_validation.go#L85

To fix the issue in the test:
- Remove the `configNamespace("knative-eventing")` call.
- Create the configmap in broker's namespace advance
- Pass the configmap's name: `broker.WithConfig(configName),`

So, something like:
```
f.Setup("create broker config", brokerconfigmap.Install(
		configName,
		brokerconfigmap.WithBootstrapServer(testpkg.BootstrapServersPlaintext),
		//... rest of the incorrect topic config,
	))

f.Setup("install broker", broker.Install(
		brokerName,
		append(
			broker.WithEnvConfig(),
			broker.WithConfig(configName),
			broker.WithAnnotations(
				map[string]interface{}{
					kafkabroker.ExternalTopicAnnotation: "my-topic",
				}))...))
```

Or, if you want to test the case that the configmap doesn't exist, then:
```
f.Setup("install broker", broker.Install(
		brokerName,
		append(
			broker.WithEnvConfig(),
			broker.WithConfig("inexistent-configmap"),
			broker.WithAnnotations(
				map[string]interface{}{
					kafkabroker.ExternalTopicAnnotation: "my-topic",
				}))...))
```

matzew:
It is really not about "non existing" bits. it's about bogus config, when finalizing.

I added a new test: https://github.com/knative-sandbox/eventing-kafka-broker/pull/2853/commits/5959d757813afb7baa16655d6a28ecc2254fabc6 

Also, there is no secret, no finalizer on the CM

matzew:
There could be still a secret, with the bogus/invalid config, and we than need to tear that down as well.
So: I will add more test for this - but in 2023

/hold

control-plane/pkg/reconciler/broker/broker.go
matzew: @aliok I updated a comment: basically w/ invalid config a secret is not touched etc.

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -360,7 +360,15 @@ func (r *Reconciler) finalizeKind(ctx context.Context, broker *eventing.Broker)
 	if !externalTopic {
 		topicConfig, err := r.topicConfig(logger, broker, brokerConfig)
 		if err != nil {
-			return fmt.Errorf("failed to resolve broker config: %w", err)
+
+			// On finalize, we fail to get a valid config, we can safely ignore and return nil
+			// no further actions are needed since we are also not putting the finalizer on given secret
+			// if we are not having a valid topic config
+			if strings.Contains(err.Error(), "validating topic config") {
+				return nil
```
aliok, line 368: @matzew I don't think this commit addresses the discussion we had in 2022.

See https://github.com/knative-sandbox/eventing-kafka-broker/pull/2853#issuecomment-1363998260.

Basically, here, when we return nil, we keep the finalizer string still kept on the secret because line 399 (`removeFinalizerSecret`) is never called. However, we should remove the finalizer string on the secret, even when topic config is wrong.

matzew: There is no finalizer on a secret, if we have "invalid topic config".

I guess I can update the test to have some secret there

aliok: On the other hand, the finalizer is not set at all on the secret in `reconcileKind`, if the topicConfig is wrong...

/lgtm

PR #2854, opened 2022-12-21
Templating support in namespaced broker additional resources

aliok:
Fixes #

## Proposed Changes

- Templating support in namespaced broker additional resources
- We need to *inject* the namespace into the spec of the resources created

Sample resource to put in additional resources CM is following. The `{{.Namespace}}` variable will be changed to the broker namespace.
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
...
spec:
  endpoints:
    - tlsConfig:
        ca: {}
        caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
        cert: {}
        serverName: "kafka-broker-receiver-sm-service.{{.Namespace}}.svc"
  namespaceSelector:
    matchNames:
      - "{{.Namespace}}"
...
```

**Release Note**

```release-note
Added templating support in namespaced broker additional resources, which allows injecting information to those resources.
```

**Docs**

### DIFF
--- a/control-plane/pkg/propagator/resource_propagator.go
+++ b/control-plane/pkg/propagator/resource_propagator.go
@@ -17,7 +17,9 @@
 package propagator
 
 import (
+	"bytes"
 	"fmt"
+	"text/template"
 
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


aliok:
/hold

I am getting `panic: cannot deep copy int` and I don't get it

aliok:
/unhold

Ok, issue I mentioned turned out to be OK

aliok:
/hold

Found the root cause of the panic. It is because we use the standard YAML package and not "sigs.k8s.io/yaml". Standard package causes issues when marshalling/unmarshalling YAML as it uses `int` for integers, which is ambiguous. `runtime.DeepCopy` functions expect `int64` though.

aliok:
/unhold

Now we're using the correct YAML lib

aliok:
Rebased

matzew (approved):
/lgtm
/approve

PR #2856, opened 2022-12-21
Granting permissions to delete consumer groups metadata

matzew:

## Proposed Changes

- Granting permissions to delete consumer groups metadata which is (and actually already was) needed to delete consumer groups. (something we did downstream as well)

**Release Note**

```release-note

```

**Docs**


aliok (approved):
/lgtm
/approve

PR #2872, opened 2023-01-09
Improve ConsumerGroup finalization when missing resources

pierDipi:
When the secret is gone we successfully delete the KafkaSource but ConsumerGroup might be stuck in deletion since we can't connect to Kafka anymore.

Fixes #

## Proposed Changes

- Improve ConsumerGroup finalization when missing resources

**Release Note**

```release-note
Improve ConsumerGroup finalization when missing resources
```

**Docs**

### DIFF
--- a/control-plane/pkg/counter/counter.go
+++ b/control-plane/pkg/counter/counter.go
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2023 The Knative Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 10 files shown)


control-plane/pkg/counter/counter.go
```diff
@@ -0,0 +1,49 @@
...
+
+type Counter struct {
+	counterLock sync.RWMutex
+	counterMap  map[string]int
+
+	cache prober.Cache
+}
+
+func NewExpiringCounter(ctx context.Context) *Counter {
+	cache := prober.NewLocalExpiringCache(ctx, 10*time.Minute)
```
matzew, line 35: shuld we (eventually) extract that cache? I guess might be handy for a lot more

pierDipi: Yes, any suggestion on where to place it?

aliok:
/retest

Infra

aliok:
/lgtm
/approve

I like the idea

PR #2887, opened 2023-01-12
Fix Consumer resource finalization

pierDipi:
When we use the DeleteResource function we check
for changed == coreconfig.ResourceChanged to then
proceed to update the CM, however, by returning idx, 
changed == coreconfig.ResourceChanged is almost never true, 
so we never update the CM with the resource removed.


matzew (approved):
/lgtm
/approve

PR #2919, opened 2023-01-25
Improve source reconciler KEDA enabled check 

pierDipi:
KEDA enabled check for the source controller was using a
subset of the expected conditions, so in this PR we're
moving to use a common function to check whether KEDA is
enabled or not

- Add unit tests
- Add E2E tests for KafkaSource manual scaling


control-plane/pkg/autoscaler/keda/keda.go
```diff
@@ -260,13 +265,28 @@ func setAnnotation(objAnnotations map[string]string, key string, cgAnnotations m
 	cgAnnotations[key] = value
 }
 
-func IsEnabled(cg *kafkainternals.ConsumerGroup) bool {
-	v, ok := cg.GetAnnotations()[autoscaler.AutoscalingClassAnnotation]
+func IsEnabled(ctx context.Context, features *config.KafkaFeatureFlags, client versioned.Interface, object metav1.Object) bool {
+	if !features.IsControllerAutoscalerEnabled() {
```
matzew, line 269: why?

matzew:
/lgtm

/hold

As a follow up, it would be nice to generally have some better doc on the "new" source, and its configs, including `config-kafka-features`

control-plane/pkg/autoscaler/keda/keda.go
matzew: nevermind

matzew (approved):
/lgtm
/approve

PR #2920, opened 2023-01-25
Fix tests + upgrade dependencies

pierDipi:
Adopt tests to https://github.com/knative-sandbox/reconciler-test/pull/394

Pull forward dependencies, including a breaking change in `eventing`, and fix breakages.


psschwei (commented):
/lgtm

PR #2924, opened 2023-01-26
Enforce owner references on contract ConfigMap creation

pierDipi:
When contract configmaps are created, we don't enforce
owner references to be set when they are create which
could lead to leftovers ConfigMaps when:

- KafkaNamespaced brokers are deleted
- kafka-source-dispatcher pods are deleted (for source v2)

### DIFF
--- a/control-plane/pkg/reconciler/base/reconciler.go
+++ b/control-plane/pkg/reconciler/base/reconciler.go
@@ -153,6 +153,10 @@ func (r *Reconciler) createDataPlaneConfigMap(ctx context.Context) (*corev1.Conf
 		},
 	}
 
+	if r.DataPlaneConfigMapTransformer != nil {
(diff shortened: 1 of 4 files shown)


control-plane/pkg/reconciler/base/reconciler.go
```diff
@@ -153,6 +153,10 @@ func (r *Reconciler) createDataPlaneConfigMap(ctx context.Context) (*corev1.Conf
 		},
 	}
 
+	if r.DataPlaneConfigMapTransformer != nil {
+		r.DataPlaneConfigMapTransformer(cm)
+	}
```
aliok, line 158: Good catch!

We can probably just delete the same call in line 137 as that sets the labels on in-memory resources, but let's keep it for stability. Maybe there's some code which persists that in-memory resources.

aliok (approved):
/lgtm
/approve

PR #2932, opened 2023-01-31
Migrate `TestBrokerTriggersSink` broker e2e test to new reconciler-test framework 

creydr:
PR to migrate the `TestBrokerTriggersSink` broker e2e test to new reconciler-test framework.

## Proposed Changes
- :broom: Migrate `TestBrokerTriggersSink` (0dbf48180916a09a6f4497bfa54c975644fa102c)

**Hits for reviewers:**
PR is currently WIP as:
* requires knative-sandbox/reconciler-test#452 to be merged (got around it temporarily by bfafd744f5c25df1f44cadbdf278838880f9d2ae)


test/e2e_new/broker_trigger_sink_test.go
```diff
@@ -0,0 +1,157 @@
...
+	sink := feature.MakeRandomK8sName("sink")
+	verifyMessagesJobName := feature.MakeRandomK8sName("verify-messages-job")
+
+	topic := kafka.BrokerTopic(brokerreconciler.TopicPrefix, &eventingv1.Broker{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      brokerName,
+			Namespace: env.Namespace(),
+		},
+	})
+
```
pierDipi, line 87: We usually separate feature and test because it can useful to reuse them, even if in this case it doesn't seem useful but I'd do that for consistency.

test/e2e_new/broker_trigger_sink_test.go
```diff
@@ -0,0 +1,157 @@
...
+	f.Requirement("Send event for trigger 2", eventshub.Install(feature.MakeRandomK8sName("sender-2"),
+		eventshub.InputEvent(eventTrigger2),
+		eventshub.StartSenderToResource(broker.GVR(), brokerName)))
+
+	f.Requirement("Install verify-messages job", job.Install(verifyMessagesJobName, kafkaConsumerImage,
+		job.WithEnvs(map[string]string{
+			"BOOTSTRAP_SERVERS": testpkg.BootstrapServersPlaintext,
+			"TOPIC":             topic,
+			"IDS":               fmt.Sprintf("%s,%s", eventIdTrigger1, eventIdTrigger2),
+			"CONTENT_MODE":      "structured",
```
pierDipi, line 149: I just noticed this and affects the existing test as well, so we can ignore it for the purpose of this PR.

Since we are not setting the content mode when we install the sink [1] I would expect this to be "binary" since that's the default content mode?

https://github.com/knative-sandbox/eventing-kafka-broker/blob/9459144e95bd13df164340970a96b58a2761daa4/control-plane/pkg/apis/eventing/v1alpha1/kafka_sink_defaults.go#L28-L32

so:
- that verification is not happening, or
- there is a bug in the sink implementation for changing content mode

The existing test is doing the same thing and it passes as well

[1]
```golang
	f.Setup("Install kafkasink", kafkasink.Install(sink, topic, testpkg.BootstrapServersPlaintextArr,
		kafkasink.WithNumPartitions(10),
		kafkasink.WithReplicationFactor(1)))
```

test/e2e_new/broker_trigger_sink_test.go
creydr: Yeah I didn't see a reason for reusing it too. Anyhow I updated it, so we have consistency :+1:

pierDipi (approved):
/lgtm
/approve

PR #2937, opened 2023-02-03
Remove unused dependency to eventing event-library test-image

creydr:
The event-library test image from eventing is not used in EKB anymore (#2920) and thus the dependency can be removed.


pierDipi (approved):
/lgtm
/approve

PR #2965, opened 2023-02-15
Remove mention for kind.local in Development docs

pierDipi:
Jib doesn't support `kind.local`


matzew (approved):
/lgtm
/approve

good point, to not confuse contributors

PR #2946, opened 2023-02-08
Migrate broker SASL/SSL e2e tests to new rekt test framework

creydr:
Another PR of the series to migrate the e2e tests in `test/e2e_broker` to the new reconciler-test framework.

## Proposed Changes
- :broom: Migrate `TestBrokerAuthPlaintext` test
- :broom: Migrate `TestBrokerAuthSsl` test
- :broom: Migrate `TestBrokerAuthSaslPlaintextScram512` test
- :broom: Migrate `TestBrokerAuthSslSaslScram512` test

*Hints for reviewers:*
Tests for missing auth secret are added in #2968

PR #2968, opened 2023-02-16
Mark broker as not ready when auth secret is missing

creydr:
When we create a Kafka Broker with SASL/SSL configuration, but with having the auth secret missing, the broker is in state UNKNOWN instead of not-ready:

```
NAME              URL   AGE   READY     REASON
broker-btgvxywu         9s    Unknown
```

This PR addresses it and sets the Ready condition to false, in case the auth secret is missing.

## Proposed Changes

- :bug: Set Broker to NotReady state when SASL/SSL auth secret is missing

**Release Note**
```release-note

```

/hold as this builds on top of #2946 to use some rekt test features


control-plane/pkg/reconciler/base/receiver_condition_set.go
```diff
@@ -183,6 +183,18 @@ func (manager *StatusConditionManager) TopicReady(topic string) {
 	)
 }
 
+func (manager *StatusConditionManager) FailedToGetBrokerAuthSecret(err error) reconciler.Event {
+
+	manager.Object.GetConditionSet().Manage(manager.Object.GetStatus()).MarkFalse(
+		apis.ConditionReady,
```
pierDipi, line 189: We shouldn't use the ready condition directly but through other conditions, the ready condition will be marked true only when all other conditions are true, so we can reuse the `ConditionTopicReady` in this case.

creydr: Ah ok. Didn't know that. I updated it...

PR #2979, opened 2023-02-22
Do not inject namespace to additional resources using Manifestival's namespace injection

aliok:
In EKB controller, we use Manifestival to create resources that are in "additional resources". When we have the RoleBinding like this in the Manifest:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: knative-prometheus-k8s
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: knative-prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: openshift-monitoring           # --> changed
```

will become:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: knative-prometheus-k8s
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: knative-prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: <user-namespace>        # --> changed
```

This is done here https://github.com/manifestival/manifestival/blob/82196bd303e23c641fda3d002c0663b892092e9c/transform.go#L66-L73 

To prevent this, we don't run the Manifestival's namespace injection on "additional resources". We can inject namespaces to them with templating anyway.

### DIFF
--- a/control-plane/pkg/reconciler/broker/namespaced_broker.go
+++ b/control-plane/pkg/reconciler/broker/namespaced_broker.go
@@ -38,7 +38,6 @@ import (
 	rbaclisters "k8s.io/client-go/listers/rbac/v1"
 	"k8s.io/utils/pointer"
 
-	"knative.dev/pkg/controller"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/reconciler"
 	"knative.dev/pkg/resolver"
@@ -97,21 +96,6 @@ func (r *NamespacedReconciler) ReconcileKind(ctx context.Context, broker *eventi
 	}
 
 	br := r.createReconcilerForBrokerInstance(broker)
-
-	statusConditionManager := base.StatusConditionManager{
-		Object:     broker,
-		SetAddress: broker.Status.SetAddress,
-		Env:        r.Env,
-		Recorder:   controller.GetEventRecorder(ctx),
-	}
-
-	// Get contract config map. Do this in advance, otherwise
-	// the dataplane pods that need volume mounts to the contract configmap
-	// will get stuck and will never be ready.
-	_, err := br.GetOrCreateDataPlaneConfigMap(ctx)
-	if err != nil {
-		return statusConditionManager.FailedToGetConfigMap(err)
-	}
 
 	event := r.reconcileDataPlane(ctx, broker)
 	if event != nil {
--- a/control-plane/pkg/reconciler/broker/namespaced_broker.go
+++ b/control-plane/pkg/reconciler/broker/namespaced_broker.go
@@ -38,6 +38,7 @@ import (
 	rbaclisters "k8s.io/client-go/listers/rbac/v1"
 	"k8s.io/utils/pointer"
 
+	"knative.dev/pkg/controller"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/reconciler"
 	"knative.dev/pkg/resolver"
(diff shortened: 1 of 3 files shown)


control-plane/pkg/reconciler/broker/namespaced_broker.go
```diff
@@ -97,6 +98,21 @@ func (r *NamespacedReconciler) ReconcileKind(ctx context.Context, broker *eventi
...
+		Recorder:   controller.GetEventRecorder(ctx),
+	}
+
+	// Get contract config map. Do this in advance, otherwise
+	// the dataplane pods that need volume mounts to the contract configmap
+	// will get stuck and will never be ready.
+	_, err := br.GetOrCreateDataPlaneConfigMap(ctx)
+	if err != nil {
+		return statusConditionManager.FailedToGetConfigMap(err)
+	}
```
pierDipi, line 114: We don't need this since it will be created once we run `ReconcileKind`

pierDipi: > // Get contract config map. Do this in advance, otherwise
   // the dataplane pods that need volume mounts to the contract configmap
   // will get stuck and will never be ready.

Or to put it differently, can you show me why this is true?

aliok: If we don't get this configmap created before the dataplane is created by `br.ReconcileKind()`, then dataplane pods are stuck waiting for the configmap.

That's why I do this call in advance. It is idempotent anyway.

pierDipi: They are not stuck because the CM is eventually created after a few miliseconds, so a warning event is not the same as being stuck

aliok: No, they're actually stuck. I've seen them being stuck.

BTW, I updated my comment above https://github.com/knative-sandbox/eventing-kafka-broker/pull/2979#discussion_r1121395121 I was describing it wrong.

pierDipi: That's not possible, it would work in CI currently on main

pierDipi: What did you see when you say it was stuck? Can you describe how to reproduce the behavior?

aliok: Hm, maybe they were stuck "only" because of a missing secret then... (which I solved later)

Let me check.

aliok: @pierDipi You are right!

I removed this code now

aliok:


/test unit-tests
--


<br class="Apple-interchange-newline">

aliok:
```
Failed Pod (data in YAML format) - chaosduck-ff4675f47-wmjx6
```

matzew:
:duck:

control-plane/pkg/reconciler/broker/namespaced_broker.go
```diff
@@ -624,6 +652,9 @@ func filterMetadata(u *unstructured.Unstructured) error {
 
 func setLabel(u *unstructured.Unstructured) error {
 	labels := u.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
```
pierDipi, line 656: nit
```suggestion
		labels = make(map[string]string, 1)
```

pierDipi (approved):
/lgtm
/approve

PR #2981, opened 2023-02-23
Fix bash scripts import paths

pierDipi:
`pwd` or `dirname $0` is not reliable way of getting the repo root
dir, so I'm using `$(dirname "$(realpath "${BASH_SOURCE[0]}")")`.

Tested this in eventing-istio.

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -19,10 +19,12 @@ readonly LOCAL_DEVELOPMENT=${LOCAL_DEVELOPMENT:-false}
 export REPLICAS=${REPLICAS:-3}
 export KO_FLAGS="${KO_FLAGS:-}"
 
-source $(pwd)/vendor/knative.dev/hack/e2e-tests.sh
-source $(pwd)/hack/data-plane.sh
-source $(pwd)/hack/control-plane.sh
-source $(pwd)/hack/artifacts-env.sh
+repo_root_dir=$(dirname "$(realpath "${BASH_SOURCE[0]}")")/..
+
+source "${repo_root_dir}"/vendor/knative.dev/hack/e2e-tests.sh
+source "${repo_root_dir}"/hack/data-plane.sh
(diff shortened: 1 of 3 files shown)


aliok (approved):
LGTM, if tests pass 😄

test/e2e-common.sh
```diff
@@ -19,10 +19,12 @@ readonly LOCAL_DEVELOPMENT=${LOCAL_DEVELOPMENT:-false}
...
-source $(pwd)/vendor/knative.dev/hack/e2e-tests.sh
-source $(pwd)/hack/data-plane.sh
-source $(pwd)/hack/control-plane.sh
-source $(pwd)/hack/artifacts-env.sh
+repo_root_dir=$(dirname "$(realpath "${BASH_SOURCE[0]}")")/..
+
+source "${repo_root_dir}"/vendor/knative.dev/hack/e2e-tests.sh
+source "${repo_root_dir}"/hack/data-plane.sh
+source "${repo_root_dir}"/hack/control-plane.sh
+source "${repo_root_dir}"/hack/artifacts-env.sh
```
aliok, line 27: I am not 100% sure but, can we create a function for this?

Like I call it this way:

```
source_res "../vendor/knative.dev/hack/e2e-tests.sh"
source_res "../hack/data-plane.sh"
```

Or for `source "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/label.sh`

```
source_res "/label.sh"
```

pierDipi: I see, where would this function live and how would that understand the directory that the caller is in?

pierDipi: Nevermind, I see how to solve them

pierDipi: but this should become:
```
source_res "/vendor/knative.dev/hack/e2e-tests.sh"
source_res "/hack/data-plane.sh"
```

basically the input directory should be based on the root of the repository

PR #2982, opened 2023-02-23
Added nil as return value for rebuildCMFromStatusAnnotations

### LINKED ISSUE #2577: RebuildCMFromAnnotations should only return a configmap if there are values on the status annotations
**Describe the bug**
the `rebuildCMFromAnnotations` function always returns a `ConfigMap`, regardless if there are values on the `broker.status.annotations` map or not, see:

https://github.com/knative-sandbox/eventing-kafka-broker/blob/d6614b3aa02ff48f983193840485d321368201ef/control-plane/pkg/reconciler/broker/broker.go#L484

 **Expected behavior**
Return only the `ConfigMap` when we have actual data to return

Vishal-Chdhry:
Signed-off-by: Vishal Choudhary <sendtovishalchoudhary@gmail.com>

Fixes #2577 

## Proposed Changes
- :broom: Update or clean up current behavior
- RebuildCMFromAnnotations should only return a configmap if there are values on the status annotations

**Release Note**

```release-note
Added nil as return value for rebuildCMFromStatusAnnotations if there are values on the status annotations
```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -504,7 +504,7 @@ func (r *Reconciler) brokerConfigMap(logger *zap.Logger, broker *eventing.Broker
 		return cm, fmt.Errorf("failed to get configmap %s/%s: %w", namespace, broker.Spec.Config.Name, getCmError)
 	}
 	if apierrors.IsNotFound(getCmError) {
-		// will at least return an empty CM
+		// will return nil if there are no status annotations
 		cm = rebuildCMFromStatusAnnotations(broker)
 		if cm == nil {
 			return cm, fmt.Errorf("failed to get configmap %s/%s: %w", namespace, broker.Spec.Config.Name, getCmError)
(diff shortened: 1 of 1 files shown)


Vishal-Chdhry:
@matzew Is this the required change?

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -545,6 +548,9 @@ func rebuildCMFromStatusAnnotations(br *eventing.Broker) *corev1.ConfigMap {
 			Name:      br.Spec.Config.Name,
 		},
 	}
+	if len(br.Status.Annotations) == 0 {
```
matzew, line 551: if we do this, we should move the check _before_ we create the CM.

matzew:
A couple of things.

Eventually this all goes down into
https://github.com/knative-sandbox/eventing-kafka-broker/blame/main/control-plane/pkg/reconciler/broker/broker.go#L515

and this _assumes_ we do have a `cm` object (no nil checks)

The code is complicated, yes - hence we did not really change a lot there.

A couple of weeks ago I did some cosmetic changes, very carefully - and I'd potentially not update this, if not needed

Let's perhaps close the referenced issue, as not relevant - sorry, Ihave not seen the ping from @pierDipi on the issue

/hold

matzew:
/ok-to-test

Just to check, but I think/expect that tests will fail for the change

matzew:
@Vishal-Chdhry can we close this? 
I do not see that much value here - also the code is a bit coupled.

Feel free to close this PR and the referenced issue :pray:

Vishal-Chdhry:
@matzew Yes we should close this issue, as i am not a member i cannot do it, can you please close the issue. I will close this pr

PR #2985, opened 2023-02-27
Replace the old slack channel link with the new proper one

debasishbsws:
Though the slack channel is now under CNCF slack it was not updated here and still pointing to the old slack channel.


pierDipi (approved):
/lgtm
/approve

PR #2987, opened 2023-02-28
Remove unneeded event-flaker test image import

creydr:
event-flaker image was removed in https://github.com/knative/eventing/pull/6750


pierDipi (approved):
/lgtm

PR #2990, opened 2023-02-28
Remove skipped `BrokerWithManyTriggers` test

creydr:
Remove the skipped `BrokerWithManyTriggers` test

/kind cleanup


test/e2e_broker/broker_with_many_triggers_test.go
```diff
@@ -1,35 +0,0 @@
...
-	"context"
-	"testing"
-
-	"knative.dev/eventing/test/e2e/helpers"
-
-	testbroker "knative.dev/eventing-kafka-broker/test/pkg/broker"
-)
-
-func TestBrokerWithManyTriggers(t *testing.T) {
-	t.Skip("Pass more events than necessary (related? https://knative.slack.com/archives/C9JP909F0/p1595244377489600)")
```
matzew, line 32: I am fine with removing "legacy" e2e tests, but does this work when not being skipped?

Unfortunately it points to slack instead of GH :sweat_smile:

pierDipi:
We have added a new e2e test, rekt-based, can we add that test in place of this one?

PR #2999, opened 2023-03-03
[WIP] Use broker template dir in rekt tests

creydr:
Currently most of the broker rekt tests, define the same broker config. This PR addresses it and uses a broker template with some predefined default config (which still allows to specify the whole config).

## Proposed Changes

- :gift: Use the broker template dir in the rekt tests so we don't have to specify the same config each time and can rely more on defaults of the broker config.

WIP as I still need to run the tests and cleanup the tests from the redundant config declarations.

### DIFF
--- a/test/reconciler-tests.sh
+++ b/test/reconciler-tests.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 source $(dirname $0)/e2e-common.sh
+export BROKER_TEMPLATES=./templates/kafka-broker
 
 if ! ${SKIP_INITIALIZE}; then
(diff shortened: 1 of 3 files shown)


creydr:
/assign @matzew @pierDipi @aliok 
WDYT?

test/reconciler-tests.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 source $(dirname $0)/e2e-common.sh
+export BROKER_TEMPLATES=./templates/kafka-broker
```
pierDipi, line 4: Can we put this in e2e-common.sh?

pierDipi (approved):
/lgtm
/approve

PR #3006, opened 2023-03-11
Fix: Experemental filters `Any`Filter working bug

### LINKED ISSUE #2942: Experemental filters does not work
**Describe the bug**
Hello! It seems to me that experemental filters work not all.

 **Expected behavior**
Seems to me it should work, but it does not
```
  filters:
    - and:
        - exact:
             type: apple
        - exact:
             type: banana
```
also cannot get ce filters worked
filters:
   - cesql: type = 'bananar'
   
Seems to me there is a type ("==" instead of "=") in ce docs  https://knative.dev/docs/eventing/experimental-features/new-trigger-filters/#conflict-with-the-current-filter-field
   
 **To Reproduce**
Switch experemental filters on
Try to make filters above

 **Knative release version**
Knative Serving 1.8.3
networking layer (kourier) 1.8.1
Serving extensions 1.8.3
Eventing 1.8.5
Apache Kafka Channel 1.8.6
Broker layer 1.8.6
Source controller 1.8.6
 **Additional context**
More info here: 
https://github.com/knative/eventing/discussions/6717

Thank you !

debasishbsws:
Fixes #2942

instead of passing `f.any` filter as argument in `newAnyFilter()` it was passing `f.all` which was causing this issue

## Proposed Changes

- passing required arguments to create a proper `Any` filter.
- put the test to check if `any` filter gets its format or not

### DIFF
--- a/control-plane/pkg/contract/subscriptionsapi.go
+++ b/control-plane/pkg/contract/subscriptionsapi.go
@@ -96,7 +96,7 @@ func newAllFilter(filters []v1.SubscriptionsAPIFilter) *DialectedFilter {
 // newAnyFilter converts the SubscriptionsAPIFilter into the any dialect of the
 // DialectedFilter as defined in CloudEvents Subscriptions API.
 //
-// All filter evaluates to true when all nested filters evaluate to true.
+// Any filter evaluates to true when at least one nested filters evaluate to true.
 //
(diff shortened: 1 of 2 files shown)


pierDipi (approved):
Thanks and good catch!

/lgtm
/approve

PR #3008, opened 2023-03-14
Conformance tests against MTChannelBasedBroker

mgencur:
It is useful to allow for running REKT conformance tests with a different implementation as well. For example, MTChannelBasedBroker backed by KafkaChannel which was developed in knative-sandbox/eventing-kafka.

With this PR, it is possible to run them by:
```
export BROKER_CLASS="MTChannelBasedBroker"
go test -tags=e2e -timeout=100m -run TestBrokerConformance ./test/e2e_new
```

## Proposed Changes

- Remove the restriction to run tests only with Kafka and NamespacedKafka broker class.
- Provide additional configmap element for (channel-template-spec) when the type is MTChannelBasedBroker

**Release Note**

```release-note

```

**Docs**


test/rekt/resources/configmap/broker/broker_config.yaml
```diff
@@ -30,3 +30,11 @@ data:
   {{ if .authSecret }}
   auth.secret.ref.name: "{{ .authSecret }}"
   {{ end }}
+  {{ if .kafkaChannel }}
+  channel-template-spec: |
```
matzew, line 34: in theory if we stash `KafkaChannel` here, we do not need the other values - but even if they are there, they are not causing any harm.  Just a note

mgencur: > we do not need the other values
What do you mean by "other values"? Thanks

matzew (approved):
/lgtm
/approve

matzew (approved):
/lgtm
/approve

/retest

PR #3009, opened 2023-03-16
Retry failed reconciliations on the data plane

pierDipi:
Data plane reconciliation might fail for many reasons, especially when using Istio with EKB and namespaced broker that creates data plane automatically when a broker is created the first reconciliation on startup often fails since istio proxy might not have started yet.

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerMessageHandler.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerMessageHandler.java
@@ -37,6 +37,7 @@ public class ResourcesReconcilerMessageHandler implements Handler<Message<Object
 
   public final static String ADDRESS = "resourcesreconciler.core";
   public static final int RECONCILE_TIMEOUT = 10000;
+  public static final int RECONCILE_FAILED_RETRY_DELAY = 5000;
 
   private final Vertx vertx;
   private final ResourcesReconciler resourcesReconciler;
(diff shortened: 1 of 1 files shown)


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerMessageHandler.java
```diff
@@ -109,6 +110,13 @@ private void reconcileLast(final DataPlaneContract.Contract newContract) {
           // If that's the case, reconcile it.
           if (last.get().getGeneration() != contract.getGeneration()) {
             reconcileLast(null);
+          } else if (r.failed()) {
+            // Retry failed reconciliations after a few seconds.
+            // We add a delay to avoid looping too much on errors.
+            vertx.setTimer(RECONCILE_FAILED_RETRY_DELAY, v -> {
+              logger.info("Retrying previously failed reconciliation");
+              reconcileLast(null);
```
aliok, line 118: Would this cause the reconciler to try forever every 5 seconds?

pierDipi: No, see docs for setTimer:
```
  /**
   * Set a one-shot timer to fire after {@code delay} milliseconds, at which point {@code handler} will be called with
   * the id of the timer.
   *
   * @param delay  the delay in milliseconds, after which the timer will fire
   * @param handler  the handler that will be called with the timer ID when the timer fires
   * @return the unique ID of the timer
   */
```

pierDipi: it's called only once

aliok: kk

/lgtm
/approve

PR #3023, opened 2023-03-27
Fix broken initial offset on reset on a non64 bit boundary

EndPositive:
Fixes broken initial offset on reset on a non64 bit boundary

## Proposed Changes

- Take the `relativeOffset`/`wordSize` remainder into account when resetting `initialOffset`

@hadriansecurity


WouterHadrian:
thanks @EndPositive for helping us with this issue. 

Some background info. In our production system, after a "random" time we saw that the lag on a single partition was rising and slowly extending to other partitions as well. A reboot of the dispatcher was required to get it up and running again. Starting with the last committed offset. Impact is thus significant.

We decided to dive into it and found at that the offset of that partition is not committed anymore. Each time after a partition receives roughly 1.000.000 new events. This bug (especially with out-of-order processing) can explain this. The offset administration does not correspond anymore with the actual offset.

pierDipi:
Flake
/test unit-tests

pierDipi (approved):
Thank you!

/lgtm
/approve

PR #3039, opened 2023-04-12
Update the patch

matzew:
## Proposed Changes

- Update a patch since applying fails

E.g. see here: 
https://github.com/knative/eventing/actions/runs/4673652391/jobs/8277105585


matzew:
seems not really needed after the bump we did in https://github.com/knative-sandbox/eventing-kafka-broker/pull/3041

matzew:
Looks like this is not needed anymore, and we can always revisit this, if really needed

/close

PR #3073, opened 2023-05-08
WIP: Adds tls.crt and tls.key support to read from TLS secrets

### LINKED ISSUE #2818: TLS secret should conform to Kubernetes convention (tls.crt and tls.key)
**Problem**

Kubernetes has existing conventions related to TLS secrets, specifically that secrets contain keys named `tls.crt` and `tls.key` for the public and private key respectively. The secret required for client authentication requires keys `user.crt` and `user.key`, which does not comply with this convention and makes the Kafka broker incompatible with many useful tools that do, particularly Hashicorp Vault via Cert Manager. Certificates created by Cert Manager produce keys that comply to the Kubernetes convention.

System Operator

**Exit Criteria**
Provide a way to override the name of the secret key containing the user's client certificate and private key.

**Time Estimate (optional):** 
Code effort is negligible (<1d). More effort would likely be spent on QA and contemplating an API change which is difficult to estimate.

MdSahil-oss:
Fixes #2818

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


MdSahil-oss:
@pierDipi I just added a sample. Please review it then I'll make changes other places which reads `user.crt` & `user.key`.

control-plane/pkg/security/secret.go
```diff
@@ -175,18 +175,24 @@ func sslConfig(protocol string, data map[string][]byte) kafka.ConfigOption {
...
-						`[protocol %s] required user key (key: %s) - use "%s: true" to disable client auth`,
-						protocol, UserCertificate, UserSkip,
-					)
+					userCert, ok = data["tls.crt"]
+					if !ok {
+						return fmt.Errorf(
+							`[protocol %s] required user key (key: %s) - use "%s: true" to disable client auth`,
+							protocol, UserCertificate, UserSkip,
+						)
+					}
```
pierDipi, line 195: There is a bit of repetition here: maybe we could extract a function like
```golang
func oneValueOf(data map[string][]byte, keys...string) ([]byte, bool) {
   // TODO go through keys and find the first present and non empty key in data
}
```

and then we can use it like this:
```golang
userKeyCert, ok := oneValueOf(data, UserKey, "tls.key")
/// ...
userCert, ok := oneValueOf(data, UserCertificate, "tls.crt")
```

pierDipi (commented):
This is a good start!

control-plane/pkg/security/secret.go
MdSahil-oss: Done 👍🏽

MdSahil-oss:
@pierDipi Created a function to follow DRY convention & wrote a condition in `keda.go` to check secret type `kubernetes.io/tls` or not. please review it again, I think now Golang related changes are done, Java code remains to be changed.

MdSahil-oss:
@pierDipi I think this PR is ready for final review.

control-plane/pkg/autoscaler/keda/keda.go
```diff
@@ -198,11 +198,19 @@ func GenerateTriggerAuthentication(cg *kafkainternals.ConsumerGroup, secretData
...
-			secretTargetRefs = append(secretTargetRefs, cert)
+			if secret.Type == "kubernetes.io/tls" {
+				cert := kedav1alpha1.AuthSecretTargetRef{Parameter: "cert", Name: secret.Name, Key: "tls.crt"}
+				secretTargetRefs = append(secretTargetRefs, cert)
 
-			key := kedav1alpha1.AuthSecretTargetRef{Parameter: "key", Name: secret.Name, Key: security.UserKey}
-			secretTargetRefs = append(secretTargetRefs, key)
+				key := kedav1alpha1.AuthSecretTargetRef{Parameter: "key", Name: secret.Name, Key: "tls.key"}
+				secretTargetRefs = append(secretTargetRefs, key)
+			} else {
```
pierDipi, line 207: I don't think using the `Type` to infer using the news keys is correct, we need to check the actual secret and check whether is using one key or the key

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/security/KubernetesCredentials.java
```diff
@@ -103,9 +103,12 @@ public String userCertificate() {
       return null;
     }
     if (userCertificate == null) {
-      final var keystore = secret.getData().get(USER_CERTIFICATE_KEY);
+      var keystore = secret.getData().get(USER_CERTIFICATE_KEY);
       if (keystore == null) {
-        return null;
+        keystore = secret.getData().get("tls.crt");
+        if (keystore == null) {
+          return null;
```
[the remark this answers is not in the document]
MdSahil-oss: Done! @pierDipi Please review

control-plane/pkg/security/secret.go
```diff
@@ -173,15 +173,15 @@ func sslConfig(protocol string, data map[string][]byte) kafka.ConfigOption {
 			}
 
 			if !skipClientAuth {
-				userKeyCert, ok := data[UserKey]
+				userKeyCert, ok := firstNonEmptyValueOf(data, UserKey, "tls.key")
```
pierDipi, line 176: Can we use constants instead of hardcoded strings?

MdSahil-oss: Okay!

control-plane/pkg/security/secret.go
```diff
@@ -220,3 +220,11 @@ func skipClientAuthCheck(data map[string][]byte) (bool, error) {
...
 	return enabled, nil
 }
+
+func firstNonEmptyValueOf(data map[string][]byte, keys ...string) ([]byte, bool) {
+	userKeyOrCert, ok := data[keys[0]]
+	if !ok {
+		userKeyOrCert, ok = data[keys[1]]
+	}
+	return userKeyOrCert, ok
+}
```
pierDipi, line 230: We shouldn't assume only 2 keys, so we need a for loop here

pierDipi (commented):
Can we add unit tests to verify that the patch is working as intended now and in the future?

control-plane/pkg/security/secret.go
MdSahil-oss: Done!

control-plane/pkg/autoscaler/keda/keda.go
```diff
@@ -198,11 +198,19 @@ func GenerateTriggerAuthentication(cg *kafkainternals.ConsumerGroup, secretData
...
 
-			cert := kedav1alpha1.AuthSecretTargetRef{Parameter: "cert", Name: secret.Name, Key: security.UserCertificate}
-			secretTargetRefs = append(secretTargetRefs, cert)
+			if secret.Type == "kubernetes.io/tls" {
+				cert := kedav1alpha1.AuthSecretTargetRef{Parameter: "cert", Name: secret.Name, Key: "tls.crt"}
+				secretTargetRefs = append(secretTargetRefs, cert)
 
-			key := kedav1alpha1.AuthSecretTargetRef{Parameter: "key", Name: secret.Name, Key: security.UserKey}
-			secretTargetRefs = append(secretTargetRefs, key)
+				key := kedav1alpha1.AuthSecretTargetRef{Parameter: "key", Name: secret.Name, Key: "tls.key"}
```
pierDipi, line 205: We can use the constants you just defined for the keys

MdSahil-oss: Before to use these constants I think I must resolve the review https://github.com/knative-sandbox/eventing-kafka-broker/pull/3073#discussion_r1200142275

control-plane/pkg/autoscaler/keda/keda.go
MdSahil-oss: @pierDipi Just updated the way of checking the secret, Please review it.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/security/KubernetesCredentials.java
```diff
@@ -197,4 +196,15 @@ public String SASLPassword() {
...
+  private String firstNonEmptyValueOf(String keyOrCrt, String tlsKeyOrTlsCrt) {
+    var userKeyOrUserCrt = secret.getData().get(keyOrCrt);
+    if (userKeyOrUserCrt == null) {
+      userKeyOrUserCrt = secret.getData().get(tlsKeyOrTlsCrt);
+      if (userKeyOrUserCrt == null) {
+        return null;
+      }
+    }
+    return userKeyOrUserCrt;
+  }
```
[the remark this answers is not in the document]
MdSahil-oss: Yes! Go Code's `firstNonEmptyValueOf()` function can be made more meaningful if we use limited number of variables as arguments instead of `varargs`.

control-plane/pkg/security/secret.go
pierDipi: Something like this:

```golang
func firstNonEmptyValueOf(data map[string][]byte, keys ...string) ([]byte, bool) {
	for _, key := range keys {
	    if v, ok := data[key]; ok {
	        return v, ok
	     }
	}
	return "", false
}
```

control-plane/pkg/autoscaler/keda/keda.go
MdSahil-oss: Done 👍🏽

control-plane/pkg/security/secret.go
MdSahil-oss: Done! 👍🏽 , please Review it

MdSahil-oss:
@pierDipi Added tests with review :)

control-plane/pkg/autoscaler/keda/keda.go
```diff
@@ -198,11 +198,19 @@ func GenerateTriggerAuthentication(cg *kafkainternals.ConsumerGroup, secretData
...
-			secretTargetRefs = append(secretTargetRefs, key)
+				key := kedav1alpha1.AuthSecretTargetRef{Parameter: "key", Name: secret.Name, Key: security.TlsKey}
+				secretTargetRefs = append(secretTargetRefs, key)
+			} else {
+				cert := kedav1alpha1.AuthSecretTargetRef{Parameter: "cert", Name: secret.Name, Key: security.UserCertificate}
+				secretTargetRefs = append(secretTargetRefs, cert)
+
+				key := kedav1alpha1.AuthSecretTargetRef{Parameter: "key", Name: secret.Name, Key: security.UserKey}
+				secretTargetRefs = append(secretTargetRefs, key)
+			}
```
pierDipi, line 213: The implementation in `control-plane/pkg/security/secret.go` allows setting mixed keys while this one doesn't

for example:

```
tls.key: <...>
user.crt: <...>
```
would be valid for the function `sslConfig` in `control-plane/pkg/security/secret.go` and it would use:
- the value of `tls.key` for the private key 
- `user.crt` for the public key

On the other end in this logic we would reference:
- the value of `user.key` for the private key 
- the value of `user.crt` for the public key
which is not consistent with the other file, so we need to check and set values individually and fallback to `user*` keys when not found:

Something like this (and the same for the other key):
```golang
if secret.Data[security.TlsKey] != nil {
    key := kedav1alpha1.AuthSecretTargetRef{Parameter: "key", Name: secret.Name, Key: security.TlsKey}
    secretTargetRefs = append(secretTargetRefs, key)
} else {
    key := kedav1alpha1.AuthSecretTargetRef{Parameter: "key", Name: secret.Name, Key: security.UserKey}
    secretTargetRefs = append(secretTargetRefs, key)
}

pierDipi (commented):
Thanks @MdSahil-oss for the updates!

I left another comment, in addition, this comment is missing a resolution: https://github.com/knative-sandbox/eventing-kafka-broker/pull/3073/files#r1205254926

PR #3078, opened 2023-05-10
Implement the configurable ordering for kafka source

### LINKED ISSUE #3076: Enable orderedConsumerVerticle as an option for kafkaSource
**Problem**
Make Ordering configurable.
Dispatcher has both orderedConsumerVerticle and unorderedConsumerVerticle implemented.
So far only orderedConsumerVerticle is for use, and unorderedConsumerVerticle is not for use. We should enable unorderedConsumerVerticle, by adding a configuration option in kafkaSource CRD.

End users.

**Exit Criteria**
Add a configuration option in the CRD od kafkaSource, e.g. spec.ordering to specify either ordered or unordered, for the dispatcher to create either orderedconsumerverticle or unorderedconsumerverticle.

**Time Estimate (optional):** 
One.

None

houshengbo:
Fixes #3076 

## Proposed Changes

- Added a configuration option in the CRD of kafkaSource, to configure ordered or unordered consumer verticle.

**Release Note**

```release-note

```

**Docs**


control-plane/config/eventing-kafka-broker/100-source/100-kafka-source.yaml
```diff
@@ -40,6 +40,14 @@ spec:
           #
           # see issue: https://github.com/knative/serving/issues/912
           x-kubernetes-preserve-unknown-fields: true
+          properties:
+            spec:
```
matzew, line 44: I think I'd like to see the other "spec" properties first, before we _just_ add this here

houshengbo: That's refactoring of the crd. I prefer to change in other prs.

matzew: yeah, exactly. That's why I said before to isolate the change, and land it _before_ just doing this one

PR #3080, opened 2023-05-10
Implement the configurable ordering for kafka source

### LINKED ISSUE #3076: Enable orderedConsumerVerticle as an option for kafkaSource

houshengbo:
Fixes #3076 

## Proposed Changes

- Added a configuration option in the CRD of kafkaSource, to configure ordered or unordered consumer verticle.

**Release Note**

```release-note

```

**Docs**


matzew:
Should we add some DP/e2e tests for this behavior, with the source?

control-plane/pkg/apis/sources/v1beta1/kafka_types.go
```diff
@@ -82,6 +83,12 @@ type KafkaSourceSpec struct {
 	// +optional
 	Delivery *eventingduckv1.DeliverySpec `json:"delivery,omitempty"`
 
+	// Ordering is the type of the consumer verticle.
+	// Should be ordered or unordered.
+	// By default, it is ordered.
+	// +optional
+	Ordering eventing.DeliveryOrdering `json:"ordering,omitempty"`
```
pierDipi, line 90: We should:
- add validation for this field in `Validate`
- add defaulting for this field in `SetDefaults`
- since it's optional I'd make it a pointer instead of relying on "empty string == unset"

control-plane/pkg/reconciler/source/source.go
```diff
@@ -116,7 +116,7 @@ func (r Reconciler) reconcileConsumerGroup(ctx context.Context, ks *sources.Kafk
 				BackoffDelay:  pointer.String("PT0.3S"),
 				Timeout:       pointer.String("PT600S"),
 			},
-			Ordering: DefaultDeliveryOrder,
+			Ordering: *ks.Spec.Ordering,
```
pierDipi, line 119: Same here, I guess we can do it once for both branches

houshengbo:
@pierDipi I realized there is a calculation of the test coverage. We do not have any test coverage for the directory `control-plane/pkg/apis/sources/v1beta1`. If I change any file under this directory, it will calculate the test coverage of this directory, which pulls down the test coverage of this PR.
How shall we handle this?

control-plane/pkg/reconciler/source/source.go
houshengbo: added the logic check back.

pierDipi (approved):
Thanks!

/lgtm
/approve

pierDipi:
> @pierDipi I realized there is a calculation of the test coverage. We do not have any test coverage for the directory `control-plane/pkg/apis/sources/v1beta1`. If I change any file under this directory, it will calculate the test coverage of this directory, which pulls down the test coverage of this PR. How shall we handle this?

No worries!

PR #3084, opened 2023-05-15
Remove KafkaSource and KafkaChannel migrators

pierDipi:
In 1.10 we didn't realease the KafkaSource and KafkaChannel in eventing-kafka since they were deprecated, so we don't need the migration logic anymore.

`post-install` job will be kept but it will be a no-op until we need migration logic for future needs.


matzew (approved):
/lgtm
/approve

PR #3091, opened 2023-05-20
replaced error to stop consumer from reconciling with exponentially delay

### LINKED ISSUE #3046: Slow consumergroup reconciliation under load
**Describe the bug**

In [1], we return an error, however, that method is called in the consumer group reconciler when a consumer is not ready which is a normal state at the beginning when consumers have been just created, so we shouldn't return an error because that causes the consumer group to be reconciled again with an exponentially increasing delay causing slow time to ready.

This is especially evident when scaling up with a high load (therefore when dispatcher pod is slow to become ready).

I'd replace that returned error with:
```
	// It is "normal" to have non-ready consumers, and we will get notified when their status change,
	// so we don't need to return an error here which causes the object to be queued with an
	// exponentially increasing delay.
	return nil
```

[1] https://github.com/knative-sandbox/eventing-kafka-broker/blob/5cda5463aa2fa060179674fe7b3237abb836ee06/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_lifecycle.go#L57-L65

 **Expected behavior**

No error is returned by `MarkReconcileConsumersFailedCondition`

 **To Reproduce**

It's hard to reproduce but you need to create lots of KafkaSources while scaling them up and down concurrently

 **Knative release version**
`dev`

 **Additional context**

/good-first-issue

Abhishek357:
Fixes #3046

## Proposed Changes

- I replaced the returned error, which prevents the consumer group from being reconciled again, with an exponentially increasing delay which eventually caused a slow time to ready.

### DIFF
--- a/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_lifecycle.go
+++ b/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_lifecycle.go
@@ -61,7 +61,10 @@ func (cg *ConsumerGroup) MarkReconcileConsumersFailedCondition(condition *apis.C
 		condition.GetMessage(),
 	)
 
-	return fmt.Errorf("consumers aren't ready, %v: %v", condition.GetReason(), condition.GetMessage())
+	// It is "normal" to have non-ready consumers, and we will get notified when their status change,
+	// so we don't need to return an error here which causes the object to be queued with an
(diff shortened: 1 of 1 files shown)


pierDipi:
@Abhishek357 can you take a look at the test failures?

pierDipi:
Superseded by https://github.com/knative-extensions/eventing-kafka-broker/pull/3293

PR #3103, opened 2023-05-25
Retry offset commit

pierDipi:
Before this patch, we were caching the new offset before committing to Kafka the offset, causing the offset to not be stored until a new record is successfully processed.


matzew (approved):
/lgtm
/approve

PR #3111, opened 2023-05-26
Update e2e tests to have default value for BROKER_CLASS

Cali0707:
# Fixes

When trying to run the e2e tests by following the guide provided in `DEVELOPMENT.md`, I was shown the following error: `ERROR: Broker class is not defined. Specify it with 'BROKER_CLASS' env var`.

## Proposed Changes

- Set a default value for `BROKER_CLASS` if it is undefined when running the e2e tests.

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -29,10 +29,11 @@ if [ "${EVENTING_KAFKA_BROKER_CHANNEL_AUTH_SCENARIO:-""}" != "" ]; then
 fi
 
 if [[ -z "${BROKER_CLASS}" ]]; then
-  fail_test "Broker class is not defined. Specify it with 'BROKER_CLASS' env var."
-else
-  echo "BROKER_CLASS is set to '${BROKER_CLASS}'. Running tests for that broker class."
+  BROKER_CLASS=KafkaBroker
+  echo "BROKER_CLASS is not defined, falling back to the default BROKER_CLASS=KafkaBroker. Override this with the 'BROKER_CLASS' env var."
 fi
+
+echo "BROKER_CLASS is set to '${BROKER_CLASS}'. Running tests for that broker class."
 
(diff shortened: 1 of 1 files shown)


test/e2e-tests.sh
```diff
@@ -29,11 +29,12 @@ if [ "${EVENTING_KAFKA_BROKER_CHANNEL_AUTH_SCENARIO:-""}" != "" ]; then
 fi
 
 if [[ -z "${BROKER_CLASS}" ]]; then
-  fail_test "Broker class is not defined. Specify it with 'BROKER_CLASS' env var."
-else
-  echo "BROKER_CLASS is set to '${BROKER_CLASS}'. Running tests for that broker class."
+  BROKER_CLASS=KafkaBroker
```
matzew, line 32: on the actual test config, on the infra repo, we do populate them. 

Perhaps it's also good to reflect this in the `DEVELOPMENT` md file? 

/cc @aliok

aliok: I am ok with it. I've seen this change suggested many times now :)

test/e2e-tests.sh
```diff
@@ -29,11 +29,12 @@ if [ "${EVENTING_KAFKA_BROKER_CHANNEL_AUTH_SCENARIO:-""}" != "" ]; then
...
+  echo "BROKER_CLASS is not defined, falling back to the default BROKER_CLASS=KafkaBroker. Override this with the 'BROKER_CLASS' env var."
```
pierDipi, line 33:
```suggestion
  BROKER_CLASS=Kafka
  echo "BROKER_CLASS is not defined, falling back to the default BROKER_CLASS=Kafka. Override this with the 'BROKER_CLASS' env var."
```

pierDipi (approved):
/lgtm
/approve

PR #3122, opened 2023-05-31
Support exposing https address ekb controller

### LINKED ISSUE #3108: Eventing TLS: support exposing https address in Kafka Broker controller
**Problem**

This is similar to https://github.com/knative/eventing/issues/6918.

As the [Eventing TLS feature track](https://docs.google.com/document/d/1H-x_oji8LqkCyd7tlsSyclmUe7FAmEJPgRxOU_0pkn8/edit) describes we should support TLS for Broker.

In particular, kafka-controller should:
- when the feature flag (see https://github.com/knative/eventing/issues/6836) is set to disabled
  - no behavior change 
- when the feature flag (see https://github.com/knative/eventing/issues/6836) is set to permissive
  - expose both http and https endpoints in the new `status.addresses` (see https://github.com/knative/pkg/issues/2711) field
  - the existing `status.address` will continue to have the existing `http` endpoint
- when the feature flag (see https://github.com/knative/eventing/issues/6836) is set to strict
  - expose only the `https` endpoints in both the new `status.addresses` and the existing `status.address`

**Time Estimate (optional):** 
1

- Depends on https://github.com/knative-sandbox/eventing-kafka-broker/issues/3107

/good-first-issue

Cali0707:
Fixes #3108 

## Proposed Changes

- Update the broker reconciler to update the `Status.Address` and `Status.Addresses` as required
- Update the broker reconciler tests to account for the updates

**Release Note**

```release-note
Added support for exposing an https address from the broker.
```


control-plane/pkg/reconciler/broker/broker_test.go
```diff
@@ -2399,6 +2662,143 @@ func brokerFinalization(t *testing.T, format string, env config.Env) {
...
+						BrokerAddressable(&env),
+						StatusBrokerProbeSucceeded,
+						BrokerDLSResolved(ServiceURL),
+						BrokerConfigMapAnnotations(),
+						WithTopicStatusAnnotation(BrokerTopic()),
+						WithBrokerAddresses([]duckv1.Addressable{
+							{
+								Name:    pointer.String("https"),
+								URL:     httpsURL(BrokerName, BrokerNamespace),
+								CACerts: pointer.String(testCaCerts),
```
Vishal-Chdhry, line 2790: Very minor, If you put the `testCaCerts` in the `var()` block, you won't have to create a seperate pointer to it.

control-plane/pkg/reconciler/broker/controller.go
```diff
@@ -119,6 +119,10 @@ func NewController(ctx context.Context, watcher configmap.Watcher, env *config.E
 	reconciler.Tracker = impl.Tracker
 
 	secretinformer.Get(ctx).Informer().AddEventHandler(controller.HandleAll(reconciler.Tracker.OnChanged))
+	secretinformer.Get(ctx).Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: controller.FilterWithName(brokerIngressTLSSecretName),
+		Handler:    controller.HandleAll(globalResync),
+	})
```
Vishal-Chdhry, line 125: Since we are tracking all changes in secrets in the line above already, we probably don't have to specifically listen for changes to `brokerIngressTLSSecretName`. 

Can you check whether the tests pass without adding this line

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -226,7 +235,34 @@ func (r *Reconciler) reconcileKind(ctx context.Context, broker *eventing.Broker)
...
+		}
+
+		httpsAddress := receiver.HTTPSAddress(ingressHost, broker, caCerts)
+		broker.Status.Address = &httpsAddress
+		broker.Status.Addresses = []pkgduckv1.Addressable{httpsAddress}
+	} else {
+		httpAddress := receiver.HTTPAddress(ingressHost, broker)
+		broker.Status.Address = &httpAddress
+		broker.Status.Addresses = []pkgduckv1.Addressable{httpAddress}
+	}
```
pierDipi, line 263: We need to set the address only after the prober is successful

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -226,7 +235,34 @@ func (r *Reconciler) reconcileKind(ctx context.Context, broker *eventing.Broker)
...
+
+	address := broker.Status.Address.URL.URL()
```
pierDipi, line 265: As a follow up, we need to make the prober TLS aware, because because currently if I'd enable strict transport encryption: line 265 would use the HTTPS endpoint but the prober is not configured or aware of the CA certs and it would fail

Cali0707: I made an issue for this here: https://github.com/knative-sandbox/eventing-kafka-broker/issues/3128

control-plane/pkg/reconciler/broker/broker.go
Cali0707: Good point, I've updated the order of events so that it makes more sense

Cali0707:
@pierDipi I'm not totally sure why the `Test/test/Unit Tests` action is failing - the tests seem to be passing on my system locally

control-plane/pkg/reconciler/broker/controller_test.go
```diff
@@ -46,6 +47,22 @@ func TestNewController(t *testing.T) {
...
+		Name:      brokerIngressTLSSecretName,
+	}
+
+	_ = secretinformer.Get(ctx).Informer().GetStore().Add(&corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: secret.Namespace,
+			Name:      secret.Name,
+		},
+		Data: map[string][]byte{
+			"ca.crt": loadCert(),
```
pierDipi, line 61:
```suggestion
			"ca.crt": eventingtlstesting.CA,
```

pierDipi: and we can remove the loadCert function

control-plane/pkg/receiver/address_test.go
```diff
@@ -40,3 +40,60 @@ func TestAddress(t *testing.T) {
...
+A1UdEwQCMAAwCwYDVR0PBAQDAgTwMBQGA1UdEQQNMAuCCWxvY2FsaG9zdDAdBgNV
+HQ4EFgQUnu/3vqA3VEzm128x/hLyZzR9JlgwDQYJKoZIhvcNAQELBQADggEBAFc+
+1cKt/CNjHXUsirgEhry2Mm96R6Yxuq//mP2+SEjdab+FaXPZkjHx118u3PPX5uTh
+gTT7rMfka6J5xzzQNqJbRMgNpdEFH1bbc11aYuhi0khOAe0cpQDtktyuDJQMMv3/
+3wu6rLr6fmENo0gdcyUY9EiYrglWGtdXhlo4ySRY8UZkUScG2upvyOhHTxVCAjhP
+efbMkNjmDuZOMK+wqanqr5YV6zMPzkQK7DspfRgasMAQmugQu7r2MZpXg8Ilhro1
+s/wImGnMVk5RzpBVrq2VB9SkX/ThTVYEC/Sd9BQM364MCR+TA1l8/ptaLFLuwyw8
+O2dgzikq8iSy1BlRsVw=
+-----END CERTIFICATE-----
+`
```
pierDipi, line 69: Can we use `eventingtlstesting.CA` instead of defining our own?

pierDipi (approved):
/lgtm
/approve

PR #3124, opened 2023-06-01
Eventing TLS: Pass CA certs to the data plane

### LINKED ISSUE #3118: Eventing TLS: Pass CA certs to the data plane
**Problem**

After https://github.com/knative-sandbox/eventing-kafka-broker/issues/3109, we're now resolving destinations to full addressables which contains URL and CA certs that we need to pass to the data plane through the "contract". 

Our contract (which is a protobuf message) contains a few URLs:

destination: https://github.com/knative-sandbox/eventing-kafka-broker/blob/06e73351124a536eeb208f223e2b654b9a044dfa/proto/contract.proto#L111-L112

deadLetter: https://github.com/knative-sandbox/eventing-kafka-broker/blob/06e73351124a536eeb208f223e2b654b9a044dfa/proto/contract.proto#L75-L76

replyUrl: https://github.com/knative-sandbox/eventing-kafka-broker/blob/06e73351124a536eeb208f223e2b654b9a044dfa/proto/contract.proto#L115-L116

All of them are HTTP URLs that might be TLS endpoints and our data plane needs to trust a specific CA for each of them.

In this issue, we need to:
- add CA certs fields to our contract, for example: destinationCACerts, replyUrlCACerts, and deadLetterCACerts as strings (each living alongside the URL counterpart)
- generate the code that is generated out of the protobuf definition with `hack/update-codegen.sh`
- Pass CA Certs to the data plane [1]

[1] for example, in the trigger destination case, we, currently, have this:

https://github.com/knative-sandbox/eventing-kafka-broker/blob/06e73351124a536eeb208f223e2b654b9a044dfa/control-plane/pkg/reconciler/trigger/trigger.go#L308-L327

we would need to additionally do something like this:
```golang
if destination.CACerts != nil {
  egress.DestinationCACerts = *destination.CACerts
}
```

For other fields, the logic is the same, names might vary.
To find the places, we need such logic, use the URIResolver usages as in https://github.com/knative-sandbox/eventing-kafka-broker/issues/3109

**Exit Criteria**

- Pass CA certs to the data plane
- Unit tests that verify that the contract contains CA certs

**Time Estimate (optional):** 
2

/help

Vishal-Chdhry:
Fixes #3118 

## Proposed Changes
WIP: Needs tests

- :gift: Add new feature
- Pass CA certs to the data plane

**Release Note**

```release-note
Eventing TLS: Pass CA certs to the data plane
```

**Docs**


proto/contract.proto
```diff
@@ -111,16 +114,20 @@ message Egress {
...
   string destination = 2;
 
+  //destination CA Cert is the CA Cert used for HTTPS communication through destination
+  string destinationCACerts = 15;
+
   oneof replyStrategy {
-    // Send the response to an url
-    string replyUrl = 3;
+    // Sends message to reply url with the CA Cert provided
+    replyUrlAndCert reply = 3;
```
pierDipi, line 122: Isn't this a breaking change? 

Like is it possible to serialize and deserialize a proto message of an earlier version with this new version? Ideally, it should be possible to serialize and deserialize a proto message of this new version with the old version (in case of downgrades).
Basically, we need to make sure that the new proto schema is forward and backward compatible with the previous version.

If the above are not possible, what I'd is to add a separate field on the `Egress`: `string replyCACerts = 15;` and we need to make sure we set that field when `ReplyStrategy` is `replyUrl`

Vishal-Chdhry: @pierDipi Can we deprecate reply url and add replyUrlAndCert?

Vishal-Chdhry:
```
-     (Apache License 2.0) native-linux-x86_64 (com.aayushatharva.brotli4j:native-linux-x86_64:1.8.0 - https://github.com/hyperxpro/Brotli4j/natives/native-linux-x86_64)
+     (Apache License 2.0) native-linux-aarch64 (com.aayushatharva.brotli4j:native-linux-aarch64:1.8.0 - https://github.com/hyperxpro/Brotli4j/natives/native-linux-aarch64)
```

How do I fix this error, It is probably because I am using ARM64 

/cc @matzew @pierDipi @creydr

proto/contract.proto
pierDipi: We can, does that ensure backward and forward compatibility?

We need to think about:
- what happens (or how to solve) when a previous version of the dispatcher sees a message with `replyUrlAndCert` (it doesn't know anything about `replyUrlAndCert`)
- what happens (or how to solve) when the new version sees the old `replyUrl` and there isn't a `replyUrlAndCert` (this might happen when the upgrade is in progress and so the controller is at the previous version and dispatcher is at the "current version")

In theory, to solve those by adding and deprecating we need to set both fields for a few releases but that clashes with "oneof"

Vishal-Chdhry:
@pierDipi sorry for the delay on this, I have exams right now, I will be free by Friday evening

Vishal-Chdhry:
Most of the failing tests are because I cant generate `getReplyUrl()` function in java. It was removed by this https://github.com/knative-sandbox/eventing-kafka-broker/pull/3124#discussion_r1212607715 and I cant regenerate it
/cc @pierDipi

control-plane/pkg/contract/contract.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.26.0
-// 	protoc        v3.17.3
+// 	protoc-gen-go v1.28.1
+// 	protoc        v3.12.4
```
pierDipi, line 4: I think one of the reason you have that problem is that the versions don't align, you need to have these exact versions for protoc and protoc-gen-go
```
// 	protoc-gen-go v1.26.0
// 	protoc        v3.17.3
```


if I run it on this PR, I get a commit like https://github.com/pierdipi/eventing-kafka-broker/commit/87de9f101a49edfb9cb789ab6e4ddca3349e3431

proto/contract.proto
```diff
@@ -111,6 +114,9 @@ message Egress {
   // destination is the sink where events are sent.
   string destination = 2;
 
+  //destination CA Cert is the CA Cert used for HTTPS communication through destination
```
pierDipi, line 117: nit
```suggestion
  // destination CA Cert is the CA Cert used for HTTPS communication through destination
```

proto/contract.proto
```diff
@@ -121,6 +127,10 @@ message Egress {
     // Discard response.
     Empty discardReply = 9;
   }
+
+  //replyUrl CA Cert is the CA Cert used for HTTPS communication through replyUrl
```
pierDipi, line 131: nit
```suggestion
  // replyUrl CA Cert is the CA Cert used for HTTPS communication through replyUrl
```

control-plane/pkg/reconciler/trigger/trigger_test.go
```diff
@@ -2825,6 +2827,85 @@ func triggerFinalizer(t *testing.T, format string, env config.Env) {
 				},
 			},
 		},
+		{
+			Name: "Reconciled normal - With Subscriber CA Cert",
```
pierDipi, line 2831: Can we add another similar test for the dead letter sink CA certs? or even expand this one covering dead letter sink CA certs?

pierDipi (commented):
Thanks @Vishal-Chdhry, I left a few comments.

Can we cover KafkaChannel subscribers using the field `SubscriberCACerts` on `subscriber` below? https://github.com/knative-sandbox/eventing-kafka-broker/blob/898e479be20128819351ca341aff9d4e6f8ba50e/control-plane/pkg/reconciler/channel/channel.go#L512-L557

---

There are build errors and you need to re-run `hack/update-codegen.sh` https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-sandbox_eventing-kafka-broker/3124/build-tests_eventing-kafka-broker_main/1668565006649659392

pierDipi:
@Vishal-Chdhry I think you might need to run hack/update-codegen.sh https://github.com/Vishal-Chdhry/eventing-kafka-broker/pull/1

Vishal-Chdhry:
Yes, I have actually tried to do that, but it doesn't add anything for some reason. One reason I can think of is that I am on a Mac and bash4 is not directly available on Mac.

So I am using Lima to run a WSL-like Ubuntu setup on my Mac. That might be causing update -codegen to not work properly.

@pierDipi Can you please run it on top of my PR and commit it here, that will be really helpful.

pierDipi:
@Vishal-Chdhry yes, that's what I did in this PR: https://github.com/Vishal-Chdhry/eventing-kafka-broker/pull/1 you can merge it and it will be added to this PR

Vishal-Chdhry:
Thank you pier!

pierDipi (approved):
Thanks!

/lgtm
/approve

PR #3130, opened 2023-06-01
Support TLS in prober

### LINKED ISSUE #3128: Eventing TLS: support https in the prober
**Problem**
The prober currently only supports http. As mentioned in #3122 the prober should support TLS, so that when we have strict transport encryption the prober does not fail.

**Exit Criteria**
A test showing that the prober works with TLS and https addresses.

**Time Estimate (optional):** 
1

- Related to #3108 

/good-first-issue

Cali0707:
Fixes #3128 

## Proposed Changes

- Add a new constructor for the async prober so that it can support TLS, allowing us to probe even when the address uses https

**Release Note**

```release-note
Added TLS support to the prober.
```

### DIFF
--- a/control-plane/pkg/prober/async_prober.go
+++ b/control-plane/pkg/prober/async_prober.go
@@ -18,6 +18,9 @@ package prober
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"net/http"
 	"sync"
 	"time"
 
@@ -55,6 +58,21 @@ func NewAsync(ctx context.Context, client httpClient, port string, IPsLister IPs
 		IPsLister: IPsLister,
 		port:      port,
 	}
+}
+
+func NewAsyncWithTLS(ctx context.Context, port string, IPsLister IPsLister, enqueue EnqueueFunc, caCerts *string) Prober {
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM([]byte(*caCerts))
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: caCertPool,
+			},
+		},
+	}
+
+	return NewAsync(ctx, client, port, IPsLister, enqueue)
 }
 
 func (a *asyncProber) Probe(ctx context.Context, addressable Addressable, expected Status) Status {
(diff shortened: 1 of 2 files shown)


pierDipi:
(I left a early comment since it's WIP)

control-plane/pkg/prober/async_prober.go
```diff
@@ -57,6 +60,21 @@ func NewAsync(ctx context.Context, client httpClient, port string, IPsLister IPs
...
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: caCertPool,
+			},
+		},
+	}
+
+	return NewAsync(ctx, client, port, IPsLister, enqueue)
```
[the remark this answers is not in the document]
Cali0707: Ah I didn't know this existed, thanks for pointing it out - this is exactly what I needed!

Cali0707:
> Another interesting aspect for TLS support on the prober is, we would need to offer/implement a mechanism to rotate CA certs on the underlying HTTP(s) client.

How do you see this working?

Cali0707:
@pierDipi its worth noting that I had to make a TLS cert that had an IP SAN - is there anything we need to do to ensure that the certs being issued in the cluster all have an IP SAN?

control-plane/pkg/prober/async_prober_test.go
```diff
@@ -158,3 +215,99 @@ func TestAsyncProber(t *testing.T) {
 		})
 	}
 }
+
+// Adapted from https://github.com/knative/eventing/blob/57d78e060db6e0d2f3046eeedd27137cfa4fe0bc/pkg/eventingtls/eventingtlstesting/eventingtlstesting.go#L84
```
pierDipi, line 219: Did you consider changing the one in Eventing instead adding our own?
I'm sure we will need the same feature for other use cases

pierDipi:
> @pierDipi its worth noting that I had to make a TLS cert that had an IP SAN - is there anything we need to do to ensure that the certs being issued in the cluster all have an IP SAN?

Awesone, I'd update the one in Eventing with the additional IP SAN, so that we can reuse them here

control-plane/pkg/prober/async_prober_test.go
Cali0707: I didn't think of this at the time because the one in Eventing isn't exported, but I can see why it would be a good idea to add this there. I'll open a PR

pierDipi:
/lgtm
/approve

Using the updated Eveting library can be done in a follow up

PR #3134, opened 2023-06-05
[WIP] Unique secret names

### LINKED ISSUE #3133: Tests should use unique names when copying secrets into test namespace
**Describe the bug**
Some tests copy secrets from the default namespace into the test namespace.
[Example 1](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/test/rekt/features/kafka_source_create_secrets_after.go#L56)
[Example 2](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/test/rekt/features/kafka_source.go#L323)

The secrets should be copied under unique names into the target namespace, this would allow for:
* running multiple tests using the secrets in parallel
* use a common/single namespace to run the tests (e.g. when testing with Istio in a pre-defined namespace)
* proper teardown of resources (secrets) even if tests run in parallel

Currently, the secrets are copied into the test namespace and left there until the namespace is deleted. When running `CreateSecretsAfterKafkaSource` in the same namespace again the secret is already there and the test fails.

 **Expected behavior**

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative release version**

 **Additional context**

mgencur:
Fixes #3133

TBD: Send change to knative.dev/eventing

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


pierDipi (approved):
/lgtm
/approve

PR #3146, opened 2023-06-09
Support rotating CA certs in prober

### LINKED ISSUE #3140: Eventing TLS: Support rotating CA certs in prober
**Problem**
When using a prober with TLS, the prober should support rotating CA certs on the underlying https client.

I think this can be achieved by updating the prober to use `GetCertificate()` instead of providing a `Certificates` array, and using the secret informer to provide an up to date certificate to the prober.

**Exit Criteria**
A unit test where the certificate is updated and the prober is still able to probe successfully after the update.

**Time Estimate (optional):** 
1

Cali0707:
Fixes #3140 

## Proposed Changes

- Provide a method to rotate the ca certs for a given prober, which can be called as appropriate by a given controller

**Release Note**

```release-note
Probers now support rotating CA certificates
```

### DIFF
--- a/control-plane/pkg/prober/async_prober.go
+++ b/control-plane/pkg/prober/async_prober.go
@@ -28,6 +28,10 @@ import (
 	"knative.dev/pkg/logging"
 )
 
+var (
+	cacheExpiryTime = time.Minute * 30
+)
+
 type IPsLister func(addressable Addressable) ([]string, error)
 
 type asyncProber struct {
@@ -54,7 +58,7 @@ func NewAsync(ctx context.Context, client httpClient, port string, IPsLister IPs
 		client:    client,
 		enqueue:   enqueue,
 		logger:    logger,
-		cache:     NewLocalExpiringCache(ctx, 30*time.Minute),
+		cache:     NewLocalExpiringCache(ctx, cacheExpiryTime),
 		IPsLister: IPsLister,
 		port:      port,
 		clientMu:  sync.Mutex{},
@@ -62,18 +66,11 @@ func NewAsync(ctx context.Context, client httpClient, port string, IPsLister IPs
 }
 
 func NewAsyncWithTLS(ctx context.Context, port string, IPsLister IPsLister, enqueue EnqueueFunc, caCerts *string) (Prober, error) {
-	tlsClient := eventingtls.ClientConfig{CACerts: caCerts}
-	tlsClientConfig, err := eventingtls.GetTLSClientConfig(tlsClient)
+	newClient, err := makeHttpClientWithTLS(caCerts)
 	if err != nil {
 		return nil, err
 	}
-
-	client := &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: tlsClientConfig,
-		},
-	}
-
+	client := newClient
 	return NewAsync(ctx, client, port, IPsLister, enqueue), nil
 }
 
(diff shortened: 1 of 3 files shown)


control-plane/pkg/prober/async_prober.go
```diff
@@ -161,3 +165,28 @@ func (a *asyncProber) Probe(ctx context.Context, addressable Addressable, expect
 func (a *asyncProber) enqueueArg(_ string, arg interface{}) {
 	a.enqueue(arg.(types.NamespacedName))
 }
+
+// Resets the cache to a new empty cache with the given expiry time
+func (a *asyncProber) resetCache(ctx context.Context, expiry time.Duration) {
+	a.clientMu.Lock()
+	defer a.clientMu.Unlock()
+	a.cache = NewLocalExpiringCache(ctx, expiry)
+}
```
pierDipi, line 174: Is this method used only for testing?

control-plane/pkg/prober/async_prober.go
```diff
@@ -161,3 +165,28 @@ func (a *asyncProber) Probe(ctx context.Context, addressable Addressable, expect
...
+		return err
+	}
+
+	a.clientMu.Lock()
+	defer a.clientMu.Unlock()
+	a.client = &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsClientConfig,
+		},
+	}
```
pierDipi, line 189: This code is very similar to the one in `New..TLS` above, can we refactor it to reduce duplication?

control-plane/pkg/prober/async_prober.go
```diff
@@ -161,3 +165,28 @@ func (a *asyncProber) Probe(ctx context.Context, addressable Addressable, expect
...
+func (a *asyncProber) RotateRootCaCerts(caCerts *string) error {
+	tlsClient := eventingtls.ClientConfig{CACerts: caCerts}
+	tlsClientConfig, err := eventingtls.GetTLSClientConfig(tlsClient)
+	if err != nil {
+		return err
+	}
+
+	a.clientMu.Lock()
+	defer a.clientMu.Unlock()
+	a.client = &http.Client{
```
pierDipi, line 185: Before resetting this, we need to call `CloseIdleConnections` on the previous "client"

control-plane/pkg/prober/async_prober.go
Cali0707: Yeah this method is currently only used for tests to ensure that the prober re-probes the pod with the new certificates instead of falling back to the cached state. It could be useful if we want to always reset the prober cache when the CA certs get rotated, to ensure that everything is still reachable. WDYT?

pierDipi: Did you consider setting a relatively low expiry time at the creation time and wait for it to expire for tests?

Making the code more testable is absolutely necessary, what I'd avoid is creating a different code path that is just used for testing

control-plane/pkg/prober/async_prober.go
```diff
@@ -161,3 +161,30 @@ func (a *asyncProber) Probe(ctx context.Context, addressable Addressable, expect
...
+
+func makeHttpClientWithTLS(caCerts *string) (*http.Client, error) {
+	tlsClient := eventingtls.ClientConfig{CACerts: caCerts}
+	tlsClientConfig, err := eventingtls.GetTLSClientConfig(tlsClient)
+	if err != nil {
+		return nil, err
+	}
+
+	return &http.Client{
+		Transport: &http.Transport{
```
pierDipi, line 186: Instead of creating a new one, I'd close the default one: https://github.com/knative/eventing/blob/825a237c4255f18f3006d96bfa97accab39473f8/pkg/adapter/v2/cloudevents.go#L154-L155

The difference is that, it has some good default values that we're not setting here https://go.dev/src/net/http/transport.go:
```golang
// DefaultTransport is the default implementation of Transport and is
// used by DefaultClient. It establishes network connections as needed
// and caches them for reuse by subsequent calls. It uses HTTP proxies
// as directed by the environment variables HTTP_PROXY, HTTPS_PROXY
// and NO_PROXY (or the lowercase versions thereof).
var DefaultTransport RoundTripper = &Transport{
	Proxy: ProxyFromEnvironment,
	DialContext: defaultTransportDialContext(&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
	}),
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}
```

pierDipi (approved):
/lgtm
/approve

PR #3148, opened 2023-06-12
Upgrade to latest dependencies and fix compile issues

creydr:
As per title and fixing the compile issues from #3126


creydr:
@pierDipi could you PTAL at this?

pierDipi (approved):
Thanks!

/lgtm
/approve

Issue #3153, opened 2023-06-15
Allow defining topic prefix for Kafka Broker and KafkaChannel


pierDipi:
**Problem**

Currently, topic names are auto-generated with a prefix based on the resource name and namespace.

- https://github.com/knative-sandbox/eventing-kafka-broker/blob/898e479be20128819351ca341aff9d4e6f8ba50e/control-plane/pkg/reconciler/broker/broker.go#L52-L54
- https://github.com/knative-sandbox/eventing-kafka-broker/blob/898e479be20128819351ca341aff9d4e6f8ba50e/control-plane/pkg/reconciler/channel/channel.go#L63-L64

This makes difficult to use the same Apache Kafka cluster with multiple Knative Eventing installations in different Kubernetes clusters.

Similarly to https://github.com/knative-sandbox/eventing-kafka-broker/commit/293b7989eaa9df60525df4515a0eff060a95eeb9, we could define in [`config-kafka-features`](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/test/config/100-config-kafka-features.yaml) topic names template:

```yaml
brokers.topic.template: "knative-broker-{{ .Namespace }}-{{ .Name }}"
channels.topic.template: "knative-messaging-kafka.{{ .Namespace }}.{{ .Name }}"
```

and then use those templates for the underlying topic of Kafka Broker or KafkaChannel.

This will allow users to change the template for each Eventing installation.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?
System Operator

**Exit Criteria**

New tests are added that verify that by changing the template, the used topic are different from the default one.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?
2

**Additional context (optional)**

/area broker
/area channel
/area control-plane

matzew:
Each broker installation could than have their own topics, in a centralized kafka system

PR #3157, opened 2023-06-19
Allow defining topic prefixes

Cali0707:
As part of #3153 we need to be able to add broker and channel topic prefixes.
## Proposed Changes

- Update the kafka features to support broker topic template and channel topic template

### DIFF
--- a/control-plane/pkg/apis/config/features.go
+++ b/control-plane/pkg/apis/config/features.go
@@ -40,6 +40,8 @@ type features struct {
 	DispatcherOrderedExecutorMetrics feature.Flag
 	ControllerAutoscaler             feature.Flag
 	TriggersConsumerGroupTemplate    *template.Template
+	BrokersTopicTemplate             *template.Template
+	ChannelsTopicTemplate            *template.Template
 }
 
 type KafkaFeatureFlags struct {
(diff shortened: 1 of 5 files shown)


control-plane/pkg/apis/config/features.go
```diff
@@ -95,13 +107,15 @@ func (f *KafkaFeatureFlags) IsControllerAutoscalerEnabled() bool {
...
+}
 
-	return result.String(), nil
+func (f *KafkaFeatureFlags) ExecuteBrokersTopicTemplate(brokerMetadata v1.ObjectMeta) (string, error) {
+	return executeTemplateToString(f.features.BrokersTopicTemplate, brokerMetadata, "unable to execute brokers topic template: %w")
+}
+
+func (f *KafkaFeatureFlags) ExecuteChannelsTopicTemplate(channelMetadata v1.ObjectMeta) (string, error) {
+	return executeTemplateToString(f.features.ChannelsTopicTemplate, channelMetadata, "unable to execute channels topic template: %w")
 }
```
pierDipi, line 119: Are these used somewhere?

Cali0707: I'm currently using them on a local branch, to replace some of the logic in places like here: https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/pkg/reconciler/broker/broker.go#L314-L316

I am trying to split the feature up into smaller PRs rather than one big one

pierDipi: No, that's good!

/lgtm
/approve

PR #3162, opened 2023-06-20
[WIP]: Update Broker and Controller to use topic name templates

### LINKED ISSUE #3153: Allow defining topic prefix for Kafka Broker and KafkaChannel

Cali0707:
Fixes #3153 

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -37,6 +37,7 @@ import (
 	"knative.dev/pkg/reconciler"
 	"knative.dev/pkg/resolver"
 
+	apisconfig "knative.dev/eventing-kafka-broker/control-plane/pkg/apis/config"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/config"
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/contract"
 	coreconfig "knative.dev/eventing-kafka-broker/control-plane/pkg/core/config"
@@ -50,9 +51,6 @@ import (
 )
 
 const (
-	// TopicPrefix is the Kafka Broker topic prefix - (topic name: knative-broker-<broker-namespace>-<broker-name>).
-	TopicPrefix = "knative-broker-"
-
 	// ExternalTopicAnnotation for using external kafka topic for the broker
 	ExternalTopicAnnotation = "kafka.eventing.knative.dev/external.topic"
 
@@ -80,8 +78,9 @@ type Reconciler struct {
 
 	BootstrapServers string
 
-	Prober  prober.Prober
-	Counter *counter.Counter
+	Prober            prober.Prober
+	Counter           *counter.Counter
+	KafkaFeatureFlags *apisconfig.KafkaFeatureFlags
 }
 
 func (r *Reconciler) ReconcileKind(ctx context.Context, broker *eventing.Broker) reconciler.Event {
@@ -311,7 +310,10 @@ func (r *Reconciler) reconcileBrokerTopic(broker *eventing.Broker, securityOptio
 		}
 	} else {
 		// no external topic, we create it
-		topicName = kafka.BrokerTopic(TopicPrefix, broker)
+		topicName, err = r.KafkaFeatureFlags.ExecuteBrokersTopicTemplate(broker.ObjectMeta)
+		if err != nil {
+			return "", statusConditionManager.TopicsNotPresentOrInvalidErr([]string{topicName}, err)
+		}
 
 		topic, err := kafka.CreateTopicIfDoesntExist(kafkaClusterAdminClient, logger, topicName, topicConfig)
 		if err != nil {
@@ -503,7 +505,11 @@ func (r *Reconciler) finalizeNonExternalBrokerTopic(broker *eventing.Broker, sec
 	}
 	defer kafkaClusterAdminClient.Close()
 
-	topic, err := kafka.DeleteTopic(kafkaClusterAdminClient, kafka.BrokerTopic(TopicPrefix, broker))
+	topicName, err := r.KafkaFeatureFlags.ExecuteBrokersTopicTemplate(broker.ObjectMeta)
+	if err != nil {
+		return err
+	}
+	topic, err := kafka.DeleteTopic(kafkaClusterAdminClient, topicName)
 	if err != nil {
 		return err
 	}
--- a/control-plane/pkg/reconciler/broker/broker_test.go
+++ b/control-plane/pkg/reconciler/broker/broker_test.go
@@ -37,11 +37,13 @@ import (
 	"knative.dev/eventing-kafka-broker/control-plane/pkg/contract"
 
 	"github.com/Shopify/sarama"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	clientgotesting "k8s.io/client-go/testing"
+	apisconfig "knative.dev/eventing-kafka-broker/control-plane/pkg/apis/config"
 	eventing "knative.dev/eventing/pkg/apis/eventing/v1"
 	"knative.dev/eventing/pkg/apis/feature"
 	"knative.dev/eventing/pkg/eventingtls/eventingtlstesting"
@@ -98,8 +100,9 @@ var (
 	createTopicError = fmt.Errorf("failed to create topic")
 	deleteTopicError = fmt.Errorf("failed to delete topic")
 
-	linear      = eventingduck.BackoffPolicyLinear
-	exponential = eventingduck.BackoffPolicyExponential
+	linear            = eventingduck.BackoffPolicyLinear
+	exponential       = eventingduck.BackoffPolicyExponential
+	kafkaFeatureFlags = apisconfig.DefaultFeaturesConfig()
 )
 
 var DefaultEnv = &config.Env{
@@ -2789,7 +2792,8 @@ func useTable(t *testing.T, table TableTest, env *config.Env) {
 			expectedTopicDetail = td.(sarama.TopicDetail)
 		}
 
-		expectedTopicName := fmt.Sprintf("%s%s-%s", TopicPrefix, BrokerNamespace, BrokerName)
+		expectedTopicName, err := kafkaFeatureFlags.ExecuteBrokersTopicTemplate(metav1.ObjectMeta{Namespace: BrokerNamespace, Name: BrokerName})
+		require.NoError(t, err, "Failed to create broker topic name from feature flags")
 		if t, ok := row.OtherTestData[externalTopic]; ok {
 			expectedTopicName = t.(string)
 		}
@@ -2830,9 +2834,10 @@ func useTable(t *testing.T, table TableTest, env *config.Env) {
 					T:                                      t,
 				}, nil
 			},
-			Env:     env,
-			Prober:  proberMock,
-			Counter: counter.NewExpiringCounter(ctx),
+			Env:               env,
+			Prober:            proberMock,
+			Counter:           counter.NewExpiringCounter(ctx),
+			KafkaFeatureFlags: apisconfig.DefaultFeaturesConfig(),
 		}
 
 		reconciler.Tracker = &FakeTracker{}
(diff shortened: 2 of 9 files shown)


Cali0707:
> How do we make sure that if I change the template while I have existing brokers/channels, that change is not applied to existing objects?

Good catch! I hadn't considered that. Do you think that checking if there is a `broker.Status.Annotations[kafka.TopicAnnotation]` and using that if it exists would be sufficient?

Cali0707:
@pierDipi there seem to be 2 main things left to work on before this PR is ready, and I am not sure what to do for either.
1. I managed to use the `broker.Status.Annotations[kafka.TopicAnnotation]` to check for if a broker had an existing topic, as we currently set that field in the broker `reconcileKind` function. However, we don't do this for the channel (v1 or v2) so I am not sure how to check if there is an existing topic for a channel.
2. We should add unit tests to check if the templates are being applied properly for a custom template, but I am not sure how to write these tests as the `features` of the `KafkaFeatureFlags` are private, so I can't directly change them. Should we make the `features` public or is there some way to patch the resource so that the correct features will be seen as the broker/channel is reconciled?

Any ideas/input would be appreciated :)

pierDipi:
> @pierDipi there seem to be 2 main things left to work on before this PR is ready, and I am not sure what to do for either.
> 
> 1. I managed to use the `broker.Status.Annotations[kafka.TopicAnnotation]` to check for if a broker had an existing topic, as we currently set that field in the broker `reconcileKind` function. However, we don't do this for the channel (v1 or v2) so I am not sure how to check if there is an existing topic for a channel.

Can't we set it on the channel and if it's not there we assume it will use the old pattern?

> 2. We should add unit tests to check if the templates are being applied properly for a custom template, but I am not sure how to write these tests as the `features` of the `KafkaFeatureFlags` are private, so I can't directly change them. Should we make the `features` public or is there some way to patch the resource so that the correct features will be seen as the broker/channel is reconciled?
> 

Can't we make public `newFeaturesConfigFromMap` and a create `KafkaFeatureFlags` for the test?

> Any ideas/input would be appreciated :)

Cali0707:
@pierDipi is there any kind of "upgrade" code that gets called when we install a new version of ekb into a cluster? It seems like that may be a good place to add the controller channel kafka topic annotation if it exists

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -503,7 +510,11 @@ func (r *Reconciler) finalizeNonExternalBrokerTopic(broker *eventing.Broker, sec
 	}
 	defer kafkaClusterAdminClient.Close()
 
-	topic, err := kafka.DeleteTopic(kafkaClusterAdminClient, kafka.BrokerTopic(TopicPrefix, broker))
+	topicName, err := r.KafkaFeatureFlags.ExecuteBrokersTopicTemplate(broker.ObjectMeta)
+	if err != nil {
+		return err
+	}
```
pierDipi, line 516: We should take the topic name from the annotation to account for changes from when broker created to when broker is deleted.

test/rekt/features/kafka_sink.go
```diff
@@ -84,6 +76,13 @@ func BrokerWithTriggersAndKafkaSink(env environment.Environment) *feature.Featur
 	eventTrigger2.SetType(trigger2FilterType)
 
 	f := feature.NewFeatureNamed("Trigger with KafkaSink")
+	topic, err := apisconfig.DefaultFeaturesConfig().ExecuteBrokersTopicTemplate(metav1.ObjectMeta{Name: brokerName, Namespace: env.Namespace()})
+
+	f.Assert("No error creating broker topic", func(ctx context.Context, t feature.T) {
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
```
pierDipi, line 85: I'm not entirely sure about this new test here, what's the test goal here?

test/rekt/features/broker_config.go
```diff
@@ -53,11 +53,11 @@ func BrokerWithCustomReplicationFactorAndNumPartitions(env environment.Environme
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      brokerName,
-			Namespace: env.Namespace(),
-		},
+	topic, err := apisconfig.DefaultFeaturesConfig().ExecuteBrokersTopicTemplate(metav1.ObjectMeta{Name: brokerName, Namespace: env.Namespace()})
+	f.Assert("No error creating broker topic", func(ctx context.Context, t feature.T) {
+		if err != nil {
+			t.Fatal(err)
+		}
 	})
```
pierDipi, line 61: Same question here, what are we trying to test here?

control-plane/pkg/apis/config/features.go
```diff
@@ -58,7 +58,8 @@ var (
 func init() {
 	DefaultTriggersConsumerGroupTemplate, _ = template.New("triggers.consumergroup.template").Parse("knative-trigger-{{ .Namespace }}-{{ .Name }}")
 	DefaultBrokersTopicTemplate, _ = template.New("brokers.topic.template").Parse("knative-broker-{{ .Namespace }}-{{ .Name }}")
-	DefaultChannelsTopicTemplate, _ = template.New("channels.topic.template").Parse("knative-channel-{{ .Namespace }}-{{ .Name }}")
+	// This will resolve to the old naming convention, to prevent errors switching over to the new topic templates approach
+	DefaultChannelsTopicTemplate, _ = template.New("channels.topic.template").Parse("knative-messaging-kafka.{{ .Namespace }}.{{ .Name }}")
```
pierDipi, line 62: `knative-messaging-kafka.{{ .Namespace }}.{{ .Name }}` should be set here too: https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/config/eventing-kafka-broker/200-controller/100-config-kafka-features.yaml#L50C3-L50C26, we don't want to change the default values (I think, that is the reason upgrade test fails, it's not a flake)

pierDipi, line 62: I believe the 3 variables can be made package private

control-plane/pkg/reconciler/broker/broker_test.go
```diff
@@ -2135,6 +2147,81 @@ func brokerReconciliation(t *testing.T, format string, env config.Env) {
...
+							URL:  brokerAddress,
+						}),
+						WithBrokerAddessable(),
+					),
+				},
+			},
+
+			OtherTestData: map[string]interface{}{
+				useCustomTopicTemplate: true,
+			},
```
pierDipi, line 2224: Instead of passing a flag, I'd set something like:
```suggestion
			OtherTestData: map[string]interface{}{
				kafkaFeatureFlags: newKafkaFeaturesConfigFromMap(&corev1.ConfigMap{Data: ...}),
			},
```

in this way, by looking at the test declaration is clear the "given context" and the "expected context" without looking at the if magic below:
```golang
		var featureFlags *apisconfig.KafkaFeatureFlags
		var err error
		if useCustomTemplate, ok := row.OtherTestData[useCustomTopicTemplate]; ok == true && useCustomTemplate == true {
			_, example := cm.ConfigMapsFromTestFile(t, FlagsConfigName)
			featureFlags, err = apisconfig.NewFeaturesConfigFromMap(example)
			require.NoError(t, err)
		} else {
			featureFlags = apisconfig.DefaultFeaturesConfig()
		}
```

and that if will become:
```golang
		var featureFlags *apisconfig.KafkaFeatureFlags
		if v, ok := row.OtherTestData[kafkaFeatureFlags]; ok {
			featureFlags := v.(*apisconfig.KafkaFeatureFlags)
		} else {
			featureFlags = apisconfig.DefaultFeaturesConfig()
		}
```

control-plane/pkg/reconciler/broker/broker.go
pierDipi: (same for channel reconciler)

test/rekt/features/kafka_sink.go
Cali0707: I guess it doesn't necessarily make sense as a test, I was just trying to add some kind of check to the error which may be returned by `ExecuteBrokersTopicTemplate`. I'll switch it to just panic if there is an error, because there really shouldn't be an error here.

test/rekt/features/broker_config.go
Cali0707: Same answer as above

control-plane/pkg/apis/config/features.go
Cali0707: Thanks for catching this!! I'll fix it :+1:

control-plane/pkg/reconciler/broker/broker_test.go
Cali0707: Great point!!

control-plane/pkg/reconciler/broker/broker.go
Cali0707: Good point!

Cali0707:
@pierDipi I've addressed your earlier comments and fixed the broken tests. Let me know if there is anything else I should change on this PR!

control-plane/pkg/apis/config/features.go
```diff
@@ -50,32 +50,47 @@ type KafkaFeatureFlags struct {
...
-	DefaultChannelsTopicTemplate, _ = template.New("channels.topic.template").Parse("knative-channel-{{ .Namespace }}-{{ .Name }}")
+	defaultTriggersConsumerGroupTemplate, _ = template.New("triggers.consumergroup.template").Parse("knative-trigger-{{ .Namespace }}-{{ .Name }}")
+	defaultBrokersTopicTemplate, _ = template.New("brokers.topic.template").Parse("knative-broker-{{ .Namespace }}-{{ .Name }}")
+	// This will resolve to the old naming convention, to prevent errors switching over to the new topic templates approach
+	defaultChannelsTopicTemplate, _ = template.New("channels.topic.template").Parse("knative-messaging-kafka.{{ .Namespace }}.{{ .Name }}")
 }
 
 func DefaultFeaturesConfig() *KafkaFeatureFlags {
+	// we need to clone the default values when creating a new features config
+	// otherwise, when calling NewFeaturesConfigFromMap, we will overwrite the default variables
```
pierDipi, line 67: I believe a cleaner (and arguably safer) alternative is to have all templates variables like `features.TriggersConsumerGroupTemplate` to be non-pointers and pass them to `asTemplate` with:

```golang
asTemplate("triggers.consumergroup.template", &nc.features.TriggersConsumerGroupTemplate)
```

Cali0707: Oh that is a better solution! I'll implement this now

control-plane/pkg/reconciler/broker/controller.go
```diff
@@ -73,6 +75,7 @@ func NewController(ctx context.Context, watcher configmap.Watcher, env *config.E
 		ConfigMapLister:            configmapInformer.Lister(),
 		Env:                        env,
 		Counter:                    counter.NewExpiringCounter(ctx),
+		KafkaFeatureFlags:          featureFlags,
```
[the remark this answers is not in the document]
pierDipi: See here on how we do it for the trigger reconciler: https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/control-plane/pkg/reconciler/trigger/controller.go#L191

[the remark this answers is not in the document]
Cali0707: @pierDipi good catch! I've now fixed this and verified that it works for both brokers and channels.

pierDipi (approved):
/lgtm
/approve

PR #3171, opened 2023-06-26
Add spotless support

### LINKED ISSUE #3166: Java Code Formatter
**Problem**
I was working on a Java ticket these days, and found that the format of our java codebase sometime is not consistent. 
Therefore, I am trying to propose a Java formatter for the team.

Knative developers

**Exit Criteria**
- The formatter is easy to use for developers
- The CI tests should be configured as well. 
https://github.com/knative-sandbox/eventing-kafka-broker/blob/93b0ecb56d9550f7cff6f0089073e935fe2a3cd3/hack/update-codegen.sh#L80-L83

**Time Estimate (optional):** 
1

The fotmatter I am proposing is called [Spotless](https://github.com/diffplug/spotless).

The Pros for choosing Spotless:
- Just run mvn spotless:apply does magic
- It is widely used, and actively been maintained!
- Can be integrated with CI testing
- It supports [many different](https://github.com/diffplug/spotless/tree/main/plugin-maven#google-java-format) formatting policy (google java style, eclipse jdt...)
- Support with IntelliJ IDEA and VSCode

Leo6Leo:
Fixes #3166 

## Proposed Changes

- Added the support for Spotless
- Automate the formatting by integrating this with update-codegen.sh
- Add the format check to the CI testing set

**Release Note**

```release-note
Added Spotless as the Java formatter support
```

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java
@@ -15,6 +15,8 @@
  */
 package dev.knative.eventing.kafka.broker.core.reconciler.impl;
 
+import static dev.knative.eventing.kafka.broker.core.utils.Logging.keyValue;
+
 import dev.knative.eventing.kafka.broker.contract.DataPlaneContract;
 import dev.knative.eventing.kafka.broker.core.reconciler.EgressReconcilerListener;
 import dev.knative.eventing.kafka.broker.core.reconciler.IngressReconcilerListener;
@@ -22,9 +24,6 @@ import dev.knative.eventing.kafka.broker.core.reconciler.ResourcesReconciler;
 import dev.knative.eventing.kafka.broker.core.utils.CollectionsUtils;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


debasishbsws:
The `google.java.format` uses 2 spaces as indentation by default for Java code, which I don't like personally. Is it possible to overwrite it?
Or we can use `palantir-java-format` which is a fork of `google.java.format` with some better features IG, [example](https://github.com/palantir/palantir-java-format#motivation--examples).

Let's see what others have to say about this.

pierDipi:
> @pierDipi Currently, the maven version on my local machine is 3.8.6, but on GitHub the maven version is 3.8.3. I am wondering that should we bump up the maven version, or I downgrade to 3.8.3 instead?
> 
> #1301

What's the problem with that?

pierDipi:
> The `google.java.format` uses 2 spaces as indentation by default for Java code, which I don't like personally. Is it possible to overwrite it? Or we can use `palantir-java-format` which is a fork of `google.java.format` with some better features IG, [example](https://github.com/palantir/palantir-java-format#motivation--examples). It may also reduce code changes in this PR as well.
> 
> Let's see what others have to say about this.

Regarding the format, I don't have strong opinions, I just like a consistent style, the shared link does make valid arguments in favor of the Palantir format, up to you all on which format you think is the best for our project

Cali0707:
> > The `google.java.format` uses 2 spaces as indentation by default for Java code, which I don't like personally. Is it possible to overwrite it? Or we can use `palantir-java-format` which is a fork of `google.java.format` with some better features IG, [example](https://github.com/palantir/palantir-java-format#motivation--examples). It may also reduce code changes in this PR as well.
> > Let's see what others have to say about this.
> 
> Palantir's format looks more organized to me as well! I have updated the pom.xml

+1 on Palantir's formatting looking more organized

pierDipi:
> > @pierDipi The formatter check runs successfully and report no format error on my laptop, but it failed on GitHub Action. I tried to debug, and found the difference in the maven version. So I am guessing that's the problem.
> 
> Heyy @pierDipi , I spent quite a long time on debugging this, and I think might need some of your help face_exhaling face_exhaling face_exhaling
> 
> ## The problem
> 1. As quoted above
> 2. The [error log](https://github.com/knative-sandbox/eventing-kafka-broker/actions/runs/5403789607/jobs/9817092805?pr=3171#step:5:3846) from GitHub Action shows that the error appear in the file `src/test/java/dev/knative/eventing/kafka/broker/core/utils/BaseEnvTest.java` com.palantir.javaformat.java.FormatterException: 25:68: error: illegal start of expression -> [Help 1]
>    [private static final Function<String, String> provider = s -> switch (s) {](https://github.com/Leo6Leo/eventing-kafka-broker/blob/41484617b5c152ea4700b8a12f34eeb3cb412ce8/data-plane/core/src/test/java/dev/knative/eventing/kafka/broker/core/utils/BaseEnvTest.java#L25)
> 
> which means the `switch` statement is causing the problem.
> 
> ## I tried
> 1. People have had the same issue before, but is has been fixed by the community
>    [Exception formatting switch expressions google/google-java-format#477](https://github.com/google/google-java-format/issues/477)
>    I have double checked that this fix has been merged.
>    I tried with both `google-java-format` and `palantir-java-format`, but they reported the same error
> 2. I asked Calum to run the spotless command on his laptop, and everything works as expected.
> 3. Verified all the versions are consistent (Spotless version,`google-java-format` and `palantir-java-format` version, Java version, maven version). I downgraded maven on my laptop to 3.8.3, still no luck

We cannot run java commands in GH actions, that is also why there is this check: https://github.com/knative-sandbox/eventing-kafka-broker/blob/3c98b7c083fdcc8a8c0bd127138506de422c7924/hack/update-codegen.sh#L77-L84, so we need to place the new command inside the `if`

pierDipi:
> @pierDipi If we move the mvn within the if statement, then how can the GitHub action check the formatting is actually correct? And I remember that you said we just need to make sure that hack/verify-codegen.sh is taking into account all the Java code, which means we don't need the separate CI test for formatting check.

It will be verified by the build-tests in Prow

pierDipi:
*I believe* by changing the code to do:
- apply on codegen
- changing this line to cover the entire data plane directory (ie only `/data-plane`) https://github.com/knative-sandbox/eventing-kafka-broker/blob/4dd4f24c4f4f653d8d364d5ff4516195bdb330e5/hack/verify-codegen.sh#L41

should give us what we want without the exclude feature

pierDipi:
> > _I believe_ by changing the code to do:
> > 
> > * apply on update-codegen.sh
> > * changing this line to cover the entire data plane directory (ie only `/data-plane`) https://github.com/knative-sandbox/eventing-kafka-broker/blob/4dd4f24c4f4f653d8d364d5ff4516195bdb330e5/hack/verify-codegen.sh#L41
> > 
> > should give us what we want without the exclude feature
> 
> @pierDipi Then that will apply the format change to the auto-generated files?

what's the problem with that?

pierDipi:
> > `src/main/java/dev/knative/eventing/kafka/broker/contract/DataPlaneContract.java` is a generated file, so I think we should exclude it
> 
> @pierDipi Because you mentioned this before, and we should not change the auto-generated files

yeah, I thought you were doing `apply` in `hack/update-codegen.sh` and it was still failing, so I amend that comment!

pierDipi:
Basically, `hack/update-codegen.sh` should setup the code base the way we want it to be and verify-codegen.sh should do the diff on every file/dir `hack/update-codegen.sh` may change

debasishbsws:
I think the build-tests is failing because you have added a new module and that should be reflected in `/data-plane/THIRD-PARTY.txt`
Run `./hack/update-codegen.sh` And commit the modified `THIRD-PARTY.txt` file

pierDipi:
This patch might fix the problem:
```patch
diff --git a/hack/update-codegen.sh b/hack/update-codegen.sh
index af3b8106a..d77b1000b 100755
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -77,14 +77,13 @@ group "Update deps post-codegen"
 if ! ${GITHUB_ACTIONS:-false}; then
   "${REPO_ROOT_DIR}"/hack/generate-proto.sh
 
+  pushd "${REPO_ROOT_DIR}/data-plane"
   # Update Java third party file
-  pushd data-plane
   ./mvnw -Dlicense.outputDirectory=. license:aggregate-add-third-party
-  popd
-
   # Run maven command to apply spotless formatting
-  pushd "${REPO_ROOT_DIR}/data-plane"
-  mvn spotless:apply
+  ./mvnw spotless:apply
+  # Clean target directories
+  ./mvnw clean
   popd
 fi
 
...skipping...
+  ./mvnw clean
   popd
 fi
 
diff --git a/hack/verify-codegen.sh b/hack/verify-codegen.sh
index 1286dc245..2766c5a71 100755
--- a/hack/verify-codegen.sh
+++ b/hack/verify-codegen.sh
@@ -38,7 +38,13 @@ mkdir -p "${TMP_DIFFROOT}"
 DIRS=(
   "/go.mod"
   "/go.sum"
-  "/data-plane"
+  "/data-plane/contract"
+  "/data-plane/core"
+  "/data-plane/dispatcher"
+  "/data-plane/dispatcher-vertx"
+  "/data-plane/receiver"
+  "/data-plane/receiver-vertx"
+  "/data-plane/tests"
   "/control-plane/pkg/core/config"
   "/control-plane/pkg/apis"
   "/control-plane/pkg/client"
```

pierDipi:
Because the error says that the target directory and binary files are different so we need to avoid comparing those

hack/update-codegen.sh
```diff
@@ -36,58 +36,63 @@ group "Kubernetes Codegen"
...
 
   # Update Java third party file
   pushd data-plane
   ./mvnw -Dlicense.outputDirectory=. license:aggregate-add-third-party
   popd
+
+  # Run maven command to apply spotless formatting
+  pushd "${REPO_ROOT_DIR}/data-plane"
+  mvn spotless:apply
+  popd
```
pierDipi, line 88: nit: why entering the dir and existing it and then entering it again to then exit it?

hack/update-codegen.sh
```diff
@@ -36,58 +36,63 @@ group "Kubernetes Codegen"
...
+  "${REPO_ROOT_DIR}"/hack/generate-proto.sh
 
   # Update Java third party file
   pushd data-plane
   ./mvnw -Dlicense.outputDirectory=. license:aggregate-add-third-party
   popd
+
+  # Run maven command to apply spotless formatting
+  pushd "${REPO_ROOT_DIR}/data-plane"
+  mvn spotless:apply
```
pierDipi, line 87: use maven wrapper
```suggestion
  ./mvnw spotless:apply
```

PR #3178, opened 2023-06-27
Create composite prober

### LINKED ISSUE #3176: Create a composite prober
**Problem**
Per discussion [here](https://github.com/knative-sandbox/eventing-kafka-broker/issues/3141#issuecomment-1609786438) there is a need to have a composite prober, which can take an `Addressable`, and probe all of the addresses using the appropriate scheme (i.e. the prober with TLS for an https address)

**Exit Criteria**
A set of unit tests where the composite prober uses the correct prober to probe different pods.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve? 1-2

Cali0707:
Fixes #3176 

## Proposed Changes

- Update the `Prober` interface to take in a slice of `Addressable`s instead of a single `Addressable` so that a composite prober still conforms to the interface.
- Create a composite prober that will iterate through all addresses provided and probe all addresses according to their scheme.
- Create tests for the composite prober to ensure it uses the correct prober for the correct scheme

### DIFF
--- a/control-plane/pkg/prober/async_prober.go
+++ b/control-plane/pkg/prober/async_prober.go
@@ -73,12 +73,7 @@ func NewAsyncWithTLS(ctx context.Context, port string, IPsLister IPsLister, enqu
 	return NewAsync(ctx, newClient, port, IPsLister, enqueue), nil
 }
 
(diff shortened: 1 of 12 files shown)


pierDipi (approved):
/lgtm
/approve

PR #3183, opened 2023-06-28
Bump Quarkus to 3.1.2 Final 

Leo6Leo:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


pierDipi:
> @pierDipi Are there anything I might need to pay attention to when bumping up the packages?
> 

I don't have a rule of thumb I follow, it really depends on what is failing and why

I also check release notes of the direct and transitive dependencies, looking for troubleshooting hints in case the root cause isn't clear.

> And above are the failing tests that I found. I can work on trying to solve them if I am on the right track.

pierDipi:
Thanks for looking into this @Leo6Leo!

pierDipi:
@Leo6Leo maybe we can try to force use `4.9.2` by explicitly adding that dependency version and excluding the one coming from Quarkus?

pierDipi:
> @pierDipi I overwrite the dependency of antlr in pom.xml, and it works!
> 
> I fixed all the tests by upgrading the related dependencies. So right now, upgrading to 3.1.2 has no problems.
> 
> But as you suggested in the slack channel: to upgrade quarkus to 3.2.0.
> 
> This is causing some problems, and I have spent few hours debugging that and hasn't see any progress yet. So I would say we upgrade to 3.1.2 first because that is already working, what do you guys think? @pierDipi @creydr

Sounds good to me to bump to 3.2 in a follow up PR

data-plane/pom.xml
```diff
@@ -577,6 +589,11 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.antlr</groupId>
+        <artifactId>antlr4-maven-plugin</artifactId>
+        <version>${antlr.version}</version><!-- keep it aligned with the runtime bom-->
```
pierDipi, line 595: The comment is not really telling what "runtime" we're referring to but also I'd place in on the version properties more than here

pierDipi (approved):
Nice!

/lgtm
/approve

PR #3184, opened 2023-06-29
Send namespace header in MT components

pierDipi:
When running MT components [1] in mesh mode with Istio, 
we lose the ability to define fine grained policies since we
don't know the resource namespace that originated such
request, therefore, by having a `Kn-Namespace` header, 
in mesh mode, users case define fine-grained policies and
isolate namespaces.

[1] KafkaSource, Kafka Broker, and KafkaChannel

Related to https://github.com/knative/eventing/pull/7048

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
@@ -176,7 +176,9 @@ public final class WebClientCloudEventSender implements CloudEventSender {
     return VertxMessageFactory
       .createWriter(client.postAbs(target)
         .timeout(this.consumerVerticleContext.getEgressConfig().getTimeout() <= 0 ? DEFAULT_TIMEOUT_MS : this.consumerVerticleContext.getEgressConfig().getTimeout())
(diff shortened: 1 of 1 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
```diff
@@ -176,7 +176,9 @@ private Future<?> send(final CloudEvent event, final Promise<HttpResponse<Buffer
     return VertxMessageFactory
       .createWriter(client.postAbs(target)
         .timeout(this.consumerVerticleContext.getEgressConfig().getTimeout() <= 0 ? DEFAULT_TIMEOUT_MS : this.consumerVerticleContext.getEgressConfig().getTimeout())
-        .putHeader("Prefer", "reply"))
+        .putHeader("Prefer", "reply")
+        .putHeader("Kn-Namespace", this.consumerVerticleContext.getEgress().getReference().getNamespace())
```
matzew, line 180: so we set it all the time.

but benefit from it on the "mesh case", since we can use that for filtering/tweaking ?

pierDipi: Yes

Issue #2928, opened 2023-01-26
Dataplane migration for Apache Kafka communications: From Vert.x to Project Loom


matzew:
**Problem**

The Knative Broker's data-plane communication with Apache Kafka for consuming and producing records is currently done via the Vertx Kafka client library ([1]). The libary is basically a wrapper for communications with Apache Kafka inside of the vertx threading model.The project requires an evaluation the OpenJDK 19 "Project Loom" ([2], [3]) itself and how to leverage it for efficient communications with an Apache Kafka cluster. The goal of the project is to migrate the usage of vertx for any communications with Apache Kafka to pure OpenJDK 19's Loom, and reduce the number of 3rd party frameworks, such as vertx for Apache Kafka communication. 

**Exit Criteria**

A Knative Kafka Broker data plane implementation working on Java 19, leveraging the Project Loom APIs while communicating with Apache Kafka. No usage of the vertx-kafka-client, since all Kafka comms are done via Loom.

**Additional context (optional)**

[1] https://github.com/vert-x3/vertx-kafka-client
[2] https://wiki.openjdk.org/display/loom
[3] https://openjdk.org/projects/jdk/19/

ppatierno:
I am not sure what do you mean by `all Kafka communications are done via Loom`.

Regarding moving away from the Vert.x Kafka client, this is something we have already done in our Strimzi HTTP bridge, moving to the usage of `CompletableFuture`(s) to run the async Kafka communication.
Following a couple of PRs where it was made replacing the producer (called source endpoint) and consumer (called sink endpoint):

https://github.com/strimzi/strimzi-kafka-bridge/pull/724
https://github.com/strimzi/strimzi-kafka-bridge/pull/734

PR #3192, opened 2023-07-04
Remove the use of `KafkaProducer` from the dispatcher and use `ReactiveKafkaProducer` instead

### LINKED ISSUE #3191: The base dispatcher module still use `vertx.kafka.client` lib
**Describe the bug**
In the progress of #2928 the base dispatcher dependency on vertx kafka client lib has to remove.
The PR #3177 migrate the base dispatcher to use the custom ReactiveKafkaConsumer. But the KafkaProducer is still there.

**Exit Criteria**
The base dispatcher should be out of vertx kafka dependency and use custom Reactive interfaces instead.

debasishbsws:
Fixes #3191 and progress in #2928

## Proposed Changes

- Create new Reactive interfaces in the base module
- Implement them in the specialised module and pass the factory object
- Change necessary to use that new reactive interface  

Updates
- Move the Reactive interfaces to the core module
- For producer use the same Impl of the reciever


debasishbsws:
This commit history is not what I expect here, But I get some using rebase so I have use merge instead.😅

debasishbsws:
@pierDipi Would You take a look into this.

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/ReactiveKafkaProducer.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package dev.knative.eventing.kafka.broker.dispatcher;
+
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
+
+import io.vertx.core.Future;
+
+public interface ReactiveKafkaProducer<K, V> {
```
pierDipi, line 24: Can't we use the existing interface from the receiver?

debasishbsws: I have also thought about that and we can do that as well, but then I thought that by doing so, the dispatcher-vertx module will also get depend on the receiver modules, and I decided not to do that.
But now we definitely have redundancy of code.
We can definitely use the existing interface.

pierDipi: I would go with reducing duplication and depend on the receiver module, eventually, we can think about moving those in core module

pierDipi: What's the downside of introducing a dependency on the receiver?

debasishbsws: No, I don't think there is any downside, but as there are no dependency already so i just thought not to introduce one.

Also, I think it will be better if we move all the interfaces to the core module.

debasishbsws:
New PR #3206. To much merge conflicts in this PR

PR #3193, opened 2023-07-05
Added CA certs to web clients

### LINKED ISSUE #3129: Eventing TLS: Support sending events to TLS endpoints
**Problem**

Once we have our CA certs propagated to the data plane after https://github.com/knative-sandbox/eventing-kafka-broker/issues/3118, we can enable the Vertx webclient to send events to TLS endpoints using the provided CA certs.

[WebClientCloudEventSender](
https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
) is responsible for sending events to a specific URL

**Exit Criteria**

- WebClientCloudEventSender

**Time Estimate (optional):** 

- https://vertx.io/docs/vertx-web-client/java/#_using_https

Cali0707:
Fixes #3129 

## Proposed Changes
To send an http request with TLS in vert.x, you need to call `postAbs()` (which we already do), and have the `WebClient` configured to trust the CA certificate for the destination. So, the only necessary change is:
- When creating an instance of `WebClient`, pass in the appropriate CA certificate if it is not null/empty

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleBuilder.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleBuilder.java
@@ -46,9 +46,12 @@ import dev.knative.eventing.kafka.broker.dispatcher.impl.http.WebClientCloudEven
 import io.cloudevents.CloudEvent;
 import io.vertx.core.Future;
 import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.net.PemTrustOptions;
 import io.vertx.core.tracing.TracingPolicy;
 import io.vertx.ext.web.client.WebClient;
+import io.vertx.ext.web.client.WebClientOptions;
 import io.vertx.kafka.client.common.KafkaClientOptions;
 import io.vertx.kafka.client.common.tracing.ConsumerTracer;
 import io.vertx.kafka.client.producer.KafkaProducer;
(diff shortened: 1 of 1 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleBuilder.java
```diff
@@ -211,11 +214,29 @@ private static Filter getFilter(DataPlaneContract.DialectedFilter filter) {
     };
   }
 
+  private WebClientOptions createWebClientOptionsFromCACerts(String CACerts) {
+    if (!CACerts.isEmpty()) {
+      return consumerVerticleContext
+        .getWebClientOptions()
+        .setPemTrustOptions(
```
pierDipi, line 221: I believe we're modifying a shared object (WebClientOptions) and by doing that we're only configuring the client to trust whatever CA Certs we configure last, a more safe approach would be to consider WebClientOptions as an immutable object, so I'd copy/clone it (there should a constructor for that) and _then_ set the trust options.

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleBuilder.java
```diff
@@ -211,11 +214,29 @@ private static Filter getFilter(DataPlaneContract.DialectedFilter filter) {
     };
   }
 
+  private WebClientOptions createWebClientOptionsFromCACerts(String CACerts) {
+    if (!CACerts.isEmpty()) {
```
pierDipi, line 218: Might `CACerts` ever be `null`?

pierDipi (commented):
To support CA certs rotation and changes, we need to add them to the comparison here: https://github.com/knative-sandbox/eventing-kafka-broker/blob/4dd4f24c4f4f653d8d364d5ff4516195bdb330e5/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImpl.java#L234-L254

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/ConsumerVerticleBuilder.java
Cali0707: I'm not sure? When I added a !=null check IntelliJ said that it would always evaluate to true so I guessed it wouldn't but I am not very familiar with this part of the codebase.

pierDipi: Sounds good, just wanted to make sure we've considered that case

Cali0707: Just in case one of them can be null (I only checked for one of the CA certs variables before I settled on the approach of having a private method for this), I added a null check

PR #3200, opened 2023-07-10
Updated comments explaining how the async prober works

Cali0707:
Per conversation with @pierDipi the async prober code can be confusing when it asynchronously aggregates the status of the different IPs for each pod.

## Proposed Changes

- Add more context to the comments/update outdated comments to provide more context in the async prober


pierDipi (approved):
/lgtm
/approve

PR #3195, opened 2023-07-07
Simplify `ConsumerRebalanceListener` implementation in the dispatcher.

debasishbsws:
While working on #2928, In PR #3177, to remove the use of `onPartitionRevokeHandler(...)` I use the `ConsumerRebalanceListener`.
But in order to do that I edit the context in `ConsumerVerticleBuilder` which should not be done.
And it also made the code a little complex and inconsistent.

## Proposed Changes

- Collection of `PartitionRevokeHandler` is in `ConsumerVerticle` now
- `getConsumerRebalanceListener()` returning a new ConsumerRebalaceListener with all the Handler Function.
- Revome `ConsumerRebalanceListener` from `ConsumerVerticleContext`.

PR #3207, opened 2023-07-11
Remove `ConsumerRebalanceListener` from `ConsumerVerticleContext`

debasishbsws:
A clean conflict-free PR of #3195

## Proposed Changes

- Put the `ConsumerRebalanceListener` into `ConsumerVerticle`

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/ConsumerVerticle.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/ConsumerVerticle.java
@@ -26,7 +26,6 @@ import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 import java.util.Objects;
 import java.util.function.BiFunction;
-
 import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/ConsumerVerticle.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/ConsumerVerticle.java
@@ -26,6 +26,8 @@ import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 import java.util.Objects;
 import java.util.function.BiFunction;
+
+import org.apache.kafka.clients.consumer.ConsumerRebalanceListener;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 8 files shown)


debasishbsws:
@pierDipi 
"Did you consider passing to the consumer verticle the rebalance listener instead of passing all revoke handlers?

 consumerVerticle.setRebalanceListener(createRebalanceListener(partitionRevokedHandlers));
That would decouple the consumer verticle from the rebalance logic, in my opinion, creating and coordinating the build of the rebalance listener is more a responsibility of the builder rather than a responsibility of the consumer verticle (consumer verticle is responsible for the lifecycle of the consumer)"(https://github.com/knative-sandbox/eventing-kafka-broker/pull/3195/files#r1259704553)

In this approach, some tests are failing as the consumerVerticleListener is not getting initialised.
I can put a check like
```java
protected ConsumerRebalanceListener getConsumerRebalanceListener() {
        if(consumerRebalanceListener != null){
            return consumerRebalanceListener;
        }
        return new ConsumerRebalanceListener() {
            @Override
            public void onPartitionsRevoked(final Collection<TopicPartition> partitions) {
                getPartitionRevokedHandler().partitionRevoked(partitions).onComplete(
                        r -> logger.info("Partitions revoked {}", consumerVerticleContext.getLoggingKeyValue()));
            }

            @Override
            public void onPartitionsAssigned(final Collection<TopicPartition> partitions) {
                logger.info("Partitions assigned {}", consumerVerticleContext.getLoggingKeyValue());
            }
        };
    }
```
But I think it is like same as Building the ConsumerRebalanceListener inside it.
OR
I can create a new ConsumerRebalanceListenerImpl and move the logic there as I Previously did in Old PR old commit https://github.com/knative-sandbox/eventing-kafka-broker/pull/3195/commits/aa9b1397f78e8656bf93c32efb8ea19f897447ab

pierDipi:
> 
> In this approach, some tests are failing as the consumerVerticleListener is not getting initialised. I can put a check like
> 
> ```java
> protected ConsumerRebalanceListener getConsumerRebalanceListener() {
>         if(consumerRebalanceListener != null){
>             return consumerRebalanceListener;
>         }
>         return new ConsumerRebalanceListener() {
>             @Override
>             public void onPartitionsRevoked(final Collection<TopicPartition> partitions) {
>                 getPartitionRevokedHandler().partitionRevoked(partitions).onComplete(
>                         r -> logger.info("Partitions revoked {}", consumerVerticleContext.getLoggingKeyValue()));
>             }
> 
>             @Override
>             public void onPartitionsAssigned(final Collection<TopicPartition> partitions) {
>                 logger.info("Partitions assigned {}", consumerVerticleContext.getLoggingKeyValue());
>             }
>         };
>     }
> ```
> 
> But I think it is like same as Building the ConsumerRebalanceListener inside it. OR I can create a new ConsumerRebalanceListenerImpl and move the logic there to decouple the consumer verticle from the rebalance logic as I Previously did in Old PR old commit [aa9b139](https://github.com/knative-sandbox/eventing-kafka-broker/commit/aa9b1397f78e8656bf93c32efb8ea19f897447ab)

I didn't get what's the problem, can you expand a bit more?

debasishbsws:
I have moved the ConsumerRebalanceLitener creation Logic out from the ConsumerVerticle and use `consumerVerticle.setRebalanceListener(... );` instead. But now ConsumerRebalanceLitener is not getting initialised while running some tests like OrderedConsumerVeticleTest, UnordereConsumerVerticleTest.
So I was asking that to resolve that I can put a condition like if the ConsumerRebalanceLitener is null then create a new one and return it to make sure we never get a null.

debasishbsws:
Oh wait I can put
```java
consumerVerticle.setRebalanceListener(new ConsumerRebalanceListener() {....})
```
into the test.

PR #3211, opened 2023-07-13
Adding the Loom base modules with empty files.

debasishbsws:
Progress in #2928 

## Proposed Changes

- Added both loom-based modules with Java 19 as the latest Maven shade plugin does not support Java20 till now.
- Lower the Java version to 19 in tests as Every version has its exact preview feature and does not work in other versions.
Updates in latter comment

### DIFF
--- a/.github/workflows/knative-java-test.yaml
+++ b/.github/workflows/knative-java-test.yaml
@@ -28,7 +28,7 @@ jobs:
     name: Java Unit Tests
     strategy:
       matrix:
-        java-version: [ 20 ]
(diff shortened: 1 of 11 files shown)


debasishbsws:
Removed the shade plugin as it is not needed at least for now. and use `docker.io/library/eclipse-temurin:20-jre`

debasishbsws:
@pierDipi It's ready to review.

pierDipi (approved):
/lgtm
/approve

PR #3215, opened 2023-07-15
adding Receiver loom based module Implementation.

debasishbsws:
Progress in #2928 

## Proposed Changes

- Loom-based module implementation
- KafkaProducer unit Test
- modify the tests module DataPlaneTest to be runnable  for both vertx and loom modules

**Release Note**

```release-note

```

**Docs**


debasishbsws:
Need to add the tracing and span for the producer

pierDipi:
> I have found the problem with DataPlaneTest I was having and fixed it. Now I am trying to refactor the DataPlaneTest and trying to make it abstract for both Implementations.
> 

Great job @debasishbsws !

> Is it okay to make the methods non-static and use `@BeforeEach` and `@AfterEach`? I haven't found any reason to make them static. But just wanted to confirm. /cc @pierDipi



Yes, I don't see why not!

debasishbsws:
@pierDipi I think this one is ready for review, please take a look.

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,132 @@
...
+        if (isClosed.get()) {
+            promise.fail("Producer is closed");
+        } else {
+            eventQueue.add(new RecordPromise(record, promise));
+        }
+        return promise.future();
+    }
+
+    private void sendFromQueue() {
+        while (true) {
```
debasishbsws, line 71: Did after close() sendFromQueue VThread will get end or it will get preempted in the background, May need to change the condition.

pierDipi: In addition, we need to interrupt the thread on close because it might be waiting on `take()` and it will never exit

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,132 @@
...
+            }
+        }
+    }
+
+    @Override
+    public Future<Void> close() {
+        Promise<Void> promise = Promise.promise();
+        this.isClosed.set(true);
+        Thread.ofVirtual().start(() -> {
+            while (!eventQueue.isEmpty()) {}
```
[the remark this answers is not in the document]
debasishbsws: Yes, we can do that. 

and for interrupting the sendFromQueue thread on close I am using `while (!isClosed.get() || !eventQueue.isEmpty())`

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
debasishbsws: > In addition, we need to interrupt the thread on close because it might be waiting on take() and it will never exit

Oh I see, it might not work when the producer is not closed and we are waiting on queue.take() and without any incoming event we are closing.

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,132 @@
...
+        if (isClosed.get()) {
+            promise.fail("Producer is closed");
+        } else {
+            eventQueue.add(new RecordPromise(record, promise));
+        }
+        return promise.future();
+    }
+
+    private void sendFromQueue() {
+        while (!isClosed.get() || !eventQueue.isEmpty()) {
```
pierDipi, line 71: Both conditions must be true, the reason being:
- when close, we want to process the queue before closing, and
- we don't want to stop processing before close is called

```suggestion
        while (!isClosed.get() && !eventQueue.isEmpty()) {
```

debasishbsws: > - when close, we want to process the queue before closing, and
> - we don't want to stop processing before close is called

Yes, that is why I have used `||` as when `isClosed` is `true` and `!isClosed.get()` is false but there are elements in the eventQueue we want to continue the loop.
But still it might not work for the same reason
> when the producer is not closed and we are waiting on `queue.take()` and without any incoming event we are calling close. so the execution is never going to `while(condition..)` 

I have changed the condition to use `Thread.interrupt()` which we are calling from the `close` method.
I think it will work and also simple.

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
pierDipi, line 1: There are many annotations saying missing coverage, can we add more tests?

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,134 @@
...
+                        startedSpan.finish(ctx);
+                    }
+                } catch (Exception e) {
+                    recordPromise.getPromise().fail(e);
+                    if (startedSpan != null) {
+                        startedSpan.fail(ctx, e);
+                    }
+                }
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
pierDipi, line 90: Can we use a logger at debug level here?

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -96,8 +97,9 @@ public Future<Void> close() {
         Promise<Void> promise = Promise.promise();
         this.isClosed.set(true);
         Thread.ofVirtual().start(() -> {
-            while (!eventQueue.isEmpty()) {}
             try {
+                Thread.sleep(2000L);
```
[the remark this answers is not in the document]
debasishbsws: Oh, I thought we will consume all the remaining events in 2 sec.😅

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
debasishbsws: Yes, there are so many annotations saying missing coverage. But I am not sure what type of test I should add. Okay let me try

debasishbsws: I have added a bunch of Tests. Taking new test ideas from ChatGPT😌.

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
```diff
@@ -0,0 +1,226 @@
...
+        while (future.isComplete()) {
+            Thread.sleep(100);
+        }
+        // wait for 5 seconds to ensure the sendFromQueueThread is interrupted
+        Thread.sleep(5000L);
+        // verify the sendFromQueueThread is interrupted
+        testContext.verify(() -> {
+            assertTrue(!producer.isSendFromQueueThreadAlive());
+            testContext.completeNow();
+        });
```
debasishbsws, line 166: Hm, interesting this test is passing locally. Should I increase the wait time? Maybe it need some more time to get interrupted as it was blocked.

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,145 @@
...
+                } catch (Exception e) {
+                    recordPromise.getPromise().fail(e);
+                    if (startedSpan != null) {
+                        startedSpan.fail(ctx, e);
+                    }
+                }
+            } catch (InterruptedException e) {
+                logger.debug("Interrupted while waiting for event queue to be populated.");
+            }
+        }
```
pierDipi, line 92:
```suggestion
        }
        
        logger.debug("Background thread finished.");
```

debasishbsws: I have just changed the `sendFromQueue` function to use the callback and it is stopping now.

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,141 @@
...
+                RecordPromise recordPromise = eventQueue.take();
+                ProducerTracer.StartedSpan startedSpan =
+                        this.tracer == null ? null : this.tracer.prepareSendMessage(ctx, recordPromise.getRecord());
+                producer.send(recordPromise.getRecord(), (metadata, exception) -> {
+                    if (exception != null) {
+                        recordPromise.getPromise().fail(exception);
+                        if (startedSpan != null) {
+                            startedSpan.fail(ctx, exception);
+                        }
+                    }
```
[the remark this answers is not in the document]
debasishbsws: yup, I missed it in hurry.😅

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,146 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.slf4j.Logger;
+
+public class LoomKafkaProducer<K, V> implements ReactiveKafkaProducer<K, V> {
+
+    Logger logger = org.slf4j.LoggerFactory.getLogger(LoomKafkaProducer.class);
```
pierDipi, line 36: Can we use the shorter form with the import?

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
```diff
@@ -0,0 +1,212 @@
...
+                testContext.completeNow();
+            });
+        });
+
+        // Verify that the sent record is processed before the producer is closed
+        for (int i = 0; i < numRecords; i++) {
+            sendFutures.get(i).onComplete(ar -> {
+                testContext.verify(() -> {
+                    assertTrue(ar.succeeded());
+                    testContext.completeNow();
```
pierDipi, line 131: Calling `completeNow` multiple times doesn't do what we want here, we expect to have `numRecords + 1 (for close)` "future completions", for that we can use checkpoints:

```java

        final var checkpoints = testContext.checkpoint(numRecords + 1);
        
        testContext.verify(() -> {
                assertTrue(ar.succeeded());
                // other assertions

                checkpoints.flag(); // complete checkpoint
            });

File: data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java [hunk 11]
```diff
@@ -0,0 +1,212 @@
...
+
+        // Close the producer and wait for the event queue to be empty
+        producer.close().onComplete(ar -> {
+            testContext.verify(() -> {
+                assertTrue(ar.succeeded());
+
+                // Ensure that the eventQueue is empty after closing
+                assertTrue(producer.getEventQueueSize() == 0);
+                // Ensure that the sendFromQueueThread is not alive after closing
+                assertTrue(!producer.isSendFromQueueThreadAlive());
```
pierDipi, line 121:
```suggestion
                assertFalse(producer.isSendFromQueueThreadAlive());
```

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
```diff
@@ -0,0 +1,212 @@
...
+
+        // Verify that the sent record is processed before the producer is closed
+        for (int i = 0; i < numRecords; i++) {
+            sendFutures.get(i).onComplete(ar -> {
+                testContext.verify(() -> {
+                    assertTrue(ar.succeeded());
+                    testContext.completeNow();
+                });
+            });
+        }
```
pierDipi, line 134: Can't we set those assertions at line 110 instead of looping and aggregating futures?
`producer.send(record)` at line 110

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
```diff
@@ -0,0 +1,212 @@
...
+        // Send multiple records
+        int numRecords = 100000;
+        List<Future<RecordMetadata>> sendFutures = new ArrayList<>();
+        for (int i = 0; i < numRecords; i++) {
+            ProducerRecord<String, Integer> record = new ProducerRecord<>("test", "sequence number", i);
+            sendFutures.add(producer.send(record));
+        }
+
+        // Close the producer and wait for the event queue to be empty
+        producer.close().onComplete(ar -> {
```
pierDipi, line 114: We need to set what happens on failure with `.onFailure(testContext::failNow)` so that we get a more descriptive error message

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
```diff
@@ -0,0 +1,212 @@
...
+                assertTrue(producer.getEventQueueSize() == 0);
+                // Ensure that the sendFromQueueThread is not alive after closing
+                assertTrue(!producer.isSendFromQueueThreadAlive());
+                testContext.completeNow();
+            });
+        });
+
+        // Verify that the sent record is processed before the producer is closed
+        for (int i = 0; i < numRecords; i++) {
+            sendFutures.get(i).onComplete(ar -> {
```
pierDipi, line 128: We need to set what happens on failure with `.onFailure(testContext::failNow)` so that we get a more descriptive error message

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
```diff
@@ -0,0 +1,212 @@
...
+            });
+        });
+
+        // now sendFromQueueThread should be blocked in eventQueue.take()
+
+        // close the producer and verify the sendFromQueueThread is interrupted
+        producer.close().onComplete(ar -> {
+            testContext.verify(() -> {
+                assertTrue(ar.succeeded());
+                assertTrue(!producer.isSendFromQueueThreadAlive());
```
pierDipi, line 157:
```suggestion
                assertFalse(producer.isSendFromQueueThreadAlive());
```

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
```diff
@@ -0,0 +1,212 @@
...
+                testContext.completeNow();
+            });
```
pierDipi, line 159: Same comment above related to multiple `completeNow` calls and setting up the `onFailure` handling for  a more descriptive error message.

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
```diff
@@ -0,0 +1,212 @@
...
+        });
+
+        // Verify that the sent record is processed before the flush is completed
+        sendFuture.onComplete(ar -> {
+            testContext.verify(() -> {
+                assertTrue(ar.succeeded());
+                testContext.completeNow();
+            });
+        });
```
pierDipi, line 183: Same comment above related to multiple `completeNow` calls and setting up the `onFailure` handling for  a more descriptive error message.

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,146 @@
...
+        }
+
+        public ProducerRecord<K, V> getRecord() {
+            return record;
+        }
+
+        public Promise<RecordMetadata> getPromise() {
+            return promise;
+        }
+    }
```
pierDipi, line 150: I would rework this class to be static and defined on generic `<K, V>`

```suggestion
    private static class RecordPromise<K, V> {
        private final ProducerRecord<K, V> record;
        private final Promise<RecordMetadata> promise;

        private RecordPromise(ProducerRecord<K, V> record, Promise<RecordMetadata> promise) {
            this.record = record;
            this.promise = promise;
        }

        public ProducerRecord<K, V> getRecord() {
            return record;
        }

        public Promise<RecordMetadata> getPromise() {
            return promise;
        }
    }
```

To reduce the risk of the introduction (now and in the future) of any form of memory leak https://www.infoworld.com/article/3526554/avoid-memory-leaks-in-inner-classes.html

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
matzew, line 36: should we make this `final`?

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/tracing/kafka/TraceTags.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        @Override
+        public String value(TraceContext obj, int index) {
+            return TAGS[index].fn.apply(obj);
+        }
+    };
+
+    final String name;
+    final Function<TraceContext, String> fn;
```
matzew, line 57: I'd choose a different name vor the constant instead of just `fn`

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
matzew: that's a good call. I am also in fav. of this

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
pierDipi: and `static` as well

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,146 @@
...
+        return promise.future();
+    }
+
+    private void sendFromQueue() {
+        while (!Thread.currentThread().isInterrupted()) {
+            try {
+                RecordPromise recordPromise = eventQueue.take();
+                ProducerTracer.StartedSpan startedSpan =
+                        this.tracer == null ? null : this.tracer.prepareSendMessage(ctx, recordPromise.getRecord());
+                producer.send(recordPromise.getRecord(), (metadata, exception) -> {
```
pierDipi, line 77: Send can throw an exception and we need to handle it accordingly 

I would rework the method with something like this:
```java
  private void sendFromQueue() {
    while (!Thread.currentThread().isInterrupted()) {
      try {
        RecordPromise<K, V> recordPromise = eventQueue.take();

        ProducerTracer.StartedSpan startedSpan =
          this.tracer == null ? null : this.tracer.prepareSendMessage(ctx, recordPromise.getRecord());

        recordPromise.getPromise().future()
          .onComplete(v -> {
            if (startedSpan != null) {
              startedSpan.finish(ctx);
            }
          })
          .onFailure(cause -> {
            if (startedSpan != null) {
              startedSpan.fail(ctx, cause);
            }
          });


        try {
          producer.send(recordPromise.getRecord(), (metadata, exception) -> {
            if (exception != null) {
              recordPromise.getPromise().fail(exception);
            }
            recordPromise.getPromise().complete(metadata);
          });
        } catch (final KafkaException ex) {
          recordPromise.getPromise().fail(ex);
        }

      } catch (InterruptedException e) {
        logger.debug("Interrupted while waiting for event queue to be populated.");
      }
    }
  }
```

data-plane/receiver-loom/src/test/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducerTest.java
debasishbsws: So for all of the suggestion we will have the test something like
```java
    @Test
    public void testCloseIsWaitingForEmptyQueue(VertxTestContext testContext) {
        // Send multiple records
        int numRecords = 100000;
        final var checkpoints = testContext.checkpoint(numRecords + 1);

        for (int i = 0; i < numRecords; i++) {
            ProducerRecord<String, Integer> record = new ProducerRecord<>("test", "sequence number", i);
            producer.send(record)
                    .onSuccess(ar -> {
                        checkpoints.flag();
                    })
                    .onFailure(testContext::failNow);
        }

        // Close the producer and wait for the event queue to be empty
        producer.close()
                .onComplete(ar -> {
                    testContext.verify(() -> {
                        assertTrue(ar.succeeded());
                        assertEquals(0, producer.getEventQueueSize());
                        assertFalse(producer.isSendFromQueueThreadAlive());

                        checkpoints.flag();
                    });
                })
                .onFailure(testContext::failNow);
    }
```

pierDipi: Yes!

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/tracing/kafka/TraceTags.java
debasishbsws: I have taken this class as it is from the Vert.x tracing implementation, so I haven't changed anything. 
https://github.com/vert-x3/vertx-kafka-client/blob/master/src/main/java/io/vertx/kafka/client/common/tracing/TraceTags.java

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
debasishbsws: Done.

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -15,6 +15,162 @@
...
+        Promise<Void> promise = Promise.promise();
+        this.isClosed.set(true);
+        Thread.ofVirtual().start(() -> {
+            try {
+                while (!eventQueue.isEmpty()) {
+                    Thread.sleep(2000L);
+                }
+                sendFromQueueThread.interrupt();
+                producer.close();
+                sendFromQueueThread.join();
```
pierDipi, line 122: It seems more natural to interrupt and wait for it to terminate, what is the reason for the different order?

```suggestion
                sendFromQueueThread.join();
                producer.close();
```

debasishbsws: No, particular reason. okay I will change it.

PR #3217, opened 2023-07-18
Include io.fabric8 kubernetes-client in maven-shade-plugin to prevent ClassNotFoundException

creydr:
Currently the fabric8 Kubernetes Client loads the Client Implementation class dynamically: https://github.com/fabric8io/kubernetes-client/blob/587657dc1bdd244cd4da5352ab97cd5668081696/kubernetes-client-api/src/main/java/io/fabric8/kubernetes/client/KubernetesClientBuilder.java#L53

This can lead to a ClassNotFoundException, when using the maven-shade plugin to package an "uber jar" as the class is not directly referenced.
This PR addresses it and explicitly includes the kubernetes-client classes.

### DIFF
--- a/data-plane/dispatcher-vertx/pom.xml
+++ b/data-plane/dispatcher-vertx/pom.xml
@@ -87,6 +87,12 @@
                     <include>**</include>
                   </includes>
                 </filter>
+                <filter>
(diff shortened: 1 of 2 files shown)


pierDipi (approved):
Thanks!

/lgtm
/approve

PR #3223, opened 2023-07-21
Pass EventTypeAutoCreate enabled flag to the data plane

### LINKED ISSUE #3220: Pass EventTypeAutoCreation flag to the Data Plane
**Problem**
Before we can work on #3174 or #3181, we need to pass the EventTypeAutoCreation flag to the data plane.

**Exit Criteria**

**Time Estimate (optional):** 

Cali0707:
Fixes #3220 

## Proposed Changes

- Add the feature flag to the Ingress contract
- In each reconcile loop where we set the Ingress contract, set the feature flag


Leo6Leo:
https://github.com/Leo6Leo/eventing-kafka-broker/blob/32dfed583ceb02c4fa1581b3bffab07d6ec0e7d6/data-plane/core/src/test/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImplTest.java#L436

Should the test here also need to be modified? @Cali0707

Cali0707:
> https://github.com/Leo6Leo/eventing-kafka-broker/blob/32dfed583ceb02c4fa1581b3bffab07d6ec0e7d6/data-plane/core/src/test/java/dev/knative/eventing/kafka/broker/core/reconciler/impl/ResourcesReconcilerImplTest.java#L436
> 
> Should the test here also need to be modified? @Cali0707

@Leo6Leo I don't think so? This PR is just passing the flag into the data plane, not updating the code anywhere to use it.

@pierDipi WDYT?

pierDipi (approved):
As long as we keep in mind that we need to add tests and we do that in a follow up this is fine

/lgtm
/approve

PR #3227, opened 2023-07-25
Eventing TLS: Configure certificates to force rotate private keys

pierDipi:
As per title, without this configuration the private key is not rotated


Cali0707 (approved):
/lgtm

PR #3247, opened 2023-07-28
Created abstract FilterBenchmark class

### LINKED ISSUE #3233: New Event Filtering: Set up base benchmarking class for filters & use for attribute filter
**Problem**
To make benchmarking the Java filter implementations easier, we should set up a common filter benchmark runner classes, similar to here: https://github.com/knative/eventing/blob/248a471dc3b52b157b9dadffc288c1c49ccbcf47/pkg/eventfilter/benchmarks/common_benchmark_test.go.

To make sure this is working and set a baseline to compare other benchmarks to, we should also use this class to benchmark the attribute filter (https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/filter/AttributesFilter.java)

**Exit Criteria**
A working benchmark of the `AttributesFilter`

**Time Estimate (optional):** 

https://github.com/knative/eventing/issues/5204

Cali0707:
Fixes #3233 

## Proposed Changes

- Create an abstract class for the filter benchmarks
- Extend the abstract class to benchmark the default ExactFilter filter

### DIFF
--- a/data-plane/benchmarks/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/filter/ExactFilterBenchmark.java
+++ b/data-plane/benchmarks/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/filter/ExactFilterBenchmark.java
@@ -1,0 +1,73 @@
+package dev.knative.eventing.kafka.broker.dispatcher.impl.filter;
+
+import dev.knative.eventing.kafka.broker.dispatcher.Filter;
+import dev.knative.eventing.kafka.broker.dispatcher.impl.filter.subscriptionsapi.ExactFilter;
+import io.cloudevents.CloudEvent;
+import io.cloudevents.core.builder.CloudEventBuilder;
+import io.cloudevents.core.v1.CloudEventV1;
+
+import java.net.URI;
+import java.util.Map;
+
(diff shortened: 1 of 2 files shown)


pierDipi:
@Cali0707 did you run them? Can you share the results?

pierDipi:
```
/mvnw clean package -DskipTests -P no-release

java -jar benchmarks/target/benchmarks.jar ExactFilterBenchmarkNoMatch
```

I've noticed that there are debug log statements printed so it might be measuring that as well, ideally, we have our logging level to `INFO` which is the default value

pierDipi:
> @pierDipi to remove the debug logs, I used the command `java -Dlogback.configurationFile=benchmarks/profiler/resources/config-logging.xml -jar target/benchmarks ExactFilterBenchmark > benchmarks.txt`
> 
> This produced the output file I have attached [benchmarks.txt](https://github.com/knative-extensions/eventing-kafka-broker/files/12242835/benchmarks.txt)

Do you think we could add a little README to the benchmark module to help the next dev that needs to run such benchmarks ?

pierDipi (approved):
The different results are quite interesting, it seems something worth to investigate further 

/lgtm
/approve

PR #3254, opened 2023-08-01
Fix update-codegen.sh for update-cert-manager.sh update

pierDipi:
As per title, we're getting unwanted updates on the update deps automation PRs


Cali0707 (approved):
/lgtm

PR #3261, opened 2023-08-02
Add missing addresses field to broker finalizer

Cali0707:
The broker finalizeKind probe should have the address in the `NewProber.Addresses` field as well, because the composite prober only probes addresses in that list.

## Proposed Changes

- Add the address to the `Addresses` field


matzew (approved):
/lgtm
/approve

PR #3271, opened 2023-08-04
Add benchmark filter run script

### LINKED ISSUE #3270: New Event Filtering: Add hack/run script for java filter benchmarks
**Problem**
As we add more benchmarks for the java filter implementations and work to improve the implementation, it will become more common to want to run the benchmarks. Currently, it is difficult to run them and required multiple commands. To solve this, we should add a script which runs the benchmarks and outputs the results into a file 

**Exit Criteria**
Running `./hack/run.sh benchmark-filters` runs the benchmarks for all of the filters. Running `./hack/run.sh benchmark-filters <class_name>` runs all the filter benchmarks contained in class_name.

**Time Estimate (optional):** 

Cali0707:
Fixes #3270 

## Proposed Changes

- Add `run.sh` script in the benchmark project to run the benchmarks
- Make the `run.sh` script runnable from `hack/run.sh`
- Add documentation to `DEVELOPMENT.md` on how to run the script

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -164,6 +164,26 @@ export EVENT=alloc
 
 For more information on the profiler test, see [the profiler test doc](./data-plane/profiler/README.md).
 
+### Run Filter Benchmarks
+
+If you are building a fitler benchmark or want to benchmark the performance delta caused when changing the filters, you can run:
+
+```shell
+./hack/run.sh benchmark-filter <filter_class_name>
+```
+
+This will run the benchmarks for the class with `<filter_class_name>`. For example, if you want to run all of the Exact Filter Benchmarks, you could run:
+
+```shell
+./hack/run.sh benchmark-filter ExactFilterBenchmark
+```
+
+Alternatively, if you want to run all of the benchmarks you can run:
+
+```shell
+./hack/run.sh benchmark-filters
+```
+
 ## Code generation
 
 Sometimes, before deploying the services it's required to run our code generators, by running the following command:
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -172,7 +172,8 @@ If you are building a fitler benchmark or want to benchmark the performance delt
 ./hack/run.sh benchmark-filter <filter_class_name>
 ```
 
-This will run the benchmarks for the class with `<filter_class_name>`. For example, if you want to run all of the Exact Filter Benchmarks, you could run:
+This will run the benchmarks for the class with `<filter_class_name>`. A full list of all available classes can be seen [here](https://github.com/knative-extensions/eventing-kafka-broker/blob/main/data-plane/benchmarks/resources/filter-class-list.txt).
+For example, if you want to run all of the Exact Filter Benchmarks, you could run:
 
 ```shell
 ./hack/run.sh benchmark-filter ExactFilterBenchmark
--- a/data-plane/benchmarks/run.sh
+++ b/data-plane/benchmarks/run.sh
@@ -23,7 +23,8 @@ run_java_filter_benchmarks_for_class() {
 
   echo "Running benchmarks for ${CLASS}"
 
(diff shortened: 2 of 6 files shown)


data-plane/benchmarks/run.sh
```diff
@@ -0,0 +1,49 @@
...
+set -e
+
+run_java_filter_benchmarks_for_class() {
+  CLASS=$1
+
+  echo "Running benchmarks for ${CLASS}"
+
+  java -jar "${SCRIPT_DIR}/target/benchmarks.jar" $CLASS 2>&1 | tee "${SCRIPT_DIR}/output/${CLASS}.out.txt"
+
+  echo "Successfully ran benchmarks for ${CLASS}!\n\nThe results can be found at ${SCRIPT_DIR}/output/${CLASS}.out.txt"
```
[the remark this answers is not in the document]
Leo6Leo: If we run the benchmark for multiple times, the previous results will be overwritten.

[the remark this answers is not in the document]
Cali0707: Oh true good point!

Leo6Leo (commented):
image

@Cali0707  the file is really large lol

Cali0707:
> image
> 
> 
> 
> @Cali0707  the file is really large lol, and it is still running

This should be fixed now!

Leo6Leo (commented):
```
Running benchmarks for ExactFilterBenchmarkjjkyj
No matching benchmarks. Miss-spelled regexp?
Use EXTRA verbose mode to debug the pattern matching.
Successfully ran benchmarks for ExactFilterBenchmarkjjkyj!\n\nThe results can be found at /home/leoli/Downloads/eventing-kafka-broker/data-plane/benchmarks/output/ExactFilterBenchmarkjjkyj.1691602228.out.txt
```

When the input is wrong, the error is this. And it will run the build first as well. If the className sanity check happens at the beginning, it can save some time and resources

DEVELOPMENT.md
```diff
@@ -164,6 +164,26 @@ export EVENT=alloc
...
 
+### Run Filter Benchmarks
+
+If you are building a fitler benchmark or want to benchmark the performance delta caused when changing the filters, you can run:
+
+```shell
+./hack/run.sh benchmark-filter <filter_class_name>
+```
+
+This will run the benchmarks for the class with `<filter_class_name>`. For example, if you want to run all of the Exact Filter Benchmarks, you could run:
```
matzew, line 175: Java class (could be confused w/ a "class annotation" ? 

Perhaps list all options here? (Even we just have that `ExactFilterBenchmark` one)

Cali0707: I linked to the existing list of supported classes so that we only have to update one place going forwards

data-plane/benchmarks/run.sh
```diff
@@ -0,0 +1,68 @@
...
+# check if the benchmark class exists
+
+if [[ ! -z $1 ]]; then
+  FOUND=0
+  while IFS="" read -r p || [[ -n "$p" ]]; do
+    if [[ "$1" == "$p" ]]; then
+      FOUND=1
+      break
+    fi
+  done <"${SCRIPT_DIR}/resources/filter-class-list.txt"
```
creydr, line 45: Not an issue, but more a question: why didn't you use something like `egrep -q "^${1}$"`

Cali0707: Just didn't think of it as I had already written the loop to go through the lines of `filter-class-list.txt` to run the benchmarks for all of the classes :smile:

Cali0707: @creydr do you think I should switch? Your solution does seem cleaner to me

creydr: Whatever you prefer. I am fine with both and could change it later anyhow

creydr:
/lgtm
/approve

/hold
Feel free to unhold, when you're OK with https://github.com/knative-extensions/eventing-kafka-broker/pull/3271#discussion_r1295485866

data-plane/benchmarks/run.sh
Cali0707: I think i'll leave it for now since it keeps the script more consistent with itself

Issue #3181, opened 2023-06-28
Implement the auto-create feature flag for the Apache Kafka backed Knative Channel 


matzew:
With the following two PRs in place:

* https://github.com/knative/eventing/pull/7023
* https://github.com/knative/eventing/pull/7034

We can add support for "event auto create" also for the data plane of the Apache Kafka backed Knative Channel .

matzew:
related: https://github.com/knative-sandbox/eventing-kafka-broker/issues/3174

Cali0707:
@pierDipi for this issue it looks like I should be using the KubernetesClient from fabric.io. However, the KubernetesClient we are using doesn't have the Knative extension installed, and the Knative extension in fabric.io seems to be out of date (it's on v0.34.0 of Knative).

Should I
1. Update the Knative models in fabric.io and install the Knative extension into our project
2. Create a class which implements the needed methods for EventTypes

My impression is that 2 would be faster and easier, but I am not sure it would be the best option - WDYT?

pierDipi:
The other question is about caching, we cannot make a k8s remove call on each received event, so we need some form of caching, in Go we use the listers (+ watcher) as cache, do we have the same in Fabric that we can leverage?

Cali0707:
> The other question is about caching, we cannot make a k8s remote call on each received event, so we need some form of caching, in Go we use the EventType listers (+ watcher) as cache, do we have the same in Fabric that we can leverage?

@pierDipi it looks like there are Listers in fabric: https://javadoc.io/doc/io.fabric8/kubernetes-client/4.6.4/io/fabric8/kubernetes/client/informers/cache/Lister.html

I am going to try and use one of those

PR #3284, opened 2023-08-10
Pass eventtype owner to data plane

Cali0707:
Part of #3181

We need the eventtype Reference/Owner information in the dataplane so that we can auto create event types in #3181.

## Proposed Changes

- Pass the event type ownership info through the data plane contract


Cali0707:
I think the upgrade test failures might be real failures, I will look into it

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -608,12 +608,27 @@ func rebuildCMFromStatusAnnotations(br *eventing.Broker) *corev1.ConfigMap {
...
 		Ingress: &contract.Ingress{
 			Path:                       receiver.PathFromObject(broker),
 			EnableAutoCreateEventTypes: feature.FromContext(ctx).IsEnabled(feature.EvenTypeAutoCreate),
+			EventTypeOwnerReference: &contract.Reference{
+				Uuid:      string(broker.UID),
+				Name:      broker.Name,
+				Namespace: broker.Namespace,
+				Kind:      brokerKind,
+				Version:   brokerAPIVersion,
+			},
```
pierDipi, line 631: Why not including this in `reference` field?

pierDipi, line 631: I would simplify this to something like:
```suggestion
	resource := &contract.Resource{
		Uid:    string(broker.UID),
		Topics: []string{topic},
		Ingress: &contract.Ingress{
			Path:                       receiver.PathFromObject(broker),
			EnableAutoCreateEventTypes: feature.FromContext(ctx).IsEnabled(feature.EvenTypeAutoCreate),
			EventTypeOwnerReference: &contract.Reference{
				Uuid:      string(broker.UID),
				Name:      broker.Name,
				Namespace: broker.Namespace,
				Kind:      "Broker",
				Version:   eventing.SchemeGroupVersion.String(),
			},
```

Cali0707: @pierDipi is there a different reference field I could be including them on?

pierDipi: Yes, it's called `Reference` under `Resource`

https://github.com/knative-extensions/eventing-kafka-broker/blob/0242a8795c39a50266570f3cf3211bbabbd209b5/proto/contract.proto#L333-L339

Cali0707:
@Leo6Leo I think the test failure is an actual issue, not a flaky failure here. Not sure what is causing the issue though...

control-plane/pkg/reconciler/testing/objects_channel.go
```diff
@@ -259,6 +259,8 @@ func ChannelReference() *contract.Reference {
 		Uuid:      ChannelUUID,
 		Namespace: ChannelNamespace,
 		Name:      ChannelName,
+		Kind:      "KafkaChannel",
+		Version:   messagingv1beta1.SchemeGroupVersion.String(),
```
pierDipi, line 263: I think `Version` here is meant to be `ResourceVersion` as opposed to `GroupVersion`

Cali0707: Oh okay, I need the group version so I will add a new field for that then. Thanks!

Cali0707:
Now that changes to the contract are backwards compatible, the tests are passing on this PR.

/cc @pierDipi @Leo6Leo

pierDipi (approved):
/lgtm
/approve

PR #3286, opened 2023-08-15
Make SecretSpec field of consumers Auth omitempty

mgencur:
This will prevent writing SecretSpec with null value and the webhook will rather omit the element. 
```
 spec:
        auth:
          SecretSpec: null
```
Having the element there with null value can also caus problems during downgrades when the element was not there yet (it's a few versions back though)

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


matzew (approved):
/lgtm
/approve

PR #3292, opened 2023-08-19
Setup script for e2e reconciler test

debasishbsws:
Added blank `.sh` file for ReconcilerTest with the Loom module images for `Receiver` and `Dispatcher`


debasishbsws:
As currently we are using Env variable to specify to use loom module. We should not need a new test script.

PR #3293, opened 2023-08-21
Fix slow consumergroup reconciliation under load

### LINKED ISSUE #3046: Slow consumergroup reconciliation under load

pierDipi:
In [1], we return an error, however, that method is called in the consumer group reconciler when a consumer is not ready which is a normal state at the beginning when consumers have been just created, so we shouldn't return an error because that causes the consumer group to be reconciled again with an exponentially increasing delay causing slow time to ready.

This is especially evident when scaling up with a high load = (therefore when dispatcher pod is slow to become ready).

[1] https://github.com/knative-sandbox/eventing-kafka-broker/blob/5cda5463aa2fa060179674fe7b3237abb836ee06/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_lifecycle.go#L57-L65

Fixes #3046 

## Proposed Changes

- Fix slow consumergroup reconciliation under load

**Release Note**

```release-note
Fix slow `ConsumerGroup` reconciliation under load
```

**Docs**

### DIFF
--- a/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_lifecycle.go
+++ b/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_lifecycle.go
@@ -61,7 +61,10 @@ func (cg *ConsumerGroup) MarkReconcileConsumersFailedCondition(condition *apis.C
 		condition.GetMessage(),
 	)
 
-	return fmt.Errorf("consumers aren't ready, %v: %v", condition.GetReason(), condition.GetMessage())
+	// It is "normal" to have non-ready consumers, and we will get notified when their status change,
+	// so we don't need to return an error here which causes the object to be queued with an
+	// exponentially increasing delay.
+	return nil
 }
 
 func (cg *ConsumerGroup) MarkReconcileConsumersSucceeded() {
--- a/control-plane/pkg/reconciler/consumergroup/consumergroup_test.go
+++ b/control-plane/pkg/reconciler/consumergroup/consumergroup_test.go
@@ -399,7 +399,7 @@ func TestReconcileKind(t *testing.T) {
 					}, nil
 				}),
 			},
-			WantErr: true,
+			WantErr: false,
 			WantCreates: []runtime.Object{
 				NewConsumer(1,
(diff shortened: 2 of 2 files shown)


control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_lifecycle.go
```diff
@@ -61,7 +61,10 @@ func (cg *ConsumerGroup) MarkReconcileConsumersFailedCondition(condition *apis.C
 		condition.GetMessage(),
 	)
 
-	return fmt.Errorf("consumers aren't ready, %v: %v", condition.GetReason(), condition.GetMessage())
+	// It is "normal" to have non-ready consumers, and we will get notified when their status change,
+	// so we don't need to return an error here which causes the object to be queued with an
+	// exponentially increasing delay.
+	return nil
```
matzew, line 67: thanks for the explicit comment.

It seems reasonable

matzew (approved):
/lgtm
/approve

PR #3305, opened 2023-08-24
Expose tls port for broker

Leo6Leo:
## Proposed Changes

- Expose the port 8443


data-plane/config/broker/500-receiver.yaml
```diff
@@ -212,6 +211,10 @@ spec:
       port: 8080
       protocol: TCP
       targetPort: 8080
+    - name: https
```
pierDipi, line 214: https already exists, I think you meant to have this?
```suggestion
    - name: https-container
```

pierDipi (approved):
/lgtm
/approve

PR #3325, opened 2023-09-08
Modify the existing fileWatcher to make it more generic

### LINKED ISSUE #3323: SecretWatcher can't catch symbolic link change
**Describe the bug**

When the Secret or ConfigMap is updated, Kubernetes doesn't overwrite the original files. Instead, it creates a new temporary directory (..data), places the updated files there, and then updates the symlinks to point to the new files.

When the cert rotation happens, k8s will update the secret, and the mounted secret volume will be updated accordingly. 

 **Expected behavior**
When  k8s update the secret, and the mounted secret volume will be updated accordingly.(It is updated according to the way mentioned above). The secret watcher should be able to catch that change.

### LINKED ISSUE #3338: Use the new fileWatcher instead of secretWatcher
**Problem**
Currently in the vert.x server implementation, secretWatcher is used to monitor the changes for the k8s mounted secret volume files. After #3325 has been merged, remove the secretWatcher implementation and use the new fileWatcher.

**Exit Criteria**
fileWatcher can successfully catch the k8s mounted secret volume change.

**Time Estimate (optional):** 
1

Leo6Leo:
Fixes #3323 
Fixes #3338 

## Proposed Changes

- Modify the existing fileWatcher to make it more generic
- Remove the secretWatcher
- Update the current usage of fileWatcher

**Release Note**

```release-note

```


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/file/FileWatcher.java
```diff
@@ -38,19 +38,22 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * This class is responsible for watching a given file and reports update.
+ * This class is responsible for watching a given file and reports update or execu
```
pierDipi, line 41: ` or execu`?

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/file/FileWatcher.java
```diff
@@ -62,18 +65,27 @@ public class FileWatcher implements AutoCloseable {
...
         this.lastContract = -1;
     }
 
+    public void setContractConsumer(Consumer<DataPlaneContract.Contract> contractConsumer) {
+        this.contractConsumer = contractConsumer;
+    }
+
+    public void setTriggerFunction(Runnable triggerFunction) {
+        this.triggerFunction = triggerFunction;
+    }
```
pierDipi, line 83: Can we use the constructor rather than set methods? I'm pretty sure that set methods make the class not thread safe anymore because calling them will have an undefined behavior / data race

pierDipi: Also if we want to make this class generic it should have no awareness of `DataPlaneContract.Contract`

pierDipi: We should abstract consumer/trigger away, they are basically the same think

pierDipi (commented):
I'd consider making the class more generic that it is now, it shouldn't have awareness of `DataPlaneContract.Contract`.

Did you consider a constructor like this?
```java
public FileWatcher(File file, triggerUpdate Runnable) {
```

instead of calling the existing `update` we would call `triggerUpdate.run()`

On the usage side we would use the class with something like this for the contract update:

```java
final var fw = new FileWatcher(contractFile, new ContractPublisher(contractFile));
```

where `UpdateContract` would do what the current `update` does.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/file/FileWatcher.java
```diff
@@ -107,7 +119,8 @@ public synchronized void close() throws Exception {
     private void run() {
         try {
             // register the given watch service.
-            // Note: this watch a directory and not the single file we're interested in, so that's the
+            // Note: this watch a directory and not the single file we're interested in, so
+            // that's the
             // reason in #watch() we filter watch service events based on the updated file.
```
pierDipi, line 124: It's a little weird the broken line of the comment 
"the <ton of space>
reason"

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/file/FileWatcher.java
Leo6Leo: emm weired. Seems like spotless cuts it.

data-plane/core/src/test/java/dev/knative/eventing/kafka/broker/core/eventbus/ContractPublisherTest.java
```diff
@@ -43,4 +58,106 @@ public void publishTest(Vertx vertx, VertxTestContext testContext) {
...
+                assertThat(broker).isEqualTo(broker1);
+                waitFirst.countDown();
+            } else if (!broker.equals(broker1)) {
+                assertThat(broker).isEqualTo(broker2);
+                waitSecond.countDown();
+            }
+        };
+
+        try {
+            FileWatcher fw1 = new FileWatcher(file, () -> brokersConsumer.accept(broker1));
```
Leo6Leo, line 91: @pierDipi  I am not sure whether this is correct or not way to implement it.

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/file/FileWatcher.java
```diff
@@ -104,58 +106,52 @@ public synchronized void close() throws Exception {
...
-                // check if we're interested in the updated file
-                if (kind != OVERFLOW) {
-                    shouldUpdate = true;
-                    break;
+                // We check if the event's context (the file) matches our target file
+                final var filename = event.context();
+                if (kind != OVERFLOW && filename.toString().equals(toWatch.getName())) {
+                    if (triggerFunction != null) {
+                        triggerFunction.run();
+                    }
```
pierDipi, line 150: `triggerFunction` is never null, it's checked in the constructor
```java
Objects.requireNonNull(triggerFunction, "provide triggerFunction");
```

data-plane/benchmarks/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/filter/ExactFilterBenchmark.java
Leo6Leo, line 1: I don't know what's wrong with this file...

data-plane/benchmarks/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/filter/FilterBenchmark.java
Leo6Leo, line 1: I don't know what's wrong with this file...

pierDipi:
@Leo6Leo I see
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:testCompile (default-testCompile) on project core: Compilation failure
[ERROR] /home/prow/go/src/knative.dev/eventing-kafka-broker/data-plane/core/src/test/java/dev/knative/eventing/kafka/broker/core/eventbus/ContractPublisherTest.java:[88,51] ';' expected
```

control-plane/pkg/contract/contract.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.26.0
-// 	protoc        v3.17.3
+// 	protoc        v3.15.8
```
pierDipi, line 4: This change needs to be reverted

pierDipi (approved):
Thank you @Leo6Leo !

/lgtm
/approve

Issue #3331, opened 2023-09-12
Add Spotless Script to hack/run.sh


Cali0707:
**Problem**
Currently, spotless is run when someone runs `./hack/update-codegen.sh`, or if they know how to manually use spotless itself. We should abstract that away from contributors and allow them to run `./hack/run.sh format-java` and have it format all the java code for them.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?
Contributors

**Exit Criteria**
Running `./hack/run.sh format-java` formats all the java code.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve? 1

**Additional context (optional)**
Add any other context about the feature request here.

MeenuyD:
Hello @Cali0707 I would like to work on this issue can you please guide me on how to get started

Cali0707:
@MeenuyD thanks for picking this up! In general, the pattern for adding a command to `./hack/run.sh` is to add the command into https://github.com/knative-extensions/eventing-kafka-broker/blob/main/hack/run.sh, and then in the big set of if statements there, run whatever needs to be run (either calling a new script or running a few bash commands). Here, you would want to run `${ROOT_DIR}/mvnw spotless:apply`.
/assign @MeenuyD

Cali0707:
Completed by #3333

PR #3333, opened 2023-09-13
Added Spotless Script to hack/run.sh

MeenuyD:
Currently, spotless formatting is triggered either when someone runs ./hack/update-codegen.sh or manually uses spotless. To simplify the process for contributors, this commit abstracts the formatting step away from contributors and enables them to run ./hack/run.sh format-java to automatically format all Java code.

#3331

### DIFF
--- a/hack/run.sh
+++ b/hack/run.sh
@@ -95,8 +95,9 @@ elif [[ "${action}" == "benchmark-filter" ]]; then
   "${ROOT_DIR}/data-plane/benchmarks/run.sh" "$2"
 elif [[ "${action}" == "benchmark-filters" ]]; then
   "${ROOT_DIR}/data-plane/benchmarks/run.sh"
-elif [[ "${action}" == "./hack/update-codegen.sh" || "${action}" == "./hack/run.sh format-java" ]]; then
-    "${ROOT_DIR}"/mvnw spotless:apply
+elif [[ "${action}" == "format-java" ]]; then
+  cd "${ROOT_DIR}/data-plane"
+  ./mvnw spotless:apply
 else
   echo "Unrecognized action ${action}"
   usage "$0"
--- a/hack/run.sh
+++ b/hack/run.sh
@@ -39,7 +39,6 @@ function usage() {
   echo "   benchmark-filters                                       Run all the filter benchmarks"
   echo ""
 }
-
 if [[ "$action" == "deploy-infra" ]]; then
   source "${ROOT_DIR}"/test/e2e-common.sh && knative_setup
 elif [[ "${action}" == "teardown-infra" ]]; then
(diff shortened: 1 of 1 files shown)


hack/run.sh
```diff
@@ -96,6 +95,8 @@ elif [[ "${action}" == "benchmark-filter" ]]; then
   "${ROOT_DIR}/data-plane/benchmarks/run.sh" "$2"
 elif [[ "${action}" == "benchmark-filters" ]]; then
   "${ROOT_DIR}/data-plane/benchmarks/run.sh"
+elif [[ "${action}" == "./hack/update-codegen.sh" || "${action}" == "./hack/run.sh format-java" ]]; then
+    "${ROOT_DIR}"/mvnw spotless:apply
```
matzew, line 99: I don't see this file

Cali0707: Yeah I think I made a mistake when I commented on the initial issue - this is my bad!

MeenuyD: Can you tell me how should I correct it?

Cali0707, line 99: @MeenuyD sorry for the confusion from my original comment on the issue - I think your code needs to:
1. `cd` to `${ROOT_DIR}/data-plane`
2. Run `./mvnw spotless:apply`

hack/run.sh
```diff
@@ -96,6 +95,8 @@ elif [[ "${action}" == "benchmark-filter" ]]; then
   "${ROOT_DIR}/data-plane/benchmarks/run.sh" "$2"
 elif [[ "${action}" == "benchmark-filters" ]]; then
   "${ROOT_DIR}/data-plane/benchmarks/run.sh"
+elif [[ "${action}" == "./hack/update-codegen.sh" || "${action}" == "./hack/run.sh format-java" ]]; then
```
Cali0707, line 98:
```suggestion
elif [[ "${action}" == "format-java" ]]; then
```
Here, `update-codegen.sh` is an already existing script, so we don't need to check for it.

Also, `${action}` is the argument passed to the script, so it would be just `format-java`

Cali0707 (changes requested):
Thanks for working on this @MeenuyD !

Sorry if my initial comment on the issue caused any confusion

hack/run.sh
MeenuyD: Should I make it like this?
```
elif [[ "${action}" == "format-java" ]]; then
  cd "${ROOT_DIR}/data-plane"
  ./mvnw spotless: apply
```

Cali0707: Almost! There can't be a space between `spotless:` and `apply`, so I would do:
```bash
elif [[ "${action}" == "format-java" ]]; then
  cd "${ROOT_DIR}/data-plane"
  ./mvnw spotless:apply
```

MeenuyD: Ok thank you

Cali0707 (approved):
/lgtm
/cc @pierDipi @matzew 

Thanks for adding this @MeenuyD and welcome to Knative :tada:

pierDipi (approved):
/lgtm
/approve
/retest-required

PR #3335, opened 2023-09-13
Ignore unknown fields in data plane contract

### LINKED ISSUE #3334: Contract should ignore unkown fields when unmarshalling
**Describe the bug**
Currently, when the data plan contract is unmarshalled in the control plane, it fails if there is an unknown field in it. This can cause backwards compatibility issues if someone decides to downgrade their Knative installation

 **Expected behavior**
Unknown fields should be ignored when unmarshalling.

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative release version**

 **Additional context**

Cali0707:
Fixes #3334 

## Proposed Changes

- Use the unmarshalling option to discard unknown fields in the json 

**Release Note**

```release-note
The data plane contract now supports downgrades from versions with new fields
```

### DIFF
--- a/control-plane/pkg/reconciler/base/reconciler.go
+++ b/control-plane/pkg/reconciler/base/reconciler.go
@@ -70,6 +70,10 @@ const (
 	Json     = "json"
 )
 
+var (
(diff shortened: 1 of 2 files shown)


pierDipi (approved):
/lgtm
/approve

PR #3337, opened 2023-09-13
Update the updateServerConfig function to pass in the cert value diectly

### LINKED ISSUE #3336: KeyPairRotation update doesn't work with secretVolumePath
**Describe the bug**
KeyPairRotation update doesn't work with secretVolumePath. Has to pass in the secret value to make it reflect on the vert.x server.

Leo6Leo:
Fixes #3336 

## Proposed Changes
Now:
```
PemKeyCertOptions keyCertOptions = new PemKeyCertOptions()
            .setCertValue(Buffer.buffer(java.nio.file.Files.readString(tlsCrtFile.toPath())))
            .setKeyValue(Buffer.buffer(java.nio.file.Files.readString(tlsKeyFile.toPath())));
```
Past:
```
PemKeyCertOptions keyCertOptions =
                        new PemKeyCertOptions().setKeyPath(tlsKeyFile.getPath()).setCertPath(tlsCrtFile.getPath());
```

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
+++ b/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
@@ -49,19 +49,26 @@ import org.slf4j.LoggerFactory;
 /**
  * This verticle is responsible for implementing the logic of the receiver.
  *
- * <p>The receiver is the component responsible for mapping incoming {@link
- * io.cloudevents.CloudEvent} requests to specific Kafka topics. In order to do so, this component:
+ * <p>
+ * The receiver is the component responsible for mapping incoming {@link
+ * io.cloudevents.CloudEvent} requests to specific Kafka topics. In order to do
+ * so, this component:
  *
  * <ul>
- *   <li>Starts two {@link HttpServer}, one with http, and one with https, listening for incoming
- *       events
- *   <li>Starts a {@link ResourcesReconciler}, listen on the event bus for reconciliation events and
- *       keeps track of the {@link
- *       dev.knative.eventing.kafka.broker.contract.DataPlaneContract.Ingress} objects and their
- *       {@code path => (topic, producer)} mapping
- *   <li>Implements a request handler that invokes a series of {@code preHandlers} (which are
- *       assumed to complete synchronously) and then a final {@link IngressRequestHandler} to
- *       publish the record to Kafka
+ * <li>Starts two {@link HttpServer}, one with http, and one with https,
+ * listening for incoming
+ * events
+ * <li>Starts a {@link ResourcesReconciler}, listen on the event bus for
+ * reconciliation events and
+ * keeps track of the {@link
+ * dev.knative.eventing.kafka.broker.contract.DataPlaneContract.Ingress} objects
+ * and their
+ * {@code path => (topic, producer)} mapping
+ * <li>Implements a request handler that invokes a series of {@code preHandlers}
+ * (which are
+ * assumed to complete synchronously) and then a final
+ * {@link IngressRequestHandler} to
+ * publish the record to Kafka
  * </ul>
  */
 public class ReceiverVerticle extends AbstractVerticle implements Handler<HttpServerRequest> {
@@ -220,7 +227,7 @@ public class ReceiverVerticle extends AbstractVerticle implements Handler<HttpSe
         this.ingressRequestHandler.handle(requestContext, producer);
     }
 
-       public void updateServerConfig() {
+    public void updateServerConfig() {
 
         // This function will be called when the secret volume is updated
         File tlsKeyFile = new File(tlsKeyFilePath);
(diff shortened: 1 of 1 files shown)


data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
```diff
@@ -220,26 +228,32 @@ public void handle(HttpServerRequest request) {
...
+
+                // result is a Future object
+                Future<Void> result = httpsServer.updateSSLOptions(new SSLOptions().setKeyCertOptions(keyCertOptions));
+
+                result.onSuccess(v -> {
+                            logger.info("Succeeded to update TLS key pair");
+                        })
+                        .onFailure(e -> {
+                            logger.error("Failed to update TLS key pair while executing updateSSLOptions", e);
+                        });
```
pierDipi, line 252: Can we just chain `onSuccess` and `onFailure` to `updateSSLOptions` without intermediate variables?

Also lambdas with only a single statement can be collapsed and we can avoid using `v -> { <single statement> }` becomes `v -> <single statement>`, it's easier to read

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
```diff
@@ -220,26 +228,32 @@ public void handle(HttpServerRequest request) {
...
-                        logger.error("Failed to update TLS key pair", e);
-                    });
+            try {
+                // Update SSL configuration by using updateSSLOptions
+                PemKeyCertOptions keyCertOptions = new PemKeyCertOptions()
+                        .setCertValue(Buffer.buffer(java.nio.file.Files.readString(tlsCrtFile.toPath())))
+                        .setKeyValue(Buffer.buffer(java.nio.file.Files.readString(tlsKeyFile.toPath())));
+
+                // result is a Future object
+                Future<Void> result = httpsServer.updateSSLOptions(new SSLOptions().setKeyCertOptions(keyCertOptions));
```
pierDipi, line 245: This comment is not useful, I can see what type is that

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
```diff
@@ -220,26 +228,32 @@ public void handle(HttpServerRequest request) {
...
-            Future<Void> result = httpsServer.updateSSLOptions(new SSLOptions().setKeyCertOptions(keyCertOptions));
-
-            result.onSuccess(v -> {
-                        logger.info("Succeeded to update TLS key pair");
-                    })
-                    .onFailure(e -> {
-                        logger.error("Failed to update TLS key pair", e);
-                    });
+            try {
+                // Update SSL configuration by using updateSSLOptions
```
pierDipi, line 239: This comment is not useful, I can see that we're calling `updateSSLOptions`

pierDipi: It' is much more useful if we add a comment on *why* using "paths" as opposed to "values" doesn't work as it was previously done

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
```diff
@@ -220,26 +228,29 @@ public void handle(HttpServerRequest request) {
...
-
-            result.onSuccess(v -> {
-                        logger.info("Succeeded to update TLS key pair");
-                    })
-                    .onFailure(e -> {
-                        logger.error("Failed to update TLS key pair", e);
-                    });
+            try {
+                // Update SSL configuration by passing the new value of the certificate and key
+                // Have to use value instead of path here otherwise the changes won't be applied
```
Cali0707, line 240: I'm curious, do you know why this is the case @Leo6Leo?

Leo6Leo: I am not sure whether problem is here or not. 
https://github.com/eclipse-vertx/vert.x/blob/6954961d9b723ed245f91b6dedb41b9b09e991bd/src/main/java/io/vertx/core/net/impl/SSLHelper.java#L182C11-L182C71

Because  the line ` if (prev.succeeded() && prev.result().options.equals(options)) {` is checking whether the options are the same. If we are using the path, obviously the options doesn't get changed because the file path is the same. And so it doesn't get updated? 

That's my guess. Wdyt @Cali0707 @pierDipi

pierDipi (approved):
Nice findings @Leo6Leo 

/lgtm
/approve

Left one nit comment to be consistent with the style of the other code which often use "final var"

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
Leo6Leo: https://github.com/eclipse-vertx/vert.x/issues/4867

Issue reported @pierDipi

Cali0707 (approved):
/lgtm

PR #3354, opened 2023-09-20
Cache init offsets results in the consumer group reconciler (#817)

pierDipi:
When there is high load and multiple consumer group schedule calls,
we get many `dial tcp 10.130.4.8:9092: i/o timeout` errors when
trying to connect to Kafka.
This leads to increased "time to readiness" for consumer groups.

The downside of caching is that, in the case, partitions
increase while the result is cached we won't initialize
the offsets of the new partitions.

## Proposed Changes

- Cache init offsets results in the consumer group reconciler 

**Release Note**

```release-note
When there is high load and multiple consumer group schedule calls, we get many `dial tcp 10.130.4.8:9092: i/o timeout` errors when trying to connect to Kafka. This leads to increased "time to readiness" for consumer groups.

The downside of caching is that, in the case, partitions increase while the result is cached we won't initialize on the control plane the offsets of the new partitions.
```

/kind bug


matzew (approved):
/lgtm
/approve

PR #3355, opened 2023-09-20
Set consumer and consumergroups finalizers when creating them

pierDipi:
This is a mitigation for https://github.com/knative/pkg/issues/2828

Before this patch, it was possible that a deleted consumer or 
consumergroup might be reconciled and never finalized when
they are deleted before the finalizer is set.

This happens because the Knative generated reconciler uses
patch (as opposed to using update) for setting the finalizer
and patch doesn't have any optimistic concurrency controls.

## Proposed Changes

- Set consumer and consumergroups finalizers when creating them

**Release Note**

```release-note
Before this patch, it was possible that a deleted consumer or 
consumergroup might be reconciled and never finalized when
they are deleted before the finalizer is set.

This happens because the Knative generated reconciler uses
patch (as opposed to using update) for setting the finalizer
and patch doesn't have any optimistic concurrency controls.
```

### DIFF
--- a/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
+++ b/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
@@ -156,12 +156,13 @@ func (c *ConsumerGroup) GetStatus() *duckv1.Status {
 func (cg *ConsumerGroup) ConsumerFromTemplate(options ...ConsumerOption) *Consumer {
 	// TODO figure out naming strategy, is generateName enough?
 	c := &Consumer{
-		ObjectMeta: cg.Spec.Template.ObjectMeta,
-		Spec:       cg.Spec.Template.Spec,
+		ObjectMeta: *cg.Spec.Template.ObjectMeta.DeepCopy(),
+		Spec:       *cg.Spec.Template.Spec.DeepCopy(),
 	}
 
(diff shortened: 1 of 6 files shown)


pierDipi:
yes

/retest

control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
```diff
@@ -156,12 +156,13 @@ func (c *ConsumerGroup) GetStatus() *duckv1.Status {
 func (cg *ConsumerGroup) ConsumerFromTemplate(options ...ConsumerOption) *Consumer {
 	// TODO figure out naming strategy, is generateName enough?
 	c := &Consumer{
-		ObjectMeta: cg.Spec.Template.ObjectMeta,
-		Spec:       cg.Spec.Template.Spec,
+		ObjectMeta: *cg.Spec.Template.ObjectMeta.DeepCopy(),
+		Spec:       *cg.Spec.Template.Spec.DeepCopy(),
```
Cali0707, line 160: Why are we switching to using deep copies here?

pierDipi: because we're changing the fields down below and if objects come from informers they must not be modified, so "it's better to be safe than sorry"

Cali0707 (approved):
/lgtm

PR #3359, opened 2023-09-25
Evictor will not set unschedulable annotation anymore

pierDipi:
Sometimes the `evictor` sets the unschedulable annotation in the middle of the Statefulset sequence and that causes to have unschedulable pods running and doing nothing, since this is an optimization to avoid
consumers reshuffles but leads to wrong results, we're removing it.


matzew (approved):
/lgtm
/approve

PR #3370, opened 2023-10-02
Improve KEDA installed check efficiency

pierDipi:
Instead of listing a potentially large number of ScaledObjects, we can limit this number to the smallest number: 1.


Cali0707 (approved):
/lgtm

PR #3381, opened 2023-10-06
Refactor filters to use lists instead of sets

### LINKED ISSUE #3364: New Event Filtering: Use List instead of Set for Any and All filters
**Problem**
The current implementation of the Any and All filters uses a Set, but we will likely get better performance with a List instead.

**Exit Criteria**
The Any and All filters use a List, and there are benchmark results showing improved performance

**Time Estimate (optional):** 

Cali0707:
Fixes #3364 

## Proposed Changes

- Switch from using Set to List for Any filter
- Switch from using Set to List for All filter


pierDipi (approved):
/lgtm
/approve

PR #3401, opened 2023-10-16
Sped up benchmarks

### LINKED ISSUE #3320: New Event Filtering: Speed up Java Benchmarks
**Problem**
Currently, the Java Benchmarks for the New Event Filtering take a really long time to run. @pierDipi indicated that they could be sped up by tuning the parameters, such as what type of benchmarks we want to run (rather than all of them which is the default).

**Exit Criteria**
The benchmarks take less time to run, while still providing useful results.

**Time Estimate (optional):** 

- #3296 
- #3247

Cali0707:
Fixes #3320 

## Proposed Changes

- Only benchmark throughput rather than all modes
- Reduce the number of warmup trials and their runtimes
- Reduce the number of measurement trials
- (bonus: small bugfix to the existing benchmarks for all filter and any filter)


pierDipi (approved):
/lgtm
/approve

PR #3421, opened 2023-10-23
Made the async prober private

Cali0707:
Fixes #3190
Fixes #3141 

## Proposed Changes

- Make the async prober private, since we should only use the composite prober now that TLS is supported

PR #3422, opened 2023-10-23
Fix gofmt issue

creydr:
As per title.

Seeing this issue e.g. in #3421 while it's not related to the PR.


Cali0707 (approved):
Thanks @creydr 
/lgtm

pierDipi:
You're always on top of this things, thank you!

PR #3448, opened 2023-10-31
Change fabric8io HTTP client from okhttp to the core HTTP client to support IPv6 clusters

pierDipi:
This PR changes the fabric8io/kubernetes-client HTTP client from okhttp to the core HTTP client and will reduce the number of transitive dependencies while also supporting IPv6 clusters (To be tested)

Related to https://github.com/knative-extensions/eventing-kafka-broker/issues/3005


creydr (approved):
Nice. Thanks
/lgtm

PR #3457, opened 2023-11-04
Stash only relevant broker config keys on the broker status

### LINKED ISSUE #2914: Stash only relevant broker config keys on the broker status
**Describe the bug**

All keys from a CM are stashed as annotations. 

Given CM:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-broken-kafka-broker-config
data:
  default.topic.partitions: "10"
  bootstrap.servers: "my-cluster-kafka-bootstrap.kafka:9093"
  auth.secret.ref.name: "not-present"
  blah.blup: foo-bar
```

Resulting in :

```
...
  status:
    annotations:
      auth.secret.ref.name: not-present
      blah.blup: foo-bar
      bootstrap.servers: my-cluster-kafka-bootstrap.kafka:9093
      default.topic.partitions: "10"
    conditions:
...
```

 **Expected behavior**

Not seeing the `blah.blup` annotation key on the status

basit9958:
Fixes #2914

## Proposed Changes

- Stash only relevant broker config keys on the broker status
- 

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -576,6 +576,7 @@ func storeConfigMapAsStatusAnnotation(broker *eventing.Broker, cm *corev1.Config
 		kafka.DefaultTopicNumPartitionConfigMapKey:      true,
 		kafka.DefaultTopicReplicationFactorConfigMapKey: true,
 		kafka.BootstrapServersConfigMapKey:              true,
+		security.AuthSecretNameKey:                      true,
 	}
 
 	for k, v := range cm.Data {
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -572,8 +572,16 @@ func storeConfigMapAsStatusAnnotation(broker *eventing.Broker, cm *corev1.Config
 		broker.Status.Annotations = make(map[string]string, len(cm.Data))
 	}
 
+	keysToStore := map[string]bool{
+		kafka.DefaultTopicNumPartitionConfigMapKey:      true,
+		kafka.DefaultTopicReplicationFactorConfigMapKey: true,
+		kafka.BootstrapServersConfigMapKey:              true,
+	}
+
 	for k, v := range cm.Data {
-		broker.Status.Annotations[k] = v
+		if keysToStore[k] {
+			broker.Status.Annotations[k] = v
+		}
 	}
 }
 


basit9958:
cc @Leo6Leo @lionelvillard

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -572,8 +572,17 @@ func storeConfigMapAsStatusAnnotation(broker *eventing.Broker, cm *corev1.Config
 		broker.Status.Annotations = make(map[string]string, len(cm.Data))
 	}
 
+	keysToStore := map[string]bool{
+		kafka.DefaultTopicNumPartitionConfigMapKey:      true,
+		kafka.DefaultTopicReplicationFactorConfigMapKey: true,
+		kafka.BootstrapServersConfigMapKey:              true,
+		security.AuthSecretNameKey:                      true,
```
Cali0707, line 579: @basit9958 I don't think that we need to keep this key - the config map doesn't normally have this key (as far as I can tell). See here: https://github.com/knative-extensions/eventing-kafka-broker/blob/8a79f31bea81dcb2d94a0bb3524f080fd66ea9c3/control-plane/config/eventing-kafka-broker/100-broker/100-kafka-broker-configmap.yaml

cc @matzew @creydr in case I am missing something
```suggestion
```

control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -572,8 +572,17 @@ func storeConfigMapAsStatusAnnotation(broker *eventing.Broker, cm *corev1.Config
 		broker.Status.Annotations = make(map[string]string, len(cm.Data))
 	}
 
+	keysToStore := map[string]bool{
```
Cali0707, line 575: @basit9958 could you make this a top-level var, rather than creating this map every time this function is called? I.e. on line 67, just after the consts, could you add:

```go
var (
    keysToStore := ...
)
```

Cali0707 (changes requested):
Thanks for working on this @basit9958 !

Just a few things to change as far as I can see, then we should be good to merge this

Leo6Leo:
Hey @basit9958! Are you being blocked by anything? Any help we could provide you with?

Cali0707 (approved):
/lgtm
/approve

Sorry this got lost in my todo list @basit9958 - thanks for fixing this!

/retest-required

PR #3464, opened 2023-11-13
Add kafka ce extensions for partition and offset

### LINKED ISSUE #3072: Expose Kafka record offset and partition
**Problem**
When the offset is known to the consumer, it is no longer reliant on ordered consumption of the events. E.g. a consumer could use the offset+partition to order the events by itself.

Event consumer (developer)

**Exit Criteria**
The incoming HTTP request contains the offset and partition information.

**Time Estimate (optional):**
1 day

Cali0707:
Fixes #3072 

## Proposed Changes

- Expose kafka partition and offset as new CE extensions
-
-

**Release Note**

```release-note
Events coming from kafka components now have knativekafkaoffset and knativekafkapartition cloudevents extension attributes set
```

**Docs**


Cali0707:
@matzew @pierDipi it looks like a lot of tests are failing because they don't expect the event to have the two new extensions I've added. Since the issue doesn't specify that this needs to be an CE extension, I'm wondering - should I switch my implementation to just set a header (and not a CE extension)?

I personally think the CE extension approach is better since it will let users write filters for different conditions if they want to on their triggers. However, I'd rather check first before I fix all the tests :)

pierDipi:
Also we can ask the author of the issue to give us feedback as the exit criteria on the issue may be ambiguous as we're using binary content mode to send events over HTTP, so CE attributes are headers

> The incoming HTTP request contains the offset and partition information.

Cali0707:
> @Cali0707 the spec says this https://github.com/knative/specs/blob/main/specs/eventing/data-plane.md#event-acknowledgement-and-delivery-retry:
> 
> > In general, senders MAY add or update other CloudEvent attributes on each delivery attempt; see [observability](https://github.com/knative/specs/blob/main/specs/eventing/data-plane.md#observability) for an example case.
> 
> so tests should not fail when attributes are added

I think the issue is the tests are expecting a specific cloud event currently, which we aren't matching?

For example:

```
     topology.go:179: unexpected event routing behaviour (-want, +got) =   []event.Event{
          	{
          		Mode: "",
          		Attributes: event.ContextAttributes{
          			... // 1 ignored and 7 identical fields
          			DataContentEncoding: "",
          			DataContentType:     "text/json",
          			Extensions: event.Extensions{
          				... // 4 identical entries
          				"extime":                "2020-03-21T12:34:56.78Z",
          				"exurl":                 "http://example.com/source",
        + 				"knativekafkaoffset":    "0",
        + 				"knativekafkapartition": "0",
          			},
          		},
          		TransportExtensions: nil,
          		Data:                `"hello"`,
          	},
          } 
```

Should I be fixing these @pierDipi ?

matzew:
well:

```

        + 				"knativekafkaoffset":    "0",
        + 				"knativekafkapartition": "0",
```

when we add those to each request, I guess we either need to check they are there, or ignore them (being more lenient on extra headers, which may have a downside)

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/CloudEventOverridesMutator.java
```diff
@@ -28,21 +29,42 @@ public class CloudEventOverridesMutator implements CloudEventMutator {
...
+        return cloudEventDeserializer.deserialize(record.topic(), record.headers(), null);
+    }
+
     private void applyCloudEventOverrides(CloudEventBuilder builder) {
         cloudEventOverrides.getExtensionsMap().forEach(builder::withExtension);
     }
+
+    private void applyKafkaMetadata(CloudEventBuilder builder, Number partition, Number offset) {
+        builder.withExtension("knativekafkapartition", partition);
+        builder.withExtension("knativekafkaoffset", offset);
```
matzew, line 68: can we make those `private static final String` ?

vendor/knative.dev/eventing/test/rekt/features/broker/topology.go
```diff
@@ -172,7 +172,7 @@ func assertExpectedRoutedEvents(prober *eventshub.EventProber, expected map[stri
 			if len(want) != 0 && len(got) != 0 {
 				// ID is adjusted by eventshub.
 				except := []cmp.Option{
-					cmpopts.IgnoreFields(conformanceevent.ContextAttributes{}, "ID"),
+					cmpopts.IgnoreFields(conformanceevent.ContextAttributes{}, "ID", "Extensions"),
```
matzew, line 175: see https://github.com/knative/eventing/pull/7467

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/CloudEventOverridesMutator.java
matzew: I understand the prefix of knative...  but wondering if the actual upstream/spec, would want to say something on the metadata in https://github.com/cloudevents/spec/blob/main/cloudevents/bindings/kafka-protocol-binding.md  ?

Cali0707: I didn't see anything in particular while reading the docs there, I'll ask at the serverless WG meeting tomorrow

Cali0707:
This should be ready for review now
/cc @matzew @Leo6Leo

Cali0707:
/cc @pierDipi 
Do you think we can get this into the release?

pierDipi (approved):
/lgtm
/approve

Cali0707:
@pierDipi do you think this needs a corresponding docs update?

Cali0707:
/cc @pierDipi @Leo6Leo 

could I get an LGTM again after fixing the build issue?

PR #3475, opened 2023-11-22
Fix binary events with no data and ce overrides not being delivered

Cali0707:
Fixes bug where binary events with ce overrides don't get delivered

## Proposed Changes

- Deserialize events from the kafka record headers if needed
-
-

**Release Note**

```release-note
:bug: Fixed bug where binary events with no data were not delivered if cloud event overrides were enabled
```

**Docs**

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/Main.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/Main.java
@@ -84,7 +84,7 @@ public class Main {
         consumerConfig.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, CloudEventDeserializer.class.getName());
         consumerConfig.put(
                 ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG,
-                InvalidCloudEventInterceptor.class.getName() + "," + NullCloudEventInterceptor.class.getName());
+                NullCloudEventInterceptor.class.getName() + "," + InvalidCloudEventInterceptor.class.getName());
 
         // Read WebClient config
         JsonObject webClientConfig = Configurations.readPropertiesAsJsonSync(env.getWebClientConfigFilePath());


Cali0707:
/cc @pierDipi 
I'm running just the test first to show the bug, will push my fix in a bit :)

Cali0707:
/hold
until after the fix gets pushed

Cali0707:
The bug can be seen with the test failure from before I pushed the fix [here](https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-extensions_eventing-kafka-broker/3475/reconciler-tests_eventing-kafka-broker_main/1727357331135533056)

The test is now passing with the fix

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/CloudEventOverridesMutator.java
pierDipi, line 1: Can we add a unit test as well?

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/CloudEventOverridesMutator.java
```diff
@@ -42,6 +46,21 @@ public CloudEvent apply(CloudEvent cloudEvent) {
...
+        // A valid CloudEvent in the CE binary protocol binding of Kafka
+        // might be composed by only Headers.
+        //
+        // KafkaConsumer doesn't call the deserializer if the value
+        // is null.
+        //
+        // That means that we get a record with a null value and some CE
+        // headers even though the record is a valid CloudEvent.
+        return cloudEventDeserializer.deserialize(record.topic(), record.headers(), null);
+    }
```
pierDipi, line 62: This logic is duplicated from `RecordDispatcherImpl`, can we extract a common method, usable by both `CloudEventOverridesMutator` and `RecordDispatcherImpl`

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/CloudEventOverridesMutator.java
pierDipi: A good scenario is to test a chain `RecordDispatcherMutatorChain` similar to the one we use in reality

```java
        final var recordDispatcher = new RecordDispatcherMutatorChain(
                new RecordDispatcherImpl(
                        consumerVerticleContext,
                        getFilter(),
                        egressSubscriberSender,
                        egressDeadLetterSender,
                        responseHandler,
                        offsetManager,
                        ConsumerTracer.create(
                                ((VertxInternal) vertx).tracer(),
                                consumerVerticleContext.getConsumerConfigs(),
                                TracingPolicy.PROPAGATE),
                        Metrics.getRegistry()),
                new CloudEventOverridesMutator(
                        consumerVerticleContext.getResource().getCloudEventOverrides()));
```

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/Main.java
```diff
@@ -81,7 +82,9 @@ public static void start(
         Properties consumerConfig = Configurations.readPropertiesSync(env.getConsumerConfigFilePath());
         consumerConfig.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, KeyDeserializer.class.getName());
         consumerConfig.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, CloudEventDeserializer.class.getName());
-        consumerConfig.put(ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG, InvalidCloudEventInterceptor.class.getName());
+        consumerConfig.put(
+                ConsumerConfig.INTERCEPTOR_CLASSES_CONFIG,
+                InvalidCloudEventInterceptor.class.getName() + "," + NullCloudEventInterceptor.class.getName());
```
pierDipi, line 87: Shouldn't the order be the opposite ?

pierDipi (commented):
If we could "unit" test the 2 interceptors in a chain on various cases, that would be great mimicking the "chain" we're configuring in `Main.java`

- passing valid cloud event with null value
- passing valid cloud event with non-null value
- passing invalid cloud event with non-null value
- passing invalid cloud event with null value

data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/Main.java
Cali0707: Maybe? I thought from our original conversation you said to to chain `InvalidCloudEventInterceptor` -> `NullCloudEventInterceptor`, maybe because the Invalid interceptor will still return ConsumerRecords with null data?

If I'm honest, I'm not super confident on which one should go first vs. second

pierDipi: If we add tests we will know the correct order https://github.com/knative-extensions/eventing-kafka-broker/pull/3475#pullrequestreview-1764542407

Cali0707:
/cc @pierDipi 

I've added unit tests now, and fixed the order of the interceptors (you were correct)

pierDipi (approved):
/lgtm
/approve

Thanks!

PR #3478, opened 2023-11-22
Pass event type owner references to the data plane

Cali0707:
With the contract now including fields to ET owner references (needed for autocreate), we should pass those values to the data plane.

## Proposed Changes

- Pass the owner reference info to the data plane.


Cali0707:
Oh - just realized we probably want to backport the data plane contract fix before the upgrade tests will start to pass. I'll open a PR on Monday for that.

creydr (approved):
/lgtm

PR #3479, opened 2023-11-23
Remove activationLagThreshold to fix 0->1 scaling

Cali0707:
Fixes Kafka source tests when KEDA is enabled (hopefully).

From discussion [here](https://github.com/kedacore/keda/discussions/4466), we shouldn't be using `activationLagThreshold` unless we want to initially scale form 0 to n>1. Hence, our scaling 0 -> 1 was not working.

## Proposed Changes

- Remove activationLagThreshold parameter (letting it default to 0)
-
-

**Release Note**

```release-note

```

**Docs**


pierDipi (approved):
/lgtm
/approve

PR #3485, opened 2023-11-27
Update dependencies

creydr:
#3484 has some build issues. This PR addresses it, updates the dependencies and fixes the build issue.


control-plane/pkg/apis/messaging/v1beta1/kafka_channel_validation.go
```diff
@@ -113,3 +115,44 @@ func (kc *KafkaChannel) CheckImmutableFields(_ context.Context, original *KafkaC
 
 	return nil
 }
+
+func (kc *KafkaChannel) CheckSubscribersChangeAllowed(ctx context.Context, original *KafkaChannel) *apis.FieldError {
```
Cali0707, line 119: @creydr I think you need to call this function in `Validate` - see here: https://github.com/knative/eventing/blob/d989ca732fad3701fef677f80594504512a39ee5/pkg/apis/messaging/v1/in_memory_channel_validation.go#L45-L49

creydr: ah ***. Forgot. Thanks and good catch. Updated it

control-plane/pkg/apis/messaging/v1beta1/kafka_channel_validation.go
```diff
@@ -28,6 +28,8 @@ import (
 	"knative.dev/pkg/kmp"
 )
 
+const kafkaControllerSAName = "system:serviceaccount:knative-eventing:kafka-controller"
```
Cali0707, line 31: Maybe this should be `system:serviceaccount:knative-eventing:eventing-controller`, or we should check for both?

From the failing tests:
```
Failed to sync physical Channel: admission webhook \"validation.webhook.kafka.eventing.knative.dev\" denied the request: validation failed: Channel.Spec.Subscribers changed by user system:serviceaccount:knative-eventing:eventing-controller which was not the system:serviceaccount:knative-eventing:kafka-controller service account
```

PR #3486, opened 2023-11-27
ignore unknown fields in data plane contract

Cali0707:
Fixes data plane contract parsing failing when there is an unknown field.

## Proposed Changes

- Allow unknown fields in the contract


pierDipi (approved):
/lgtm
/approve

PR #3491, opened 2023-11-28
Test Keda Kafka Source Scaling

### LINKED ISSUE #3067: Test KafkaSource with KEDA
**Problem**

We're not testing the KafkaSource integration with KEDA, we need to have automated tests running in CI (in a separate job) for the KEDA integration.

Cali0707:
Fixes #3067 

## Proposed Changes

- Add a test which verifies KafkaSource scale from zero and scale to zero
- Reduce the cooldown period from 300 seconds to 30 seconds to align with serving defaults

### DIFF
--- a/test/rekt/features/kafka_source.go
+++ b/test/rekt/features/kafka_source.go
@@ -23,9 +23,11 @@ import (
 	"strings"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/test"
 	. "github.com/cloudevents/sdk-go/v2/test"
 	cetest "github.com/cloudevents/sdk-go/v2/test"
 	cetypes "github.com/cloudevents/sdk-go/v2/types"
+	"github.com/google/uuid"
 	"go.uber.org/zap"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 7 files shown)


test/rekt/features/kafka_source.go
```diff
@@ -561,3 +563,118 @@ func KafkaSourceWithEventAfterUpdate(kafkaSource, kafkaSink, topic string) *feat
...
+	options := []eventshub.EventsHubOption{
+		eventshub.StartSenderToResource(kafkasink.GVR(), kafkaSink),
+	}
+	options = append(options, senderOpts...)
+	f.Requirement("install eventshub sender", eventshub.Install(sender, options...))
+
+	f.Requirement("eventshub receiver gets event", matchEvent(receiver, matcher))
+
+	// we need to ensure that autoscaling is enabled for the rest of the feature to work
+	f.Prerequisite("Autoscaling is enabled", kafkafeatureflags.SourceScalingEnabled())
```
pierDipi, line 631: nit: I would place it somewhere at at the top, like 568 or something like that

test/rekt/features/kafka_source.go
```diff
@@ -561,3 +563,118 @@ func KafkaSourceWithEventAfterUpdate(kafkaSource, kafkaSink, topic string) *feat
...
+		HasSpecVersion(cloudevents.VersionV1),
+		HasType("com.github.pull.create"),
+		HasSource("github.com/cloudevents/spec/pull"),
+		HasSubject("123"),
+		HasId("A234-1234-1234"),
+		HasDataContentType("application/json"),
+		HasData([]byte(`{"hello":"world"}`)),
+		HasExtension("comexampleextension1", "value"),
+		HasExtension("comexampleothervalue", "5"),
+	)
```
pierDipi, line 599: I think we could simplify with 

```golang
event := cetest.FullEvent()
event.SetID(uuid.New().String())
```

and then we can use this _sender_ option

```
eventshub.InputEvent(event)
```

and for matching we could just match on `event.ID()`
```golang
			assert.OnStore(receiver).MatchEvent(test.HasI(event.ID())).Exact(1))
```

since in this test we are not that interested in the "shape" of the event

test/rekt/features/kafka_source.go
```diff
@@ -561,3 +563,118 @@ func KafkaSourceWithEventAfterUpdate(kafkaSource, kafkaSink, topic string) *feat
...
+		kafkasource.WithSink(service.AsKReference(receiver), ""),
+		kafkasource.WithTopics([]string{topic}),
+		kafkasource.WithBootstrapServers(testingpkg.BootstrapServersPlaintextArr),
+	}
+
+	f.Setup("install kafka source", kafkasource.Install(kafkaSource, kafkaSourceOpts...))
+	f.Setup("kafka source is ready", kafkasource.IsReady(kafkaSource))
+
+	// check that the source initially has replicas = 0
+	f.Setup("Source should start with replicas = 0", verifyConsumerGroupReplicas(kafkaSource, 0, true))
```
pierDipi, line 620: can't we move this check at requirement level so that we don't need the "allowNotFound" logic ?

pierDipi: Ah no, that doesn't work

test/rekt/features/kafka_source.go
pierDipi: Any Prerequisite step runs before Setup regardless when it's configured

test/rekt/features/kafkafeatureflags/kafkafeatureflags.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kafkafeature "knative.dev/eventing-kafka-broker/control-plane/pkg/apis/config"
+	kubeclient "knative.dev/pkg/client/injection/kube/client"
+	"knative.dev/pkg/system"
+	"knative.dev/reconciler-test/pkg/feature"
+)
+
+func SourceScalingEnabled() feature.ShouldRun {
```
pierDipi, line 30: I would call this just "AutoscalingEnabled"

pierDipi (approved):
/lgtm
/approve

Cali0707:
@pierDipi I’ll work on backports for this once I’m back from vacation

PR #3493, opened 2023-11-29
Add kafka client pool

### LINKED ISSUE #3397: Optimize Kafka cluster connections
**Problem**

In our controllers, we often re-create Kafka cluster connections, the original reason was this issue https://github.com/IBM/sarama/issues/2060 . Now, it's fixed we need to re-review our Kafka cluster interactions in the control plane (for all resources) and see if we can optimize them by reducing the number of TCP connections created.

`*`

**Exit Criteria**

Lower number of Kafka connections for a single resource reconciliation

**Time Estimate (optional):** 
7

/kind performance

Cali0707:
Fixes #3397 , https://github.com/knative-extensions/eventing-kafka-broker/issues/3139

## Proposed Changes

- Create a client pool to use to get clients and cluster admins for the ekb control plane
- Use an LRU eviction method to ensure that there are never more than a max number of connections being made
-

**Release Note**

```release-note
Kafka control plane connections are now pooled, rather than recreated.
```

**Docs**


Cali0707:
@pierDipi I think this should be ready for at least a first review now! There are probably still some issues I'll need to address, but at this point my solution allows:
1. A max number of connections
2. Reusing connections in a pool
3. Ensuring the connection a reconciler gets is correct for the secret and bootstrap servers that reconciler is using
4. Updating any connections made with a secret when the secret updates

Cali0707:
Another approach that may be worth considering (which would reduce code complexity but lead to more overall connections being opened on the controller side) would be to just make a connection pool per (bootstrapServers, secret) pair, and allow for users to set a connections number there. This would also lead to the need for a global max value as a default, but that is also a requirement for the current solution.

Cali0707:
Seeing a lot of errors along the lines of:
```
    feature.go:224: Warning, failed to delete test-wbqbqbgc/t0-lzfvvial of GVR: /v1, Resource=serviceaccounts: an error on the server ("Internal Server Error: \"/api/v1/namespaces/test-wbqbqbgc/serviceaccounts/t0-lzfvvial\": the server is currently unable to handle the request") has prevented the request from succeeding
```

In the logs from the failed tests. Not sure how my changes could have led to that...

Cali0707:
/hold
While I debug the tests

Cali0707:
/unhold
I think this works now :tada:

control-plane/pkg/kafka/clientpool/clientpool.go
```diff
@@ -0,0 +1,225 @@
...
+	})
+
+	key := clientKey{
+		bootstrapServers: strings.Join(bootstrapServers, ","),
+	}
+
+	if secret != nil {
+		key.secretName = secret.GetName()
+		key.secretNamespace = secret.GetNamespace()
+	}
```
pierDipi, line 148: I think we need to also include the "values" or the object generation / version as if I change the secret I'd like to have a new client, very relevant when creating a secret with bad data

control-plane/pkg/kafka/clientpool/clientpool.go
```diff
@@ -0,0 +1,225 @@
...
+	// if a corresponding connection already exists, lets use it
+	if val, returnClient, ok, err := cp.Get(ctx, key); ok && err == nil {
+		logger.Debug("successfully got a client from the clientpool")
+		// check that the value is still good, as there are still some write errors
+		ca, err := cp.newClusterAdminFromClient(val)
+		if err != nil {
+			returnClient()
+			return nil, NilReturnClientFunc, err
+		}
+		if _, err := ca.ListTopics(); err != nil {
```
pierDipi, line 91: I think we might not have permissions for this operation

control-plane/pkg/kafka/clientpool/clientpool.go
pierDipi: But then the eviction might be more complicated

Cali0707: Currently, there is a event handler added to the secret informer which triggers an update of the value for any matching secret keys - do you think that is good enough?

control-plane/pkg/kafka/clientpool/clientpool.go
Cali0707: Ah okay, I added this because there were still some occasional `broken pipe` errors coming through, so I picked a random method that wouldn't modify anything to verify if the connection was still valid. Should I use something else?

Cali0707:
/retest
/cc @pierDipi 

Could you TAL when you have time?

pierDipi:
/hold

Let's delay this after the release

Cali0707:
/cc @pierDipi 
Now that release is out, could you take a look again?

control-plane/pkg/kafka/clientpool/client.go
```diff
@@ -0,0 +1,238 @@
...
+
+func (c *client) LeastLoadedBroker() *sarama.Broker {
+	return c.client.LeastLoadedBroker()
+}
+
+func (c *client) Close() error {
+	// guard against panic, this makes it easier to defer client.Close() on the caller side
+	if c == nil {
+		return nil
+	}
```
pierDipi, line 210: When does this case happen?

control-plane/pkg/prober/cache.go
```diff
@@ -52,43 +52,45 @@ func (s Status) String() string {
...
+	Get(key K) (V, bool)
 	// UpsertStatus add or updates the status associated with the given key.
 	// Once the given key expires the onExpired callback will be called passing the arg parameter.
-	UpsertStatus(key string, status Status, arg interface{}, onExpired ExpirationFunc)
+	UpsertStatus(key K, value V, arg A, onExpired ExpirationFunc[K, V, A])
 	// Expire will expire the given key.
-	Expire(key string)
+	Expire(key K)
+	// Keys returns all of the keys currently in the cache that have not expired
+	Keys() []K
```
pierDipi, line 64: This is unused, why do we need this method?

control-plane/pkg/kafka/clientpool/client.go
Cali0707: Oh I've fixed this now, we can probably remove it. Basically, if you do something like:
```go
client, err := r.GetKafkaClient(...)
defer client.Close()
if err != nil {
   ...
}
```

Here, if `err` is not `nil`, the deferred `client.Close()` method panics. 

I fixed all the instances where this was happening, so I don't think we actually need this now

control-plane/pkg/prober/cache.go
Cali0707: Good catch, this was from an earlier version of the PR. Let me remove it

pierDipi (approved):
/lgtm
/approve

PR #3498, opened 2023-11-30
Fix multiple logger binding error

### LINKED ISSUE #3452: Fix Benchmarks logger multiple bindings problem/performance regression
**Describe the bug**
Since the dependency update in #3425 the filter benchmarks report the following error while being run:
```
SLF4J: Class path contains multiple SLF4J providers.
SLF4J: Found provider [ch.qos.logback.classic.spi.LogbackServiceProvider@66202ab4]
SLF4J: Found provider [org.slf4j.nop.NOPServiceProvider@5f731a2f]
SLF4J: See https://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual provider is of type [ch.qos.logback.classic.spi.LogbackServiceProvider@66202ab4]
```
The filter evaluation benchmarks are also 30-60 times slower (likely due to resolution issues with the logger).

 **Expected behavior**
The logger warning should go away. We should verify if that is what is causing the performance regression, or if the regression is independent of that change.

 **To Reproduce**
Run `./hack/run.sh BenchmarkAnyFilter` on the main branch (needs AnyFilterBenchmark added to `data-plane/resources/filter-class-list.txt`).

 **Knative release version**
1.12, next

 **Additional context**

Cali0707:
Fixes #3452

## Proposed Changes

- Instead of swapping out the "real" logger for a nop logger, just use the real logger in benchmarks. Bonus: this is a better measure of the real performance!

Testing locally, the benchmarks work now. To verify for yourself, run `./hack/run.sh benchmark-filters` and wait until the first benchmark runs, then verify that you don't get a message along the lines of:
```
SLF4J: Class path contains multiple SLF4J providers.
SLF4J: Found provider [ch.qos.logback.classic.spi.LogbackServiceProvider@66202ab4]
SLF4J: Found provider [org.slf4j.nop.NOPServiceProvider@5f731a2f]
SLF4J: See https://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual provider is of type [ch.qos.logback.classic.spi.LogbackServiceProvider@66202ab4]
```

### DIFF
--- a/data-plane/benchmarks/pom.xml
+++ b/data-plane/benchmarks/pom.xml
@@ -54,25 +54,6 @@
       <groupId>dev.knative.eventing.kafka.broker</groupId>
       <artifactId>dispatcher-vertx</artifactId>
       <version>${project.version}</version>
-      <exclusions> <!-- Exclude logging and noop -->
-        <exclusion>
-          <groupId>ch.qos.logback</groupId>
-          <artifactId>logback-core</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>ch.qos.logback</groupId>
-          <artifactId>logback-classic</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>net.logstash.logback</groupId>
-          <artifactId>logstash-logback-encoder</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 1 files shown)


Cali0707:
Do you think we can unhold this PR @pierDipi @creydr ?

PR #3524, opened 2023-12-08
[WIP] Switch to java 21 in parent pom

creydr:
As per title


data-plane/pom.xml
```diff
@@ -28,7 +28,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <java.version>15</java.version>
```
matzew, line 31: there is more, see https://github.com/knative-extensions/eventing-kafka-broker/pull/3516

PR #3562, opened 2024-01-03
Fix flaky test failures because of single partition configmaps having the same name

Cali0707:
Some tests have recently been failing in a flaky manner with errors as follows:
```
    install.go:36: failed to create resource configmaps "single-partition-kafka-broker-config" already exists - Resource:
        apiVersion: v1
        data:
          bootstrap.servers: my-cluster-kafka-bootstrap.kafka:9092
          default.topic.partitions: "1"
        kind: ConfigMap
        metadata:
          name: single-partition-kafka-broker-config
          namespace: test-dsvwxqja
```

## Proposed Changes

- Use a randomized configmap name so that the names don't conflict

### DIFF
--- a/test/e2e_new/dls_extensions_test.go
+++ b/test/e2e_new/dls_extensions_test.go
@@ -72,11 +72,13 @@ func SubscriberUnreachable() *feature.Feature {
 
 	ev := cetest.FullEvent()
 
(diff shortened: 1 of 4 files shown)


Leo6Leo (commented):
The changes look good to me!
/lgtm

PR #3566, opened 2024-01-08
Execute update-cert-manager.sh by calling bash directly

mgencur:
The script already starts with a shebang that requires "bash". It is more explicit to call "bash <script>" instead of just "<script>" and doesn't require the file to be executable.

Bash needs to be available in both cases otherwise calling the script will fail.

Fixes #

## Proposed Changes

- Call script by "bash <script>"
- Avoid calling chmod as this operation might not be allowed when running in a container with a specific non-root user

**Release Note**

```release-note

```

**Docs**


pierDipi (approved):
/lgtm
/approve

PR #3570, opened 2024-01-09
Use image digests for data plane images

pierDipi:
This ensures that even in local development when running `./hack/run.sh deploy` without a new Git commit the deployments are updated with the new image, besides the other benefits of using digests.


creydr (approved):
nice.

/lgtm

Issue #3607, opened 2024-01-15
Create experimental tests job


Cali0707:
**Problem**
To test experimental features in EKB, we should create an experimental tests job

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
An optional job which runs when `/test experimental-tests` is commented on a PR

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

Cali0707:
We only really want to test eventtype autocreate currently, which we will just enable in all tests. So, closing this for now.

PR #3609, opened 2024-01-15
Add experimental tests script

Cali0707:
Part of #3607 

## Proposed Changes

- Add a experimental-tests script which runs experimental tests after configuring the experimental features.


Cali0707:
Closing as per discussion with @pierDipi we will just enable eventtype autocreate in all tests, and re-evaluate if we see more failures in the tests

Issue #3573, opened 2024-01-09
[data-plane]: Add library for OIDC token management


creydr:
Add helpers to create and verify JWT tokens

PR #3611, opened 2024-01-16
Add OIDC TokenVerifier

creydr:
Fixes partially #3573

## Proposed Changes

- :gift: Adds a helper to verify JWT tokens and OIDC discovery.

An "in action example" of the setup and configuration can be seen at:
https://github.com/knative-extensions/eventing-kafka-broker/blob/9aafcece35118fb1290734dd9a98cb3557764d66/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java#L150-L153

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/OIDCDiscoveryConfig.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/OIDCDiscoveryConfig.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright © 2018 Knative Authors (knative-dev@googlegroups.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


creydr:
Hi @pierDipi could you PTAL?
/cc @pierDipi

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/OIDCDiscoveryConfig.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private String issuer;
+
+    private JwksVerificationKeyResolver jwksVerificationKeyResolver;
+
+    private OIDCDiscoveryConfig() {}
+
+    public String getIssuer() {
+        return issuer;
+    }
```
pierDipi, line 45: Can we make this thread safe with an atomic reference ?

If the caller calls `getIssuer` while the pipeline created by build is running we have a data race

pierDipi: Same for jwksVerificationKeyResolver

creydr: Not sure if I understand you correctly, but `build` returns a future, which resolves with the OIDCDiscoveryConfig object. So not sure how to call `getIssuer()` before.

pierDipi: Ah that's true, thanks!

PR #3631, opened 2024-01-24
Fix possible nil pointer dereferences on Consumer/ConsumerGroup

Cali0707:
Maybe helps fix the errors we are seeing with consumer/consumergroups

## Proposed Changes

- Check if the pointer fields on consumer/consumergroups are not nil before dereferencing them

### DIFF
--- a/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
+++ b/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
@@ -67,7 +67,7 @@ func (cg *ConsumerGroup) GetKey() types.NamespacedName {
 // GetVReplicas implements scheduler.VPod interface.
 func (cg *ConsumerGroup) GetVReplicas() int32 {
 	if cg.Spec.Replicas == nil {
-		return 0
+		return 1
 	}
 
 	return *cg.Spec.Replicas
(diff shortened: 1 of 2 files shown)


pierDipi:
@Cali0707 any link to the error ?

control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
```diff
@@ -66,6 +66,10 @@ func (cg *ConsumerGroup) GetKey() types.NamespacedName {
 
 // GetVReplicas implements scheduler.VPod interface.
 func (cg *ConsumerGroup) GetVReplicas() int32 {
+	if cg.Spec.Replicas == nil {
+		return 0
+	}
```
pierDipi, line 71: Usually, just like deployments, nil means 1

pierDipi: https://github.com/knative-extensions/eventing-kafka-broker/blob/e1c06aeb4adb7104365b690d2e8e0c98cec1a55b/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_defaults.go#L34-L38

Cali0707:
> @Cali0707 any link to the error ?

No this was more just a hunch as sometimes I've seen tests with very little logs if there was a panic in the controller. And the logs for the PR we have for debugging the consumergroup stuff only had 5 minutes of logs. So, I just checked all the pointers and made sure we only dereferenced them if they were not nil

matzew (approved):
/lgtm
/approve

PR #3632, opened 2024-01-25
Update contract with fields for OIDC information

### LINKED ISSUE #3571: Update contract to include Audience and OIDC Serviceaccount field
The [contract](https://github.com/knative-extensions/eventing-kafka-broker/blob/main/proto/contract.proto) between control- and data-plane needs to be configured to include the following fields for the Egress:

* Audience of the Destination
* Audience of the Reply 
* Audience of the Dead-Letter-Sink
* OIDCServiceAccountName

The Ingress needs to be configured to include the Audience of the ingress (e.g. from the Broker object).

**Additional Information:**
* contract file: https://github.com/knative-extensions/eventing-kafka-broker/blob/main/proto/contract.proto
* To update the generated files, run `./hack/generate-proto.sh` or `/hack/update-codegen.sh`

creydr:
Fixes #3571

## Proposed Changes

- Update contract to include OIDC relevant information. This includes:
  - Audience of the Destination
  - Audience of the Reply
  - Audience of the Dead-Letter-Sink
  - OIDCServiceAccountName
  - Audience of the ingress


proto/contract.proto
```diff
@@ -344,6 +355,9 @@ message Resource {
   // - setting the `source` attribute of a KafkaSource event (when it's not a CloudEvent)
   // - tagging metrics
   Reference reference = 11;
+
+  // Name of the service account to use for OIDC authentication.
+  string OIDCServiceAccountName = 12;
```
creydr, line 360: here I am not sure if the `Reference` type would be better or not. That way, we would include the namespace of the SA too, but we can get this information from resource.reference.namespace too :shrug:

pierDipi: Shouldn't this be part of "Egress" ?

creydr: Uh yes :flushed: . Will move it. Thanks for catching.

creydr:
```
feature.go:199: failed to wait for resources to be deleted: timed out waiting for the condition
```
not related.

/retest-required

pierDipi (approved):
/lgtm
/approve

PR #3634, opened 2024-01-25
Remove unneeded pkg patch

creydr:
We had a patch for pkg due to https://github.com/knative/pkg/issues/2418. This was solved in https://github.com/knative/pkg/pull/2738 therefor we can remove this patch again.

/cc @pierDipi


creydr:
/unhold
@pierDipi could you PTAL?

creydr:
@pierDipi sry. Need to recheck
/hold

PR #3642, opened 2024-01-29
Align to the defaults, list all - for convenience

matzew:
## Proposed Changes

- adding missing features to have a "complete" feature list, to easily switch _all_ existing features

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/experimental/features_config/features.yaml
+++ b/test/experimental/features_config/features.yaml
@@ -25,4 +25,8 @@ data:
   kreference-group: "disabled"
   delivery-retryafter: "disabled"
   delivery-timeout: "enabled"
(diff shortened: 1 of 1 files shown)


creydr (approved):
nit: order different as in https://github.com/knative-extensions/eventing-kafka-broker/blob/e0d1f198d2cedcaaf23119a16f1a852d6fb507a9/test/config-transport-encryption/features.yaml but LGTM

/lgtm

/hold
in case you want to have the same ordering

matzew:
@creydr it's the same order as upstream: https://github.com/knative/eventing/blob/main/config/core/configmaps/features.yaml

That's why 

/unhold

PR #3653, opened 2024-01-30
Only global resync when the replicas of the STS change

pierDipi:
This should help with reducing API server calls

### DIFF
--- a/control-plane/pkg/reconciler/consumergroup/controller.go
+++ b/control-plane/pkg/reconciler/consumergroup/controller.go
@@ -202,7 +202,32 @@ func ResyncOnStatefulSetChange(ctx context.Context, handle func(interface{})) {
 			ss := obj.(*appsv1.StatefulSet)
 			return ss.GetNamespace() == systemNamespace && kafkainternals.IsKnownStatefulSet(ss.GetName())
 		},
-		Handler: controller.HandleAll(handle),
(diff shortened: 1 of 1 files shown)


control-plane/pkg/reconciler/consumergroup/controller.go
```diff
@@ -202,7 +202,32 @@ func ResyncOnStatefulSetChange(ctx context.Context, handle func(interface{})) {
...
+					return
+				}
+
+				// This should never happen, but we check for nil to be sure.
+				if o.Spec.Replicas == nil || n.Spec.Replicas == nil {
+					return
+				}
+
+				// Only handle when replicas change
+				if *o.Spec.Replicas == *n.Spec.Replicas && o.Status.ReadyReplicas == n.Status.ReadyReplicas {
```
Cali0707, line 223: I think we currently only look at `Spec.Replicas` when scheduling, maybe we can simplify this check even further?

Cali0707 (approved):
/lgtm

@pierDipi do you think we could merge this? Seems like other things are failing, but IMO this is still a good improvement (maybe enough small fixes and consumergroups will start being faster :laughing: )

Cali0707:
/cherry-pick release-1.12

I keep forgetting you can only do one per comment :laughing:

PR #3654, opened 2024-01-31
feat: remove yaml merge tag

converge:
Fixes #

Merge keys are not officially supported in YAML according to its specifications. These keys are associated with an older version of YAML (1.1), which is now considered obsolete. Additionally, these merge keys haven't been updated or included in the current YAML version (1.2), so there's no guarantee that they will function as expected or be supported. In summary, using merge keys in YAML might not work reliably and is not recommended due to these limitations and lack of support in the latest YAML version.

I attempted to install the YAML mentioned using Kustomize, but the installation failed because Kustomize couldn't properly interpret or render the YAML file.

## Proposed Changes

- I suggest that we avoid utilizing merge tags to ensure compliance with YAML specifications, particularly in the latest versions like 1.2.x.

### DIFF
--- a/control-plane/config/eventing-kafka-broker/200-webhook/500-webhook.yaml
+++ b/control-plane/config/eventing-kafka-broker/200-webhook/500-webhook.yaml
@@ -100,7 +100,7 @@ spec:
             - name: profiling
               containerPort: 8008
 
-          readinessProbe: &probe
+          readinessProbe:
             periodSeconds: 1
             httpGet:
               scheme: HTTPS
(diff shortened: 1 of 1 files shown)


control-plane/config/eventing-kafka-broker/200-webhook/500-webhook.yaml
```diff
@@ -109,7 +109,13 @@ spec:
                 - name: k-kubelet-probe
                   value: "webhook"
           livenessProbe:
-            <<: *probe
```
matzew, line 112: we do use these in eventing core as well

matzew:
I think at general knative level we use these. Not just here 
/cc @dprotaso

dprotaso:
We dropped these in serving since we only have one version -

control-plane/config/eventing-kafka-broker/200-webhook/500-webhook.yaml
converge: > we do use these in eventing core as well

do you mean custom tag? that's true, only the source is different. in eventing-core, it comes from [eventing-core.yaml](https://github.com/knative/eventing/blob/main/config/core/deployments/webhook.yaml#L113) and the mentioned tag comes from https://github.com/knative-extensions/eventing-kafka-broker/blob/main/control-plane/config/eventing-kafka-broker/200-webhook/500-webhook.yaml#L103

matzew:
Thanks for the PR, @converge

PR #3666, opened 2024-02-02
fix selector path for cg scale subresource

Cali0707:
Currently, we don't define the scale subresource correctly on the consumergroup, as we say that the `labelSelectorPath` is `.status.selector`, while we set `.spec.selector` instead. See the code here: https://github.com/knative-extensions/eventing-kafka-broker/blob/6012e1b19b407b20286c6b7c553f7acf4ec071a0/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_defaults.go#L41-L43

I believe this should fix the "SelectorRequired" warnings we see on the scaled objects reported in https://github.com/knative/eventing/issues/7655

## Proposed Changes

- Set the .status.selector json path with the required values
-
-

**Release Note**

```release-note
:bug: the scale subresource is now correctly defined for consumergroups
```

**Docs**

### DIFF
--- a/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
+++ b/control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
@@ -144,7 +144,7 @@ type ConsumerGroupStatus struct {
 
 	// Selector is the string serialized label selector needed for the scale subresource.
 	// Defaults to ""
-	Selector string `json:"selector"`
+	Selector string `json:"selector,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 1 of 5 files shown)


Cali0707:
/hold
Looks like I need to debug this change a bit more...

Cali0707:
/unhold

The CI seems less flaky after the fix (or maybe I got lucky :shrug:), maybe the incorrect scale subresource definition was also slowing things down on the API server side @pierDipi

converge:
I have manually tested, and it works as expected! 🎉

control-plane/pkg/apis/internals/kafka/eventing/v1alpha1/consumer_group_types.go
```diff
@@ -141,6 +141,10 @@ type ConsumerGroupStatus struct {
 	// same Template, but individual replicas also have a consistent identity.
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
+
+	// Selector is the string serialized label selector needed for the scale subresource.
+	// Defaults to ""
+	Selector string `json:"selector"`
```
pierDipi, line 147:
```suggestion
	Selector string `json:"selector,omitempty"`
```

creydr (approved):
/lgtm

Thanks for fixing @Cali0707

PR #3671, opened 2024-02-05
feat: kafka broker now supports all kafka topic config options

### LINKED ISSUE #1638: Allow further configuration of topics
**Problem**
I would like to control more topic configurations than the currently exposed partitions and replication factor, such as retention period, cleanup type, and potentially 3rd-party configurations like confluent.*

* System Operator

**Exit Criteria**
I can set properties such as "cleanup.policy", "retention.ms", and "confluent.tier.enable" for the topic backed by the Broker.

**Time Estimate (optional):** 

In my mind, this seems really close to the existing config map for a topic, but with more configurations, so I would picture at most a week?

Originally requested [here](https://github.com/knative-sandbox/eventing-kafka-broker/pull/1023#issuecomment-994763143), but deserves its own issue.

Cali0707:
Fixes #1638 

## Proposed Changes

- Parse all configmap values with prefix `default.topic.config.` into the topic config entries

This approach allows us to support any future changes to the kafka topic config options automatically. The user will still get a message when they provide an incorrect config option as the kafka server will reject the create topic request telling them what went wrong:

```
lastTransitionTime: "2024-02-05T21:25:12Z"
    message: 'kafka server: Configuration is invalid - Unknown topic config name: unknown-config'
    reason: 'Failed to create topic: knative-broker-default-broker-3'
    status: "False"
    type: Ready
```

**Release Note**

```release-note
Auto created topics by brokers now support the full set of topic config options, set them by using the prefix `default.topic.config.` in your kafka-broker-config configmap. 
```

**Docs**

Will open a docs PR if this approach is accepted

### DIFF
--- a/control-plane/pkg/kafka/topic.go
+++ b/control-plane/pkg/kafka/topic.go
@@ -24,6 +24,7 @@ import (
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 	"knative.dev/pkg/configmap"
 )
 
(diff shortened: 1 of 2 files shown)


pierDipi (approved):
/lgtm
/approve

PR #3676, opened 2024-02-06
feat: remove keda objects during reconciliation

### LINKED ISSUE #3669: disabling Keda scale doesn't take effect
**Describe the bug**
[ok] Setting configmap config-kafka-features / `controller.autoscaler` from disabled to enabled takes effect.
[failing] setting configmap config-kafka-features / `controller.autoscaler` from enabled to disabled doesn't takes effect.

 **Expected behavior**
Setting configmap config-kafka-features / `controller.autoscaler` from enabled to disabled should delete related scale resources created when the flag is switched from `disabled` to `enabled`, it will include:

Keda: Scaled Object, Authentication Trigger
Others: ConsumerGroups?

 **To Reproduce**
1. configmap config-kafka-features / `controller.autoscaler` from disabled to enabled takes effect.
2. setting configmap config-kafka-features / `controller.autoscaler` from enabled to disabled doesn't takes effect.
3. kubectl get so -A

the expected behaviour is that the Keda objects would be deleted.

 **Knative release version**
1.13 (eventing and serving)

 **Additional context**

converge:
Fixes #3669 

## Proposed Changes

Remove Keda objects during reconciliation if [configmap] **config-kafka-features/controller-autoscaler-keda** is set to `disabled`.

```release-note
- remove Keda objects during reconciliation
```

* early feedback welcome / will add unit tests before marking it as ready

### DIFF
--- a/control-plane/pkg/reconciler/consumergroup/consumergroup.go
+++ b/control-plane/pkg/reconciler/consumergroup/consumergroup.go
@@ -854,45 +854,46 @@ func keyOf(cg metav1.Object) string {
 //
 // When Keda creates a scaled object, it also generates a Horizontal Pod Autoscaler. This resource will also be removed
 // when the scaled object is deleted.
-func (r *Reconciler) deleteKedaObjects(ctx context.Context) error {
+func (r *Reconciler) deleteKedaObjects(ctx context.Context, cg *kafkainternals.ConsumerGroup) error {
 
-	ns, err := r.KubeClient.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
-	if err != nil {
-		return fmt.Errorf("failed to list available namespaces: %w", err)
+	// check if scaled object exist, if not ignore
+	scaledObjectName := keda.GenerateScaledObjectName(cg)
+	_, err := r.KedaClient.KedaV1alpha1().ScaledObjects(cg.Namespace).Get(ctx, scaledObjectName, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get Keda scaled object: %w", err)
 	}
-	for _, namespace := range ns.Items {
-		// delete scaled object
-		scaledObjects, err := r.KedaClient.KedaV1alpha1().ScaledObjects(namespace.Namespace).List(ctx, metav1.ListOptions{})
+	if !apierrors.IsNotFound(err) {
+		err = r.KedaClient.KedaV1alpha1().ScaledObjects(cg.Namespace).Delete(ctx, scaledObjectName, metav1.DeleteOptions{})
 		if err != nil {
-			return fmt.Errorf("failed to list available Keda scaled objects: %w", err)
+			return fmt.Errorf("failed to delete Keda scaled object: %w", err)
 		}
+		logging.FromContext(ctx).Debugw("Keda scaled object deleted",
+			"name", scaledObjectName,
+			"namespace", cg.Namespace)
+	}
 
-		for _, scaledObject := range scaledObjects.Items {
-			err = r.KedaClient.KedaV1alpha1().ScaledObjects(scaledObject.Namespace).Delete(ctx, scaledObject.Name, metav1.DeleteOptions{})
-			if err != nil {
-				return fmt.Errorf("failed to delete Keda scaled object: %w", err)
-			}
-			logging.FromContext(ctx).Debugw("Keda scaled object deleted",
-				"name", scaledObject.Name,
-				"namespace", scaledObject.Namespace)
+	// check if trigger authentication exist , if not ignore
+	var triggerAuthName string
+	if len(cg.ObjectMeta.OwnerReferences) > 0 && cg.ObjectMeta.OwnerReferences[0].UID != "" {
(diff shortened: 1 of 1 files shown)


control-plane/pkg/reconciler/consumergroup/consumergroup.go
```diff
@@ -845,3 +848,51 @@ func metricTagsOf(ctx context.Context, cg *kafkainternals.ConsumerGroup) (contex
 func keyOf(cg metav1.Object) string {
 	return types.NamespacedName{Namespace: cg.GetNamespace(), Name: cg.GetName()}.String()
 }
+
+// deleteKedaObjects is responsible for deleting Keda (Kubernetes-based Event Driven Autoscaling) objects, "scaled
+// objects" and "trigger authentications" from all available namespaces.
+//
+// When Keda creates a scaled object, it also generates a Horizontal Pod Autoscaler. This resource will also be removed
+// when the scaled object is deleted.
+func (r *Reconciler) deleteKedaObjects(ctx context.Context) error {
```
Cali0707, line 857: It looks like this will delete all KEDA scaled objects and trigger authentications in the cluster, not just those created by the kafka-controller. Maybe we could leverage the fact that the name of the scaledobject (in keda.GenerateScaledObject) and trigger authentication (in keda.GenerateTriggerAuthentication) for each consumergroup is deterministic and we can generate it each time?

Then, since the `ReconcileKind` function will be called once per consumergroup when we update the configmap, we could:
1. Get the name of the scaledobject associated with the consumergroup and then delete it
2. Get the name of the trigger authentication associated with the consumergroup and then delete it

This way, we won't delete any extra KEDA resourced

converge: that sounds like a good idea. I tried to follow that on my last commit.

control-plane/pkg/reconciler/consumergroup/consumergroup.go
```diff
@@ -845,3 +848,52 @@ func metricTagsOf(ctx context.Context, cg *kafkainternals.ConsumerGroup) (contex
...
+	}
+	if !apierrors.IsNotFound(err) {
+		err = r.KedaClient.KedaV1alpha1().TriggerAuthentications(cg.Namespace).Delete(ctx, triggerAuthName, metav1.DeleteOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to delete Keda trigger authentication: %w", err)
+		}
+		logging.FromContext(ctx).Debugw("Keda trigger authentication deleted",
+			"name", triggerAuthName,
+			"namespace", cg.Namespace)
+	}
```
pierDipi, line 897: similar idea here, we can just call `Delete`

control-plane/pkg/reconciler/consumergroup/consumergroup.go
```diff
@@ -195,6 +195,9 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, cg *kafkainternals.Consu
 	} else {
 		// If KEDA is not installed or autoscaler feature disabled, do nothing
 		cg.MarkAutoscalerDisabled()
+		if err := r.deleteKedaObjects(ctx); err != nil {
```
pierDipi, line 198:
```suggestion
		if err := r.deleteKedaObjects(ctx, cg); err != nil {
```

control-plane/pkg/reconciler/consumergroup/consumergroup.go
converge: @pierDipi, @Cali0707 that's nice, so we don't need to care so much about validation before deleting. I have just pushed a new commit with the proposed changes.

Cali0707 (commented):
@converge I think this set of changes looks good to me, would you be able to work on updating the unit tests so that they pass with the new changes?

If you want help with that (our unit tests can take some getting used to :laughing: ), feel free to ping me on slack!

converge:
> @converge I think this set of changes looks good to me, would you be able to work on updating the unit tests so that they pass with the new changes?
> 
> If you want help with that (our unit tests can take some getting used to 😆 ), feel free to ping me on slack!

thanks xD
I pushed one commit to fix the unit test. are the unit test auto generated somehow? should I write more unit tests?

control-plane/pkg/reconciler/consumergroup/consumergroup.go
```diff
@@ -845,3 +848,45 @@ func metricTagsOf(ctx context.Context, cg *kafkainternals.ConsumerGroup) (contex
...
+	}
+
+	// delete scaled object
+	err = r.KedaClient.KedaV1alpha1().ScaledObjects(cg.Namespace).Delete(ctx, scaledObjectName, metav1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete Keda scaled object: %w", err)
+	}
+	if err == nil {
+		logging.FromContext(ctx).Infoln("Keda scaled object deleted")
+	}
```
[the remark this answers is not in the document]
converge: > @converge sorry I missed this earlier, but this should also be in the `if metav1.IsControlledBy()` statement, as we don't want to delete the scaled object if we don't control it

make sense! I have just updated, tks!

Cali0707 (approved):
/lgtm
/approve

Thanks for all your persistence on this @converge :tada:

PR #3683, opened 2024-02-08
Update deps and set statefulset scale cache config properly

Cali0707:
Recently in eventing we added support for caching the scale of statefulsets to reduce the load on the API server. This PR updates deps to include that change and also sets the config properly so that it works


matzew (approved):
/lgtm
/approve

PR #3693, opened 2024-02-12
Expose reasons for BindInProgress by adding pod status summary

pierDipi:
This will make clear the reasons for BindInProgress status, especially when pods cannot be scheduled when there are not enough resources


matzew (approved):
/lgtm
/approve

PR #3707, opened 2024-02-19
Check status code of OIDC discovery response

### LINKED ISSUE #3674: [data-plane]: Make OIDCDiscoveryConfig more resistant against invalid OIDC discovery config
Currently the loading of the OIDC config from the discovery endpoint is not very robust and fails on an invalid OIDC config (e.g. when the issuer was configured in the k8s config without https and thus the query results in a 404): https://github.com/knative-extensions/eventing-kafka-broker/blob/02be94e0e58f1755cce26af7a8d0ab8b5e57d8d3/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/OIDCDiscoveryConfig.java#L60-L84

This should be changed (e.g. status code checking) and the OIDCDiscoveryConfig fail the returned future correctly in such a case (e.g. with an additional exception, which can then be handled by the caller).

creydr:
Fixes #3674 

## Proposed Changes

- :bug: Check status code of response of OIDC discovery before setting up OIDC info to prevent JSON parse exception

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/OIDCDiscoveryConfig.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/OIDCDiscoveryConfig.java
@@ -68,7 +68,8 @@ public class OIDCDiscoveryConfig {
 
                     try {
                         if (res.statusCode() >= 400) {
-                          return Future.failedFuture("Unexpected status (" + res.statusCode() + ") on OIDC discovery endpoint: " + res.bodyAsString());
+                            return Future.failedFuture("Unexpected status (" + res.statusCode()
+                                    + ") on OIDC discovery endpoint: " + res.bodyAsString());
                         }
 
                         ObjectMapper mapper = new ObjectMapper();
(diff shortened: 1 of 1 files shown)


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/OIDCDiscoveryConfig.java
```diff
@@ -58,13 +60,18 @@ public static Future<OIDCDiscoveryConfig> build(Vertx vertx) {
...
                 .bearerTokenAuthentication(kubeConfig.getAutoOAuthToken())
                 .send()
                 .compose(res -> {
                     logger.debug("Got raw OIDC discovery info: " + res.bodyAsString());
 
                     try {
+                        if (res.statusCode() >= 400) {
+                            return Future.failedFuture("Unexpected status (" + res.statusCode()
+                                    + ") on OIDC discovery endpoint: " + res.bodyAsString());
+                        }
```
pierDipi, line 73: In this spec I read this https://openid.net/specs/openid-connect-discovery-1_0.html:

> 4.2.  OpenID Provider Configuration Response
The response is a set of Claims about the OpenID Provider's configuration, including all necessary endpoints and public key location information. A successful response MUST use the 200 OK HTTP status code and return a JSON object using the application/json content type that contains a set of Claims as its members that are a subset of the Metadata values defined in [Section 3](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata). Other Claims MAY also be returned.


Do we expect a 3xx status codes? It seems here we should fail on anything non `200`?

creydr: I only checked for failures. But can change it to `!= 200` too. Let me update it...

creydr: updated

pierDipi (approved):
/lgtm
/approve

PR #3708, opened 2024-02-19
Remove unneeded check when setting broker audience

creydr:
## Proposed Changes

- :broom: Remove unneeded check when setting broker audience


pierDipi (approved):
/lgtm
/approve

PR #3723, opened 2024-02-27
Detect trust bundles changes

pierDipi:
As per title

This will be E2E tested with tests in https://github.com/knative/eventing/pull/7726


Cali0707 (approved):
/lgtm
/approve

PR #3728, opened 2024-03-01
TokenVerifier: execute blocking calls in parallel

### LINKED ISSUE #3673: [data-plane]: TokenVerifier: execute blocking calls in parallel (`ordered = false`)
Currently we don't disable the ordering of the blocking call in https://github.com/knative-extensions/eventing-kafka-broker/blob/02be94e0e58f1755cce26af7a8d0ab8b5e57d8d3/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/TokenVerifier.java#L43-L59.
This leads to non-parallel/serial calls and thus reduced throughput. 

```
...
ordered – if true then if executeBlocking is called several times on the same context, the executions for that context will be executed serially, not in parallel. if false then they will be no ordering guarantees
```

Therefor we should check if we can use `vertx.executeBlocking(handler, ordered)` with `ordered = false`.

**Additional Information:**
* [TokenVerifier](https://github.com/knative-extensions/eventing-kafka-broker/blob/main/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/TokenVerifier.java) class (attention: might be renamed to TokenVerifierImpl in #3675)
* `executeBlocking` [java doc](https://github.com/eclipse-vertx/vert.x/blob/1cbd6622e28be37113fc6873dfc328be82fcbce6/src/main/java/io/vertx/core/Vertx.java#L559-L590)

parth721:
Fixes #3673 

Here the executeBlocking() in Vert.x framework doing serial operation, on the jwt token for authorization. Due to which as jwt token took longer to process, So I turned the default value ordered parameter to false. So, that the verification process can be done asynchronously with different verticles.


control-plane/pkg/contract/contract.pb.go
creydr, line 1: I think changes on this file are not needed for this change

creydr (changes requested):

control-plane/pkg/contract/contract.pb.go
parth721: @creydr It's automatically updated, removed the changes as suggested :)

parth721:
> Thanks a lot @parth721 for fixing this 👍

Thank you @creydr for your support  :)

PR #3730, opened 2024-03-02
add egress uuid and name to span

### LINKED ISSUE #3720: tracing: kafka-broker-dispatcher: add trigger's name and uuid to kafka-broker-dispatcher's trace tags
**Problem**
A short explanation of the problem, including relevant restrictions.
image

The 'post' operation trace generated by kafka-broker-dispatcher currently only includes the 'http.url' in its tags.
Distinguishing between traces becomes challenging when multiple triggers with different filters share the same subscriber URL.
To address this issue, I propose adding the trigger's name and UUID to the trace tags. 

I believe this is achievable since RecordDispatcherImpl class has access to the trigger's name and UUID through ```consumerVerticleContext.getEgress().getReference()```.
This information can be added to the span during creation or when invoking ```span.finish()``` or ```.fail()```.
However, I acknowledge that this is a speculative suggestion.

Event consumer (developer)

**Exit Criteria**
trace tags contains trigger uuid and trigger name tags.

**Time Estimate (optional):** 

eddy-oum:
Fixes #3720 

## Proposed Changes

- decorate current span with egress name and uuid, when we are sending event to the sink.

### DIFF
--- a/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/tracing/TracingSpan.java
+++ b/data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/tracing/TracingSpan.java
@@ -17,6 +17,7 @@ package dev.knative.eventing.kafka.broker.core.tracing;
 
 import static io.opentelemetry.api.common.AttributeKey.stringKey;
 
+import dev.knative.eventing.kafka.broker.contract.DataPlaneContract.Egress;
 import io.cloudevents.CloudEvent;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.trace.Span;
@@ -26,6 +27,8 @@ public class TracingSpan {
     private static final AttributeKey<String> MESSAGING_MESSAGE_ID = stringKey("messaging.message_id");
     private static final AttributeKey<String> MESSAGING_MESSAGE_SOURCE = stringKey("messaging.message_source");
     private static final AttributeKey<String> MESSAGING_MESSAGE_TYPE = stringKey("messaging.message_type");
+    private static final AttributeKey<String> EGRESS_UUID = stringKey("egress.uuid");
+    private static final AttributeKey<String> EGRESS_NAME = stringKey("egress.name");
 
     public static void decorateCurrentWithEvent(final CloudEvent event) {
         Span span = Span.fromContextOrNull(Context.current());
@@ -35,5 +38,14 @@ public class TracingSpan {
         span.setAttribute(MESSAGING_MESSAGE_ID, event.getId());
         span.setAttribute(MESSAGING_MESSAGE_SOURCE, event.getSource().toString());
         span.setAttribute(MESSAGING_MESSAGE_TYPE, event.getType());
+    }
+
+    public static void decorateCurrentWithEgress(final Egress egress) {
+        Span span = Span.fromContextOrNull(Context.current());
+        if (span == null) {
+            return;
+        }
+        span.setAttribute(EGRESS_UUID, egress.getReference().getUuid());
+        span.setAttribute(EGRESS_NAME, egress.getReference().getName());
     }
 }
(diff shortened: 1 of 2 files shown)


matzew:
Thanks for the PR @eddy-oum !

/assign @Cali0707

Cali0707 (commented):
Hi @eddy-oum thanks for this great PR! 

One question: do you think it is worth adding these attributes to the span when there is a reply as well? Currently, we decorate the span with the reply event in the `BaseResponseHandler` class. Since you opened the issue, up to you whether or not you think this is worth handling somewhere. If you don't think it's worth handling let me know and we can approve these changes as is, otherwise I guess we should add it to the response handler as well.

eddy-oum:
@Cali0707 
Hi thanks for the suggestion!

I do agree that it would be nice to add these attributes when there is a reply; but I have a few questions regarding current span behavior.
```BaseResponseHandler::handle``` seems to share the same span with ```WebClientCloudEvent::send```, is this intended?

image

I tried adding ```TracingSpan.decorateCurrentWithRespondent(egress)``` at ```BaseResponseHandler::handle```.
Note the respondent.name="trigger-test" at the kafka_recieve span.

As you can see, ```BaseResponseHandler::handle``` seems to share the kafka_recieve span with ```WebClientCloudEvent::send```.

1) the egress.uuid="9fa.." tag at the kafka_recieve span is also present.
2) messaging.message_[id, source, type] tags are overwritten by the incoming reply's cloudevent properties, when it should be equal to that of kafka-broker-receiver's post span.

If the two methods sharing the same span is intentional:
1) BaseResponseHandler does not have to add egress tags because it is already present.
2) messaging.message_[id, source, type] should not be overwritten by reply's cloudevent? ```BaseResponseHandler``` should probably set response.message_[id, source, type].

**Comment by Cali0707** [comment]
@eddy-oum I'll be honest, I haven't done much work on the tracing spans in this repo before. I think it makes sense for the spans to be shared though, so let's stick with the changes you have here for now. If you think it makes sense to start a separate span on the reply or to set `response.message[id, source, type]` instead of overriding `message.message[id, source, type]` or would like to start a discussion about it (I think that could be a good idea!), could you open a new issue for that and we can see what others think as well?

**Review by Cali0707 — APPROVED** [review]
/lgtm
/approve

**Comment by eddy-oum** [comment]
@Cali0707 
Sure, I agree it's better to handle that in a seperate issue. I will leave this pr as is.
When I have time, I will look into the tracing code some more and, if necessary, open a new issue regarding reply handler's span. 

Thanks!

File: data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/tracing/TracingSpan.java [hunk 1]
```diff
@@ -26,6 +27,8 @@ public class TracingSpan {
     private static final AttributeKey<String> MESSAGING_MESSAGE_ID = stringKey("messaging.message_id");
     private static final AttributeKey<String> MESSAGING_MESSAGE_SOURCE = stringKey("messaging.message_source");
     private static final AttributeKey<String> MESSAGING_MESSAGE_TYPE = stringKey("messaging.message_type");
+    private static final AttributeKey<String> EGRESS_UUID = stringKey("egress.uuid");
+    private static final AttributeKey<String> EGRESS_NAME = stringKey("egress.name");
```
pierDipi, line 31: Can we use `consumer_name` and `consumer_uuid` so that it's not using internal terminology that end users are not familiar with? 

It will also help with being consistent with metrics tags

pierDipi (commented):
Thanks @eddy-oum, this is very helpful!

I left a comment

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/tracing/TracingSpan.java
eddy-oum: sure, that's a much better name.

eddy-oum: I renamed egress to consumer. thanks

Cali0707:
@eddy-oum from the looks of the failures, I think you've just gotten unlucky with our flakier tests failing on you a lot :(

/retest-required

Issue #2994, opened 2023-03-01
"java.util.ConcurrentModificationException: KafkaConsumer is not safe for multi-threaded access" errors in dispatcher logs


maschmid:
**Describe the bug**
Spotted a CME exception in kafka-channel-dispatcher logs

```
{
  "@timestamp": "2023-03-01T06:16:42.515Z",
  "@version": "1",
  "message": "Consumer exception context={topics=[knative-messaging-kafka.kc-stable-delay-0.sender0], consumerGroup='kafka.kc-stable-delay-0.sender0.f61db18a-760e-4d72-825f-ed7eab799962', reference=uuid: \"f61db18a-760e-4d72-825f-ed7eab799962\"\nnamespace: \"kc-stable-delay-0\"\nname: \"connection0\"\n}",
  "logger_name": "dev.knative.eventing.kafka.broker.dispatcher.impl.consumer.ConsumerVerticle",
  "thread_name": "vert.x-kafka-consumer-thread-2317",
  "level": "ERROR",
  "level_value": 40000,
[...]

PR #3734, opened 2024-03-06
Match API to our Strimzi cluster version for Apache Kafka

matzew:
## Proposed Changes

- updating the clients version to the version of our testing cluster version.

**Release Note**

```release-note

```

**Docs**


matzew:
Related to https://issues.apache.org/jira/browse/KAFKA-14771
The issue was on the older `3.2.3` version, but fixed in `3.5.x`

So with the bump to the matching `3.6.x` we get this too.

Relates to #2994

matzew:
/close
doing this in https://github.com/knative-extensions/eventing-kafka-broker/pull/3714  since quarkus has same version there as its base

PR #3735, opened 2024-03-06
Replace bash script log collector with a more resilient go based log collector

Cali0707:
Currently, the e2e tests just use `kubectl logs -f ...` to collect logs. This leads to issues if the pod goes away, as no more logs are collected. We have been seeing this in CI recently with the logs from the kafka-controller only including about 5 minutes worth of logs.

## Proposed Changes

- Use a go based logging solution that picks up on any new pods and starts gathering those logs too.
- Put this into a separate directory so that we can ensure that this works before fully switching the logging solution used in CI

### DIFF
--- a/test/pkg/logging/logging.go
+++ b/test/pkg/logging/logging.go
@@ -116,12 +116,12 @@ func (l *logger) Start() error {
 	return nil
 }
 
-func (l *logger) startForPod(pod *corev1.Pod, artifactsDirt string) {
+func (l *logger) startForPod(pod *corev1.Pod, artifactsDir string) {
 	for _, container := range pod.Spec.Containers {
 		podNs, podName, containerName := pod.Namespace, pod.Name, container.Name
 
 		go func() {
-			f, err := os.OpenFile(fmt.Sprintf("%s/new-logs/%s-%s.log", artifactsDirt, podName, containerName), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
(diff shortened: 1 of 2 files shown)


test/pkg/logging/logging.go
```diff
@@ -0,0 +1,169 @@
...
+						watchedPods.Delete(pod.Name)
+					case watch.Added, watch.Modified:
+						if !watchedPods.Has(pod.Name) && isPodReady(pod) {
+							label, ok := pod.Labels["app"]
+							if !ok {
+								continue
+							}
+							fileWriter, ok := fileWriters[label]
+							if !ok {
+								f, err := os.OpenFile(fmt.Sprintf("%s/new-logs/%s.log", artifacts, label), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
creydr, line 115: Can we create a separate file for each container in the pod? And also when a new pod starts, creating a new file (e.g. having the pod name in the filename)? I think this makes the usage a bit easier (IMO)

Cali0707:
>Maybe this is something what we could have the core in eventing-core (or pkg/reconciler-test repo?!?) and then reuse it in different repos. But I am fine with starting here and then tweaking it.

@creydr yeah I think I'd just like to start this here, and then when we know it works and are happy with it, move it somewhere more general.

creydr (commented):
Looks good for me.
/lgtm

Only one thing to maybe improve (can be done on a separate PR): add the pod start time into the filename, so the order of the pod runs can be seen directly

Cali0707:
>  can we remove the other one? and perhaps just use logs as directory name?

@pierDipi for suure, we just need to add the setup/teardown calls to this new package to all of the testing `main` functions first! I was planning on doing that in a separate PR after people were happy with the changes here

pierDipi (approved):
Sounds good!

/lgtm
/approve

PR #3738, opened 2024-03-07
Use error log level when contract parsing fails

creydr:
As per title to have this issue not missed.


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/eventbus/ContractPublisher.java
```diff
@@ -98,7 +98,7 @@ public DataPlaneContract.Contract parseFromJson(final Reader content) throws IOE
             this.parser.merge(content, contract);
             return contract.build();
         } catch (final InvalidProtocolBufferException ex) {
-            logger.debug("failed to parse from JSON", ex);
+            logger.error("failed to parse from JSON", ex);
```
matzew, line 101: +1

matzew (approved):
/lgtm
/approve

PR #3754, opened 2024-03-12
Unpin CE Kafka protocol version from older version

matzew:
## Proposed Changes

- Unpin CE kafka protocol from oder version, and use latest

**Release Note**

```release-note
:broom: Unpin CE kafka protocol from oder version, and use latest
```

**Docs**


creydr (approved):
/lgtm

PR #3757, opened 2024-03-13
Update to K8s 1.28 in create-kind-cluster.sh

creydr:
As per title, as 1.28 is the [default min version](https://github.com/knative/pkg/blob/2c15a6fd07afbf402f8aa64799c4bd66cb8e22c4/version/version.go#L36)


creydr:
@matzew @pierDipi could you PTAL?

matzew (approved):
/lgtm
/approve

PR #3761, opened 2024-03-14
Cancel receiver pod start on invalid OIDC config only if authentication.oidc is enabled

creydr:
Currently we cancel the receivers pod start when the OIDC configuration could not get loaded properly (most likely the case on an invalid configuration). This happens also, when OIDC is not enabled.
This PR addresses it and only cancels the receiver pods start on an invalid config if OIDC is enabled. Otherwise it logs a warning, as this could lead to issues later, when the OIDC flag gets enabled.


data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/TokenVerifierImpl.java
```diff
@@ -44,6 +45,13 @@ public Future<JwtClaims> verify(String token, String expectedAudience) {
                 promise -> {
                     // execute blocking, as jose .process() is blocking
 
+                    if (oidcDiscoveryConfig == null) {
+                        promise.fail(
+                                "OIDC discovery config not initialized. This is most likely the case when the pod was started with an invalid OIDC config in place and then later the "
+                                        + FeaturesConfig.KEY_AUTHENTICATION_OIDC
+                                        + " flag was enabled. Restarting the pod should help.");
```
Cali0707, line 52: Would using the `FileWatcher` to rebuild the config solve the problem of needing to restart the pod?

creydr: It could. Anyhow changing something fundamental like the OIDC config will probably require some other restarts as well.

creydr:
@Cali0707 @pierDipi could you PTAL when you find some time?
This should help us to unblock https://github.com/knative-extensions/eventing-istio/pull/262 too

Cali0707 (approved):
/lgtm

@creydr would it be possible to add a unit test to the `FeaturesConfig`, as a follow up (I don't think this needs to block this PR)

PR #3766, opened 2024-03-15
[e2e test] for broker topic and trigger consumer id templates being used

maschmid:
https://github.com/knative-extensions/eventing-kafka-broker/issues/3328 has a unit test, but we don't have an e2e test that would verify that it is indeed possible to configure Kafka ACLs that would allow only topic and consumer IDs with the defined prefixes. 

## Proposed Changes

- An e2e test for https://github.com/knative-extensions/eventing-kafka-broker/issues/3328

### DIFF
--- a/test/kafka/kafka_setup.sh
+++ b/test/kafka/kafka_setup.sh
@@ -33,8 +33,11 @@ kubectl -n kafka apply -f $(dirname $0)/kafka-ephemeral.yaml || kubectl -n kafka
 echo "Create TLS user"
 kubectl apply -n kafka -f $(dirname $0)/user-tls.yaml
 
-echo "Create SASL SCRUM 512 user"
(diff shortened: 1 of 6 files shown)


test/kafka/kafka_setup.sh
```diff
@@ -33,9 +33,12 @@ kubectl -n kafka apply -f $(dirname $0)/kafka-ephemeral.yaml || kubectl -n kafka
 echo "Create TLS user"
 kubectl apply -n kafka -f $(dirname $0)/user-tls.yaml
 
-echo "Create SASL SCRUM 512 user"
```
matzew, line 36: SCRUM :sweat_smile:

matzew (approved):
/lgtm
/approve

/retest

matzew:
Thanks a lot, @maschmid

PR #3775, opened 2024-03-19
Avoid possible nil pointer dereferences in consumergroup scheduling

Cali0707:
Maybe fixes the panic seen in https://storage.googleapis.com/knative-prow/pr-logs/pull/knative-extensions_eventing-kafka-broker/3558/upgrade-tests_eventing-kafka-broker_main/1769808258517700608/artifacts/new-logs/kafka-controller-8565b5bfd7-ffw6g-controller-2024-03-18%2019:58:12.178049569%20+0000%20UTC%20m=+537.848531106.log

## Proposed Changes

- Double check that variables which could be nil are not nil before accessing properties on them

### DIFF
--- a/control-plane/pkg/reconciler/consumergroup/consumergroup.go
+++ b/control-plane/pkg/reconciler/consumergroup/consumergroup.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"math"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/IBM/sarama"
@@ -125,6 +124,14 @@ func init() {
 		panic(err)
 	}
 }
+
+type NoSchedulerFoundError struct{}
+
+func (NoSchedulerFoundError) Error() string {
+	return "no scheduler found"
+}
+
+var _ error = NoSchedulerFoundError{}
 
 type Scheduler struct {
 	scheduler.Scheduler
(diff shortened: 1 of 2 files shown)


control-plane/pkg/reconciler/consumergroup/consumergroup.go
```diff
@@ -232,7 +233,7 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, cg *kafkainternals.Consum
 	cg.Spec.Replicas = pointer.Int32(0)
 	err := r.schedule(ctx, cg) //de-schedule placements
 
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "no scheduler found") {
```
pierDipi, line 236: instead of contains on error message I think using `errors.Is(...)` with a custom error is more idiomatic and robust

pierDipi: Can we also add a test in consumergroup_test.go

control-plane/pkg/reconciler/consumergroup/controller.go
```diff
@@ -120,7 +120,7 @@ func NewController(ctx context.Context, watcher configmap.Watcher) *controller.I
 	clientPool := clientpool.Get(ctx)
 
 	r := &Reconciler{
-		SchedulerFunc:                      func(s string) Scheduler { return schedulers[strings.ToLower(s)] },
+		SchedulerFunc:                      func(s string) (Scheduler, bool) { sched, ok := schedulers[strings.ToLower(s)]; return sched, ok },
```
pierDipi, line 123: nit: would this work?
```suggestion
		SchedulerFunc:                      func(s string) (Scheduler, bool) { return schedulers[strings.ToLower(s)] },
```

Cali0707: I tried that and sadly no, it looks like go complains:
```
not enough return values have (Scheduler) want (Scheduler, bool)
```

pierDipi (approved):
/lgtm
/approve

PR #3779, opened 2024-03-21
fix: triggers use correct consumer group id generated from template

Cali0707:
Currently, the triggers will always use a consumer group id generated from the default template. Instead, they should use the most up to date version of the template.

## Proposed Changes

- Add a rekt test to test for regressions here
- Watch the configmap for changes

**Release Note**

```release-note
:bug: triggers now correctly generate the consumer group id from the template, rather than just the default template
```

**Docs**


test/rekt/resources/configmap/configkafkafeatures/configKafkaFeatures.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	return func(ctx context.Context, t feature.T) {
+		if _, err := manifest.InstallYamlFS(ctx, yaml, cfg); err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+func WithTriggerConsumerGroupIDTemplate(template string) manifest.CfgFn {
```
matzew, line 44: :+1:

matzew (approved):
/lgtm
/approve

/retest

Cali0707:
Looks like that may have been a real failure

/cc @matzew

matzew (approved):
/lgtm
/approve

PR #3818, opened 2024-04-05
Capture reason for global resync

pierDipi:
As per title, helps with debugging flakes


matzew (approved):
/lgtm
/approve

Cali0707:
@pierDipi do we still need this PR with the new flake fixes we merged?

PR #3823, opened 2024-04-05
cleanup: drop brokerv2 config

Cali0707:
We no longer need brokerv2 data plane config resource, as we have switched to using the brokerv2 implementation already. This PR cleans up the resource


Cali0707:
/retest

Looks like kafka failed to start

pierDipi (approved):
/lgtm
/approve

/retest

PR #3829, opened 2024-04-10
Remove usage of Vertx.currentContext()

pierDipi:
We're seeing errors like:

```
\"io.vertx.core.Vertx.currentContext()\" is null\n\tat dev.knative.eventing.kafka.broker.core.security.KubernetesAuthProvider.getCredentials(KubernetesAuthProvider.java:38)
```


matzew (approved):
/lgtm
/approve

maschmid:
I wonder why are the eventing and kafka controllers and webhooks being restarted so often during the tests?  (If that's what's happening... I just see suspiciously many pod instances in the logs here:

https://gcsweb.knative.dev/gcs/knative-prow/pr-logs/pull/knative-extensions_eventing-kafka-broker/3829/reconciler-tests_eventing-kafka-broker_main/1778077840277770240/artifacts/new-logs/

maschmid:
OK, the latest run seems to have failed due to webhook being down
```
    feature.go:224: Warning, failed to delete test-izvxwjyo/subscription-0000-ibdobkrh of GVR: messaging.knative.dev/v1, Resource=subscriptions: Internal error occurred: failed calling webhook "validation.webhook.eventing.knative.dev": failed to call webhook: Post "https://eventing-webhook.knative-eventing.svc:443/resource-validation?timeout=10s": EOF
```

The error will not be retried AFAICS https://github.com/knative-extensions/eventing-kafka-broker/blob/main/vendor/knative.dev/reconciler-test/pkg/feature/feature.go#L224  , so it will just timeout on " failed to wait for resources to be deleted: timed out waiting for the condition", because it will forever be "still present"

matzew:
GET IN!!

PR #3830, opened 2024-04-11
fix: evictor no longer has NPE when called with nil pod

Cali0707:
Fixes bug where kafka controller panics due to the evictor accessing properties on a nil pod.

## Proposed Changes

- Don't use the pod variable, as we only use it to decorate the logger currently.
- Add a unit test to catch regressions


pierDipi (approved):
/lgtm
/approve

matzew (approved):

/approve

/lgtm

PR #3845, opened 2024-04-22
Add `.spec.subscribers[].name` field in KafkaChannel to align with eventing-core channel definition

creydr:
In https://github.com/knative/eventing/pull/7733 we added the `.spec.subscribers[].name` field for Channels. This PR adjusts the KafkaChannel to support this field too.

This helps with tests issues from https://github.com/knative-extensions/eventing-kafka-broker/pull/3770


pierDipi (approved):
/lgtm
/approve

PR #3851, opened 2024-04-23
Add debug logs for ConsumerGroup reconciler to record actions taken

pierDipi:
Helps with debugging

### DIFF
--- a/control-plane/pkg/reconciler/consumergroup/consumergroup.go
+++ b/control-plane/pkg/reconciler/consumergroup/consumergroup.go
@@ -185,20 +185,27 @@ type Reconciler struct {
 }
 
 func (r *Reconciler) ReconcileKind(ctx context.Context, cg *kafkainternals.ConsumerGroup) reconciler.Event {
+	logger := logging.FromContext(ctx)
+	logger.Debugw("Reconciling consumergroup")
+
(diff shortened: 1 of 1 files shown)


control-plane/pkg/reconciler/consumergroup/consumergroup.go
```diff
@@ -185,20 +185,27 @@ type Reconciler struct {
 }
 
 func (r *Reconciler) ReconcileKind(ctx context.Context, cg *kafkainternals.ConsumerGroup) reconciler.Event {
+	logger := logging.FromContext(ctx)
```
creydr, line 188: nit: could we add the name&namespace of the consumergroup, so we can see it directly from the logs?

creydr (commented):
/lgtm

/hold
in case you want to address the nit

control-plane/pkg/reconciler/consumergroup/consumergroup.go
pierDipi: It's already there

pierDipi: injected loggers are pre-configured with a few standard keys, see knative/pkg for details

Cali0707:
Thanks @pierDipi !!

PR #3853, opened 2024-04-23
fix: flush all records before closing the producer

Cali0707:
Fixes a possible data race: we don't always flush all records before we stop receiving them.

## Proposed Changes

- After we stop allowing records to be sent, we should call flush one more time on the producer

### DIFF
--- a/data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
+++ b/data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
@@ -139,6 +139,8 @@ public class LoomKafkaProducer<K, V> implements ReactiveKafkaProducer<K, V> {
                 sendFromQueueThread.interrupt();
                 logger.debug("Waiting for sendFromQueueThread thread to complete");
                 sendFromQueueThread.join();
+                logger.debug("flushing the producer");
+                producer.flush();
                 logger.debug("Closing the producer");
                 producer.close();
                 closePromise.complete();


data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
```diff
@@ -130,6 +130,7 @@ public Future<Void> close() {
                 }
                 sendFromQueueThread.interrupt();
                 sendFromQueueThread.join();
+                producer.flush();
```
pierDipi, line 133: We call flush before closing here https://github.com/knative-extensions/eventing-kafka-broker/blob/49ad678b4a629be09ad40f9bee1a7c360837102a/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/IngressProducerReconcilableStore.java#L253-L258

The idea of this class is to mimic the core Kafka producer interface and expose "flush" and "close" as separate methods

Cali0707: Ah that makes sense - I think our issue is that we need to flush after we stop receiving requests with events, but before we close the producer. Currently, we:
1. Flush
2. Stop receiving events
3. Close the producer
This leads to a data race which causes a whole bunch of metadata requests (which aren't present now that we have this change)


I think we could fix this f we updated the producer to either:
1. Have another method on the interface to "mark the producer as closed" and then called `markedClosed -> flush -> close`, or
2. Have the `flush` method set the closed variable so that the producer stops receiving events

creydr (commented):
/lgtm

Cali0707:
/hold

The warnings are still present in the newest set of logs, looks like this didn't fix it properly and earlier tests we were just lucky :(

data-plane/receiver-loom/src/main/java/dev/knative/eventing/kafka/broker/receiverloom/LoomKafkaProducer.java
pierDipi: > I think we could fix this f we updated the producer to either:
> 
> 1. Have another method on the interface to "mark the producer as closed" and then called `markedClosed -> flush -> close`, or
> 2. Have the `flush` method set the closed variable so that the producer stops receiving events

Isn't this exactly what the returned future communicates to the caller ?

Cali0707: Yeah, that is what is happening currently. I guess what I meant was to set closed to true in the `flush` method we have in the interface, or call flush again here

Cali0707:
/cc @creydr @pierDipi 

I think we still want to have this change, otherwise there is technically a case where we close the producer while there are still pending sends (since we wait for the `eventQueue` to be empty, but that only means we processed the event, not that the event's promise completed)

pierDipi (approved):
/lgtm
/approve

PR #3860, opened 2024-04-26
Expose test package to reuse

dsimansk:
Fixes #

## Proposed Changes

- Expose test package to reuse

I'd like to refactor `eventing-autoscaler-keda` to depend on `e-k-b` and remove `eventing-broker` from dependencies. Hence the test package is reused there.

### DIFF
--- a/test/placeholder.go
+++ b/test/placeholder.go
@@ -1,0 +1,17 @@
+/*
+Copyright 2024 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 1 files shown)


pierDipi:
> I'd like to refactor `eventing-autoscaler-keda` to depend on `e-k-b` and remove `eventing-broker` from dependencies. Hence the test package is reused there.

can you describe more about the why? Ideally, eventing-autoscaler-keda, doesn't have the integration with Kafka resources anymore as we're adding them in this repository

cc @Cali0707

dsimansk:
> > I'd like to refactor `eventing-autoscaler-keda` to depend on `e-k-b` and remove `eventing-broker` from dependencies. Hence the test package is reused there.
> 
> can you describe more about the why? Ideally, eventing-autoscaler-keda, doesn't have the integration with Kafka resources anymore as we're adding them in this repository
> 
> cc @Cali0707

There's a smoke test setup, that uses old `e-k` bits. Perhaps the whole setups can be streamlined here?
https://github.com/knative-extensions/eventing-autoscaler-keda/blob/main/test/e2e/smoke_test.go#L51-L52

dsimansk:
Streamline or maybe completely removed. It's really rather simplistic smoke test, as far as I can tell. :)

dsimansk:
Maybe it will work even without the placeholder package. Either `Golang` or `Goland` is playing games with me. :)
/hold

dsimansk:
It's not needed in such form.
/close

Issue #3815, opened 2024-04-04
Add post install logic for namespaced brokers


Cali0707:
**Problem**
We currently only delete broker deployments in the system namespace. However, namespaced brokers could have deployments in any namespaced. We should update our post install logic to clean those up as well.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve? 1

**Additional context (optional)**
Add any other context about the feature request here.

NeerajNagure:
Hi @Cali0707 , my approach to do this would be to get the list of namespaces for the broker by `
    namespaceList,err = k.k8s.CoreV1().Namespaces().List(ctx,metav1.ListOptions{})` over here [https://github.com/knative-extensions/eventing-kafka-broker/blob/964191004c226371d657faaf4583d19c64d8c5c5/control-plane/cmd/post-install/kafka_broker_deployment_deleter.go#L49C1-L65C2](https://github.com/knative-extensions/eventing-kafka-broker/blob/964191004c226371d657faaf4583d19c64d8c5c5/control-plane/cmd/post-install/kafka_broker_deployment_deleter.go#L49C1-L65C2)   and then iterating over the namespaceList items and delete the deployments for those namespace items.Am I right in this approach?

Cali0707:
Hi @NeerajNagure that sounds reasonable to me, @pierDipi or @creydr any thoughts? My only concern (and I'm not sure how to address this) is that afaik the kafka-broker-dispatcher/kafka-broker-receiver deployments in other namespaces don't have an owner reference, so we can't tell if we own that deployment or not.

pierDipi:
I think we need to go back at using deployments for namespaced brokers

PR #3889, opened 2024-05-09
Add post install logic for namespaced brokers

pierDipi:
We currently only delete broker deployments in the system namespace. However, namespaced brokers could have deployments in any namespace. 

Fixes [#3815](https://github.com/knative-extensions/eventing-kafka-broker/issues/3815)

### DIFF
--- a/control-plane/config/post-install/200-controller-cluster-role.yaml
+++ b/control-plane/config/post-install/200-controller-cluster-role.yaml
@@ -20,4 +20,19 @@ metadata:
   name: knative-kafka-controller-post-install
   labels:
     app.kubernetes.io/version: devel
-rules: []
+rules:
+  # we need to be able to delete old deployments
+  - apiGroups:
+      - "apps"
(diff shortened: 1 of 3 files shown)


control-plane/config/post-install/200-controller-cluster-role.yaml
```diff
@@ -20,4 +20,19 @@ metadata:
...
     app.kubernetes.io/version: devel
-rules: []
+rules:
+  # we need to be able to delete old deployments
+  - apiGroups:
+      - "apps"
+    resources:
+      - "deployments"
+    verbs:
+      - "delete"
```
Cali0707, line 30:
```suggestion
      - "delete"
      - "list"
```
I think we need this to be able to make our list call in the post install code

pierDipi:
@Cali0707 does this look good?

Cali0707 (approved):
/lgtm

PR #3892, opened 2024-05-11
test: add unit test for GenerateScaledObjectName

converge:
Fixes #

## Proposed Changes

- add unit test for `GenerateScaledObjectName`

note: this is a PR to validate how unit tests should be written for knative. Upcoming unit test is for Keda integration.

### DIFF
--- a/control-plane/pkg/autoscaler/keda/resources_test.go
+++ b/control-plane/pkg/autoscaler/keda/resources_test.go
@@ -1,0 +1,45 @@
+package keda
+
+import (
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"knative.dev/eventing-kafka-broker/control-plane/pkg/autoscaler/keda/mocks"
+	"strings"
+	"testing"
+)
+
+func TestGenerateScaledObjectName(t *testing.T) {
+	type args struct {
+		obj v1.Object
+	}
+	tests := []struct {
+		name       string
+		args       args
+		wantPrefix string
+		wantSuffix types.UID
+	}{
+		{
+			name: "create scaled object name successfully",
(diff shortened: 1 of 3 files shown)


control-plane/pkg/autoscaler/keda/resources_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestGenerateScaledObjectName(t *testing.T) {
+	type args struct {
+		obj v1.Object
+	}
+	tests := []struct {
+		name       string
+		args       args
+		wantPrefix string
+		wantSuffix types.UID
+	}{
```
Cali0707, line 37: Nice use of the table testing pattern! This is how we normally do unit tests in eventing (as I'm sure you noticed)

Cali0707 (commented):
This looks good @converge - thanks! 

I'm not very familiar with when we add mocks to the tests in this repo, so I'm going to let @pierDipi take a look at the test mocks before we approve this

/assign @pierDipi

pierDipi (commented):
Did you consider having a test in https://github.com/knative-extensions/eventing-kafka-broker/blob/main/control-plane/pkg/reconciler/consumergroup/consumergroup_test.go that would assert that the `ConsumerGroup` reconciler creates all KEDA objects when the feature flag is enabeled?

Such test would cover also the `GenerateScaledObjectName` since we need to explicitly indicate the object names (and spec), similar to how is done here for `Consumer`s

https://github.com/knative-extensions/eventing-kafka-broker/blob/865fea6b7d4f16fe42a256016ed2934d0c7ed65f/control-plane/pkg/reconciler/consumergroup/consumergroup_test.go#L112-L137


The benefit is that we don't need to have mocks, while having the benefit of a "unit test" that is fast and doesn't interact with the real k8s API

PR #3896, opened 2024-05-14
feat: receiver verticle now has a grace period before closing

### LINKED ISSUE #1684: Waits for service endpoints propagation on shutdown in Receiver
**Problem**

The receiver shuts down immediately upon receiving a sigterm and since we
close the HTTP server immediately and the endpoints of the service isn't
yet fully propagated in the cluster there might be requests dropped.

To make the receiver more resilient we need to wait `x` seconds before
shutting down the server and in the deployment configuration
we need to specify a `terminationGracePeriodSeconds > x`.

Event Producer

**Time Estimate (optional):** 
1

/area data-plane
/area broker
/area channel
/area sink

Cali0707:
Fixes #1684 

## Proposed Changes

- Add parameter to ReceiverVerticleFactory for how long to wait before closing the http servers in the receiver
- Use timeout in the receiver stop handler to wait until the grace period has finished before closing the servers
-

**Release Note**

```release-note
The receiver http/https servers now wait for 20 seconds after receiving SIGTERM before closing, helping with more graceful shutdown
```

### DIFF
--- a/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
+++ b/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
@@ -97,6 +97,8 @@ public class ReceiverVerticle extends AbstractVerticle implements Handler<HttpSe
     private IngressProducerReconcilableStore ingressProducerStore;
     private FileWatcher secretWatcher;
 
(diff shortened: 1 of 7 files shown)


lamluongbinh:
Hi @Cali0707, I wanted to check on the status of this MR. 

We found some timeout errors occurring while the Knative receiver is scaling, and while pods are relocated to another node during an upgrade. It seems we need to implement graceful shutdown handling in the receiver.

PR #3915, opened 2024-05-31
Prevent generation only updates

pierDipi:
This guards and discard contract ConfigMap updates when only `generation`
changes.

While this approach is technically slightly slower, the latency is negligible,
and it is more robust and easier to maintain.

## Proposed Changes

- guards and discard contract ConfigMap updates when only `generation`
changes
- move change detection to the shared library

**Release Note**

```release-note
Prevent unnecessary ConfigMap updates and reduce API server load
```

**Docs**

### DIFF
--- a/control-plane/pkg/reconciler/trigger/trigger_test.go
+++ b/control-plane/pkg/reconciler/trigger/trigger_test.go
@@ -1060,7 +1060,9 @@ func triggerReconciliation(t *testing.T, format string, env config.Env, useNewFi
 						},
 					},
 				}, env.DataPlaneConfigMapNamespace, env.ContractConfigMapName, env.ContractConfigMapFormat),
-				BrokerDispatcherPod(env.SystemNamespace, nil),
+				BrokerDispatcherPod(env.SystemNamespace, map[string]string{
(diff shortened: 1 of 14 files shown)


control-plane/pkg/reconciler/trigger/trigger_test.go
```diff
@@ -1060,7 +1060,9 @@ func triggerReconciliation(t *testing.T, format string, env config.Env, useNewFi
 						},
 					},
 				}, env.DataPlaneConfigMapNamespace, env.ContractConfigMapName, env.ContractConfigMapFormat),
-				BrokerDispatcherPod(env.SystemNamespace, nil),
+				BrokerDispatcherPod(env.SystemNamespace, map[string]string{
+					base.VolumeGenerationAnnotationKey: "0",
+				}),
```
pierDipi, line 1065: This is because the test wants to test "no operations" and without an annotation now the control plane will set the annotation on the data plane so that the CM and the annotation is fully consistent

creydr:
Putting on hold in case you want some other reviews too
/hold

mgencur:
I'm not an expert here but it looks good.

Issue #3939, opened 2024-06-21
keda autoscaler KafkaSource with no authentication produces ScaledObject with invalid .spec.triggers.authenticationRef


maschmid:
**Describe the bug**
With `controller-autoscaler-keda: enabled` ,
Having a KafkaSource with no kafka authentication, the resulting scaled object has 

```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
spec:
  triggers:
  - authenticationRef:
      name: "" 
...
```

ie, a trigger with an empty name.

This causes errors in KEDA, as it tries to find a TriggerAuthentication with an empty string as a name.

```
ERROR    Failed to get TriggerAuthentication    {"controller": "scaledobject", "controllerGroup": "keda.sh", "controllerKind": "ScaledObject", "ScaledObject": {"name":"so-5946175e-fec9-42a5-beb5-fc420f9dbf2b","namespace":"ksc-zero-3"}, "namespace": "ksc-zero-3", "name": "so-5946175e-fec9-42a5-beb5-fc420f9dbf2b", "reconcileID": "50c8d0ab-d171-4cd1-914e-223945b48218", "error": "TriggerAuthentication.keda.sh \"\" not found"} 
```

 **Expected behavior**
KafkaSource with no authentication should not cause errors in KEDA.

 **To Reproduce**

Setup KEDA, enable `controller-autoscaler-keda: enabled` in kafka-features ConfigMap

Create a KafkaSource with no authentication, e.g.

```
apiVersion: sources.knative.dev/v1beta1
kind: KafkaSource
metadata:
  name: kafka-source
spec:
  consumerGroup: knative-group
  bootstrapServers:
  - my-cluster-kafka-bootstrap.kafka:9092
  topics:
  - knative-demo-topic
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: receiver
```

Notice the created ScalingOjbect authenticationRef  and the errors in the KEDA operator. 

 **Knative release version**
1.12

 **Additional context**
Installed using https://github.com/openshift-knative/serverless-operator release-1.33 on OpenShift

PR #3944, opened 2024-06-21
test: verify that scaledobjects have no authenticationref when it is empty

Cali0707:
As a follow up on #3939 this PR adds a regression test to prevent this issue from coming back

## Proposed Changes

- Verify that the scaledobject created by a KafkaSource's consumergroup does not have an authentication ref unless it needs to


Cali0707:
@pierDipi should we backport this along with the fixes or is it fine for this test to only be in `main` for now?

pierDipi (approved):
/lgtm
/approve

pierDipi:
Yes, @Cali0707

PR #3950, opened 2024-06-25
fix: go version in go mod is correct now

Cali0707:
Fixes issues with `go mod tidy` resulting from an incorrect go mod file. See https://github.com/golang/go/issues/62278#issuecomment-1693538776 for more info

## Proposed Changes

- Move from `go 1.22` to explicit `go 1.22.3`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module knative.dev/eventing-kafka-broker
 
-go 1.22
+go 1.22.3
 
 require (
 	github.com/IBM/sarama v1.43.1


creydr:
Hey @Cali0707,
thanks for checking on this. Do you know, if this value will then be updated over time, or do we have to bump it manually?

Cali0707:
>  Do you know, if this value will then be updated over time, or do we have to bump it manually?

Not sure, but I imagine we will have to bump this ourselves sadly...

creydr:
> Not sure, but I imagine we will have to bump this ourselves sadly...

Ok. IIRC we had to do this in the past anyhow (for minor versions)

/lgtm

PR #3951, opened 2024-06-25
Move to using namespaced listers for eventtypes in data plane

Cali0707:
Currently, there are some error logs in the namespaced broker data plane stating that "the data-plane does not have sufficient permissions to list/watch eventtypes"

## Proposed Changes

- Use namespaced listers instead of a global lister

```release-note
:bug: The namespaced broker receiver only tries to list eventtypes in it's own namespace
```


creydr (commented):
/lgtm

PR #3954, opened 2024-06-27
Revert "fix: go version in go mod is correct now"

pierDipi:
Reverts knative-extensions/eventing-kafka-broker#3950

I think if everything and us start using 1.22 installations, we wouldn't need this


creydr (approved):
/lgtm

Issue #3970, opened 2024-07-04
Delivery format: add and provision `format` field to EgressConfig in contract


Cali0707:
**Problem**
To support dispatching events in the format specified in the `format` field being added into the delivery spec of components, we need to add a field into the contract, and then provision that field from the control plane.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
Unit tests indicating that the reconciler is setting the contract properly

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve? 1-2

**Additional context (optional)**
Add any other context about the feature request here.
- Parent issue: https://github.com/knative/eventing/issues/8057
- Required before this can be worked on: https://github.com/knative/eventing/issues/8073

PR #3980, opened 2024-07-09
feat: added delivery format to the contract

Cali0707:
Part of #3970 

## Proposed Changes

- Add the delivery `format` option to the data plane contract


Leo6Leo:
/lgtm
As the number 8 is unique and not being reused in `EgressConfig`

PR #3983, opened 2024-07-10
Fix depency update

Cali0707:
Fixes the issues in the dependency update PR from @knative-automation 

## Proposed Changes

- Fix the build issues due to the `trigger.Install` test interface changing
- Fix the permissions for the kafka controller so that it can start (and scale down the statefulsets)


matzew (approved):
/lgtm
/approve

PR #3994, opened 2024-07-18
fix: knative kafka extensions are short enough to meet the SHOULD req in spec

### LINKED ISSUE #3992: knativekafkapartition CE extension does not conform to spec
**Describe the bug**
The CE Extension `knativekafkapartition` was recently added, but it does not conform to the CloudEvent spec.

https://github.com/knative-extensions/eventing-kafka-broker/pull/3464

`CloudEvents attribute names MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9') from the ASCII character set. Attribute names SHOULD be descriptive and terse and SHOULD NOT exceed 20 characters in length.`

With the NodeJS library in particular, it throws an error because "knativekafkapartition" is 21 characters.

Could it be changed to "knkafkapartition" possibly? (and probably "knativekafkaoffset" to "knkafkaoffset" too)

 **Expected behavior**
Keep all CE extension names 20 characters or less.

 **To Reproduce**
Any event produced will have this header if I'm not mistaken.

 **Knative release version**
1.13

Cali0707:
Fixes #3992 

## Proposed Changes

- Shorten the extensions

**Release Note**

```release-note
BREAKING CHANGE: the knative kafka partition/offset ce extensions have been renamed to knkafkapartition and knkafkaoffset to keep them under 20 characters in length
```

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/CloudEventOverridesMutator.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/CloudEventOverridesMutator.java
@@ -46,7 +46,7 @@ public class CloudEventOverridesMutator implements CloudEventMutator {
     }
 
     private void applyKafkaMetadata(CloudEventBuilder builder, Number partition, Number offset) {
-        builder.withExtension("knativekafkapartition", partition);
-        builder.withExtension("knativekafkaoffset", offset);
+        builder.withExtension("knkafkapartition", partition);
+        builder.withExtension("knkafkaoffset", offset);
     }
 }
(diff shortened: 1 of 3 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/consumer/CloudEventOverridesMutator.java
```diff
@@ -46,7 +46,7 @@ private void applyCloudEventOverrides(CloudEventBuilder builder) {
     }
 
     private void applyKafkaMetadata(CloudEventBuilder builder, Number partition, Number offset) {
-        builder.withExtension("knativekafkapartition", partition);
-        builder.withExtension("knativekafkaoffset", offset);
+        builder.withExtension("knkafkapartition", partition);
```
matzew, line 49: @Cali0707 can you check w/ the Java SDK why it is not failing on 20+ chars?
(and maybe double check w/ golang)?

matzew (approved):
/lgtm
/approve

/hold
this seems breaking change. should we do something there? Or is release-notes good enought (I guess it is good enought). Just wondering

pierDipi:
/hold

I'm worried this will break users, given that it is a should is it really worth it? I think the JS SDK should not throw for this as it's a `SHOULD`

Cali0707:
> I'm worried this will break users, given that it is a should is it really worth it? I think the JS SDK should not throw for this as it's a `SHOULD`

Sure, let me try opening a PR to the SDK to fix that

Cali0707:
I opened https://github.com/cloudevents/sdk-javascript/pull/593

matzew:
@Cali0707 closing.

the fix in JS seems better addressing this `SHOULD`.

/close

PR #4001, opened 2024-07-23
fix: virtual secrets created from netspec have name and namespace

Cali0707:
In the clientpool code, we use the name and namespace of a secret along with the bootstrap servers to uniquely identify the connections being created. However, the virtual secrets created from the net spec did not have these fields set.

## Proposed Changes

- Create a name for the virtual secret by concatenating all secret names
- Create a namespace for the virtual secret by concatentatinf all secret namespaces

### DIFF
--- a/control-plane/pkg/security/secrets_provider_net_spec.go
+++ b/control-plane/pkg/security/secrets_provider_net_spec.go
@@ -19,8 +19,11 @@ package security
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	corelisters "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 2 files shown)


control-plane/pkg/security/secrets_provider_net_spec.go
```diff
@@ -53,21 +56,27 @@ func ResolveAuthContextFromNetSpec(lister corelisters.SecretLister, namespace st
...
+			},
+			Data: virtualSecretData,
+		},
 		MultiSecretReference: multiSecretReference,
 	}
 	return authContext, nil
 }
 
-func toContract(securityFields []*securityField) ([]*contract.SecretReference, map[string][]byte) {
+func toContract(securityFields []*securityField) ([]*contract.SecretReference, map[string][]byte, string, string) {
```
pierDipi, line 79: can't we just secret as return value (data+name+namespace) ? Too many return values is hard to read and easy to swap wrong values (like name in place of namespace)

pierDipi (approved):
/lgtm
/approve

PR #4005, opened 2024-07-24
fix: do not double increment clientpool wg

### LINKED ISSUE #3999: Controller in version 1.14 opens more and more connections to Kafka cluster
**Describe the bug**
With the 1.14 release of the kafka-broker it looks like it ships with a problem related to the connections handling in the kafka controller. 
Over the time the controller uses more and more connections to our kafka cluster. 
In our case it is a managed kafka cluster from Confluent.
We got aware of it as we had a subscription which wasn't forwarding any received events for the subscription/channel.
After investigating we figured out it was caused by a failing vertx http client (java.lang.IllegalStateException: Client is closed). 
The problem could be solved by restarting the channel dispatcher.
After further investigation we figure out the following cause of the failing client:

1. The client was closed due to the subscription wasn't ready and therefore removed from the config. https://github.com/knative-extensions/eventing-kafka-broker/issues/3136

2. The subscription reconcile failed as we where hitting the connection rate limit of our kafka cluster.

3. We where hitting the limit as the kafka controller is opening more and more connections to the kafka cluster. The connections spike every 10 minutes.

4. Restarting the controller resets the connections.

connection metrics from our kafka cluster
image

zoomed in on the 28.06 08:00-09:45
image

used container sockets of the controller (update till rollback)
image

image

We have done a rollback (01.07 ~12:00) to the latest 1.13 releases for now to stabilize our system.

 **Expected behavior**
The controller maintains a stable amount of connections as of the 1.13 release.

 **To Reproduce**
The increasing connection count should be visible in version 1.14.5 of the kafka-broker. The connection count should increase slowly. 
The problem occurs in both of our environments QA and Production.

 **Knative release version**
 kafka-broker 1.14.5
 eventing 1.14.2
 serving 1.14.1

 **Additional context**
All the kafka deployments have an istio sidecar injected.

The failing vertx client was a symptom of the failed reconciliation of one of our subscriptions. In addition it looked like the client was "stuck" as the removal of the config causes a non graceful shutdown/removal of the client. The client is reused after the channel got added to the config again.
Let us know if you would need more logs or metrics to investigate.

Cali0707:
Fixes #3999

Currently, when creating a cluster admin from a client, we increment the client wait group once, but this causes the group to be incremented twice as it was already incremented once when retrieving the client from the clientpool. This leads to the client not being able to be closed.

## Proposed Changes

- Increment the wait group once for client, once for clusteradmin rather than twice in the latter case.

### DIFF
--- a/control-plane/pkg/kafka/clientpool/clientpool.go
+++ b/control-plane/pkg/kafka/clientpool/clientpool.go
@@ -117,7 +117,7 @@ func (cp *ClientPool) getClient(ctx context.Context, bootstrapServers []string, 
 
 			logger.Debug("Closing client")
 
-			if err := value.client.Close(); !errors.Is(err, sarama.ErrClosedClient) {
+			if err := value.client.Close(); err != nil && !errors.Is(err, sarama.ErrClosedClient) {
 				logger.Errorw("Failed to close client", zap.Error(err))
 			}
 		}()
--- a/control-plane/pkg/kafka/clientpool/clientpool.go
+++ b/control-plane/pkg/kafka/clientpool/clientpool.go
@@ -66,6 +66,16 @@ type GetKafkaClientFunc func(ctx context.Context, bootstrapServers []string, sec
 type GetKafkaClusterAdminFunc func(ctx context.Context, bootstrapServers []string, secret *corev1.Secret) (sarama.ClusterAdmin, error)
 
 func (cp *ClientPool) GetClient(ctx context.Context, bootstrapServers []string, secret *corev1.Secret) (sarama.Client, error) {
+	client, err := cp.getClient(ctx, bootstrapServers, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	client.incrementCallers()
+	return client, nil
+}
+
+func (cp *ClientPool) getClient(ctx context.Context, bootstrapServers []string, secret *corev1.Secret) (*client, error) {
 	// (bootstrapServers, secret) uniquely identifies a sarama client config with the options we allow users to configure
 	key := makeClusterAdminKey(bootstrapServers, secret)
 
@@ -76,7 +86,6 @@ func (cp *ClientPool) GetClient(ctx context.Context, bootstrapServers []string, 
 	// if a corresponding connection already exists, lets use it
 	if val, ok := cp.cache.Get(key); ok && val.hasCorrectSecretVersion(secret) {
 		logger.Debug("successfully got a client from the clientpool")
-		val.incrementCallers()
 		return val, nil
 	}
 	logger.Debug("failed to get an existing client, going to create one")
@@ -114,16 +123,16 @@ func (cp *ClientPool) GetClient(ctx context.Context, bootstrapServers []string, 
 		}()
 	})
 
-	val.incrementCallers()
 	return val, nil
 }
 
 func (cp *ClientPool) GetClusterAdmin(ctx context.Context, bootstrapServers []string, secret *corev1.Secret) (sarama.ClusterAdmin, error) {
-	c, err := cp.GetClient(ctx, bootstrapServers, secret)
+	c, err := cp.getClient(ctx, bootstrapServers, secret)
 	if err != nil {
 		return nil, err
 	}
 
+	c.incrementCallers()
 	ca, err := clusterAdminFromClient(c, cp.newClusterAdminFromClient)
 	if err != nil {
 		c.Close()
--- a/control-plane/pkg/kafka/clientpool/clientpool.go
+++ b/control-plane/pkg/kafka/clientpool/clientpool.go
@@ -127,12 +127,11 @@ func (cp *ClientPool) getClient(ctx context.Context, bootstrapServers []string, 
 }
 
 func (cp *ClientPool) GetClusterAdmin(ctx context.Context, bootstrapServers []string, secret *corev1.Secret) (sarama.ClusterAdmin, error) {
-	c, err := cp.getClient(ctx, bootstrapServers, secret)
+	c, err := cp.GetClient(ctx, bootstrapServers, secret)
 	if err != nil {
 		return nil, err
 	}
 
-	c.incrementCallers()
 	ca, err := clusterAdminFromClient(c, cp.newClusterAdminFromClient)
 	if err != nil {
 		c.Close()
--- a/control-plane/pkg/kafka/clientpool/clientpool_test.go
+++ b/control-plane/pkg/kafka/clientpool/clientpool_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/IBM/sarama"
 	"github.com/stretchr/testify/assert"
+	"go.uber.org/atomic"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


pierDipi (commented):
Can we add some unit tests?

maybe with some package local method/function for ClientPool inspection only?

pierDipi:
Actually we don't even need other methods, we could just access the private fields

control-plane/pkg/kafka/clientpool/clientpool.go
```diff
@@ -114,16 +123,16 @@ func (cp *ClientPool) GetClient(ctx context.Context, bootstrapServers []string,
...
 }
 
 func (cp *ClientPool) GetClusterAdmin(ctx context.Context, bootstrapServers []string, secret *corev1.Secret) (sarama.ClusterAdmin, error) {
-	c, err := cp.GetClient(ctx, bootstrapServers, secret)
+	c, err := cp.getClient(ctx, bootstrapServers, secret)
 	if err != nil {
 		return nil, err
 	}
 
+	c.incrementCallers()
```
pierDipi, line 135: isn't this incremented in `getClient`

pierDipi: we could leave the usage of the public one to avoid duplication, so that the increment is done only in one place

Cali0707: Not anymore, I removed the calls from there.

Now we explicitly increment it in the admin or client public methods

control-plane/pkg/kafka/clientpool/clientpool_test.go
```diff
@@ -117,6 +118,47 @@ func TestGetClusterAdmin(t *testing.T) {
...
+	clusterAdmin, err := clients.GetClusterAdmin(ctx, []string{"localhost:9092"}, nil)
+	assert.NoError(t, err)
+
+	clusterAdmin.Close()
+	client1.Close()
+
+	time.Sleep(time.Second * 2)
+
+	client2, err := clients.GetClient(ctx, []string{"localhost:9092"}, nil)
+	assert.NoError(t, err)
```
pierDipi, line 153: why do we need this access?

control-plane/pkg/kafka/clientpool/clientpool_test.go
```diff
@@ -117,6 +118,47 @@ func TestGetClusterAdmin(t *testing.T) {
...
+	}
+
+	client1, err := clients.GetClient(ctx, []string{"localhost:9092"}, nil)
+	assert.NoError(t, err)
+
+	clusterAdmin, err := clients.GetClusterAdmin(ctx, []string{"localhost:9092"}, nil)
+	assert.NoError(t, err)
+
+	clusterAdmin.Close()
+	client1.Close()
```
pierDipi, line 148: Should the admin be also closed

With this, it's failing

```diff
diff --git a/control-plane/pkg/kafka/clientpool/clientpool_test.go b/control-plane/pkg/kafka/clientpool/clientpool_test.go
index 86c59908a..2ff1066ec 100644
--- a/control-plane/pkg/kafka/clientpool/clientpool_test.go
+++ b/control-plane/pkg/kafka/clientpool/clientpool_test.go
@@ -26,6 +26,7 @@ import (
        "go.uber.org/atomic"
        corev1 "k8s.io/api/core/v1"
        metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
        kafkatesting "knative.dev/eventing-kafka-broker/control-plane/pkg/kafka/testing"
        "knative.dev/eventing-kafka-broker/control-plane/pkg/prober"
 )
@@ -124,17 +125,23 @@ func TestClientCloses(t *testing.T) {
        ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
 
        cache := prober.NewLocalExpiringCache[clientKey, *client, struct{}](ctx, time.Second*1)
-       isClosed := atomic.NewBool(false)
+       clientClosed := atomic.NewBool(false)
+       adminClosed := atomic.NewBool(false)
 
        clients := &ClientPool{
                cache: cache,
                newSaramaClient: func(_ []string, _ *sarama.Config) (sarama.Client, error) {
                        return &kafkatesting.MockKafkaClient{OnClose: func() {
-                               isClosed.Toggle()
+                               clientClosed.Toggle()
                        }}, nil
                },
                newClusterAdminFromClient: func(_ sarama.Client) (sarama.ClusterAdmin, error) {
-                       return &kafkatesting.MockKafkaClusterAdmin{ExpectedTopics: []string{"topic1"}}, nil
+                       return &kafkatesting.MockKafkaClusterAdmin{
+                               ExpectedTopics: []string{"topic1"},
+                               OnClose: func() {
+                                       adminClosed.Toggle()
+                               },
+                       }, nil
                },
        }
 
@@ -149,11 +156,9 @@ func TestClientCloses(t *testing.T) {
 
        time.Sleep(time.Second * 2)
 
-       client2, err := clients.GetClient(ctx, []string{"localhost:9092"}, nil)
-       assert.NoError(t, err)
-
        // the client should have been closed successfully now
-       assert.True(t, isClosed.Load())
+       assert.True(t, clientClosed.Load())
+       assert.True(t, adminClosed.Load())
 
        client2.Close()
        cancel()
diff --git a/control-plane/pkg/kafka/testing/admin_mock.go b/control-plane/pkg/kafka/testing/admin_mock.go
index ec1c67287..0ab5cb5fd 100644
--- a/control-plane/pkg/kafka/testing/admin_mock.go
+++ b/control-plane/pkg/kafka/testing/admin_mock.go
@@ -44,6 +44,7 @@ type MockKafkaClusterAdmin struct {
        ErrorOnDeleteTopic error
 
        ExpectedClose      bool
+       OnClose            func()
        ExpectedCloseError error
 
        // DescribeTopics
@@ -301,5 +302,8 @@ func (m *MockKafkaClusterAdmin) RemoveMemberFromConsumerGroup(groupId string, gr
 
 func (m *MockKafkaClusterAdmin) Close() error {
        m.ExpectedClose = true
+       if m.OnClose != nil {
+               m.OnClose()
+       }
        return m.ExpectedCloseError
 }
```

control-plane/pkg/kafka/clientpool/clientpool.go
pierDipi: I mean this 

```
	c, err := cp.getClient(ctx, bootstrapServers, secret)
	if err != nil {
		return nil, err
	}

	c.incrementCallers()
```

can be replaced with:

```
	c, err := cp.GetClient(ctx, bootstrapServers, secret)
	if err != nil {
		return nil, err
	}

```

control-plane/pkg/kafka/clientpool/clientpool_test.go
Cali0707: Ah no my bad, I was thinking of a different cache implementation that only removes the old entries when you access a new one.

control-plane/pkg/kafka/clientpool/clientpool_test.go
Cali0707: Oh that looks like a problem in our cluster admin proxy, let me fix that

control-plane/pkg/kafka/clientpool/clientpool.go
Cali0707: oh I see what you mean, let me change that

PR #4018, opened 2024-07-29
Remove channelv2 post install

Cali0707:
This should fix the broken upgrade tests we are seeing

## Proposed Changes

- Remove the post install job
- Remove the downgrade cleanup in the tests
- Remove the extra permissions for the post install job


Cali0707:
/cc @pierDipi 

This should fix all the upgrade tests

Issue #3901, opened 2024-05-15
kafka-broker-receiver crashes on startup starting with 1.14.0 in EKS


wSedlacek:
**Describe the bug**
The container in `kafka-broker-receiver` reports error code 143 on startup.
The only logs are
```
Picked up JAVA_TOOL_OPTIONS: -XX:+CrashOnOutOfMemoryError
{"@timestamp":"2024-05-15T19:22:46.914Z","@version":"1","message":"Registering tracing configurations backend=UNKNOWN sampleRate=0.0 loggingDebugEnabled=false headersFormat=W3C","logger_name":"dev.knative.eventing.kafka.broker.core.tracing.TracingConfig","thread_name":"main","level":"INFO","level_value":20000,"backend":"UNKNOWN","sampleRate":0.0,"loggingDebugEnabled":false,"headersFormat":"W3C"}
{"@timestamp":"2024-05-15T19:22:46.957Z","@version":"1","message":"Starting Receiver env=ReceiverEnv{ingressPort=8080, livenessProbePath='/healthz', readinessProbePath='/readyz', httpServerConfigFilePath='/etc/config/config-kafka-broker-httpserver.properties'} BaseEnv{producerConfigFilePath='/etc/config/config-kafka-broker-producer.properties', dataPlaneConfigFilePath='/etc/brokers-triggers/data', metricsPort=9090, metricsPath='/metrics', metricsPublishQuantiles=false}","logger_name":"dev.knative.eventing.kafka.broker.receiver.main.Main","thread_name":"main","level":"INFO","level_value":20000,"env":{"producerConfigFilePath":"/etc/config/config-kafka-broker-producer.properties","dataPlaneConfigFilePath":"/etc/brokers-triggers/data","metricsPort":9090,"metricsPath":"/metrics","metricsJvmEnabled":false,"metricsHTTPClientEnabled":false,"metricsHTTPServerEnabled":false,"configTracingPath":"/etc/tracing","configFeaturesPath":"/etc/features","waitStartupSeconds":8,"ingressPort":8080,"ingressTLSPort":8443,"livenessProbePath":"/healthz","readinessProbePath":"/readyz","httpServerConfigFilePath":"/etc/config/config-kafka-broker-httpserver.properties","publishQuantilesEnabled":false}}
[...]

PR #4021, opened 2024-07-30
fix: do not build OIDC config unless enabled

Cali0707:
Temporary fix for #3901 (this does not fix the underlying problem as far as I can tell, but will enable anyone not interested in OIDC to use 1.14+)

## Proposed Changes

- Do not load the OIDC config unless it is enabled
- Refactor the receiver `Main` class to allow for redeploying the receiver verticles when the OIDC config changes
- Set an annotation on the receiver pods to force k8s to update the files mounted from the `config-features` configmap, so that the receiver quickly picks up on changes to feature flags.

**Release Note**

```release-note
The receiver components now successfully start on EKS. Special shoutout to @treyhyde for help debugging this.
```


Cali0707:
TODO (as a follow up): refactor the `FileWatcher` class to be able to either watch a single file or a directory. Because of how configmaps are mounted as files into pods this PR just watches the specific file associated with the OIDC feature flag. But, going forwards we will probably end up wanting to support more feature flags, so we should make the `FileWatcher` class more generic there

Cali0707:
/cc @pierDipi 

This should be good to go now, I updated it to not redeploy the verticles (like we had discussed offline).

Now, we are using a listener + callback pattern for the OIDC Discovery stuff

Cali0707:
/cc @pierDipi @creydr @Leo6Leo 

I have fixed the merge conflicts here

Cali0707:
/hold

Looks like the tests are failing for some reason...

creydr:
`/home/prow/go/src/knative.dev/eventing-kafka-broker is out of date. Please run hack/update-codegen.sh`

@Cali0707 can you run it?

data-plane/core/src/main/java/dev/knative/eventing/kafka/broker/core/oidc/OIDCDiscoveryConfigListener.java
```diff
@@ -0,0 +1,107 @@
...
+        return oidcDiscoveryConfig;
+    }
+
+    public int registerCallback(Consumer<OIDCDiscoveryConfig> callback) {
+        if (this.callbacks == null) {
+            this.callbacks = new ArrayList<>();
+        }
+
+        this.callbacks.add(callback);
+        return this.callbacks.size();
```
creydr, line 74: IIUC you are using this return value as some "id", which you later pass to "deregisterCallback()". This uses `.set(id, null)` to remove it.
So you do the following on an initial empty arraylist
1. add a callback
2. return as "callbackId" `1`
3. later want to deregister the callbackId `1`, but actually you would need to call `deregisterCallback(0)`

sry, but I didn't find a way to explain it more complicated :facepalm: 

tl;dr; is, that IIUC, you should return here `this.callbacks.size() - 1`, if you want to use this as the "callbackId"

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/ReceiverVerticle.java
```diff
@@ -135,8 +136,8 @@ public void start(final Promise<Void> startPromise) {
                 .watchIngress(IngressReconcilerListener.all(this.ingressProducerStore, this.ingressRequestHandler))
                 .buildAndListen(vertx);
 
-        TokenVerifier tokenVerifier = new TokenVerifierImpl(vertx, oidcDiscoveryConfig);
-        this.authenticationHandler = new AuthenticationHandler(tokenVerifier);
+        this.buildAuthHandler(oidcDiscoveryConfigListener.getOidcDiscoveryConfig());
+        this.oidcDiscoveryCallbackId = this.oidcDiscoveryConfigListener.registerCallback(this::buildAuthHandler);
```
creydr, line 140: Maybe we can add a quick note somewhere, that the FileWatcher runs the callbacks, when it gets started (in it's .run() ), that way we can be "sure", that `this.authenticationHandler` is not null (which is used in line 243)

creydr (commented):
Thanks for working on this :muscle: . Did a first round. But I probably need to run it locally too :D

Cali0707:
/cc @creydr 

Thanks for taking a first look!

Cali0707:
/unhold

Seems like things are passing now, thanks for finding that off by one error @creydr !!

Cali0707:
/test upgrade-tests

Last failure looks like a flake, cert-manager failed to come ready

PR #4029, opened 2024-08-06
FileWatcherTest consistently pass

pierDipi:
Filesystem notifications might be triggered more times than expected so check for greater or equal to.


creydr (commented):
Thanks for fixing

/lgtm

PR #4046, opened 2024-08-08
Remove finalizer when trigger is deleted

pierDipi:
Add a no-op `FinalizeKind` to configure knative/pkg base reconciler to remove the finalizer when a trigger is deleted.


creydr (approved):
/lgtm

PR #4055, opened 2024-08-12
[release-1.15] fix: do not build OIDC config unless enabled (#4021)

Cali0707:
This is a cherry pick of #4021 so that the OIDC config is only built when needed


creydr (commented):
/lgtm

PR #4056, opened 2024-08-12
[release-1.14] fix: do not build OIDC config unless enabled (#4021)

Cali0707:
Manual cherry pick of #4021 so that we only rebuild the OIDC Discovery config when needed


creydr (approved):
/lgtm

Issue #4038, opened 2024-08-07
[control-plane] List applying policies in KafkaChannels `.status.policies`


creydr:
As the [Eventing Authorization](https://docs.google.com/document/d/1u6vyMviGpD13uyLKSGV7_D_p0ZRBPWg6PlLItRH8UFI/edit?resourcekey=0-gCE8brkIopDqc4jE7c9_tQ#heading=h.9zw0wrg4io53) feature track describes, target resources of `EventPolicies`, will reflect in their status, which `EventPolicy` is applied to them. 

For example:

```yaml
apiVersion: kafkachannels.messaging.knative.dev/v1
kind: KafkaChannel
metadata:
  name: my-channel
spec:
  ...
status:
  ...
  policies:
    - name: event-policy
      apiVersion: v1alpha1
    - name: another-event-policy
      apiVersion: v1alpha1

  conditions:
    - type: Ready
      status: "True"
    - type: EventPoliciesReady
      status: "True"
```

As the KafkaChannel should support `EventPolicies`, wee need to do the following:

* Update the KafkaChannel reconciler to show applying policies in the KafkaChannels `.status.policies` and set the KafkaChannels `EventPoliciesReady` condition. Also watch EventPolicies for changes.
* Set `EventPoliciesReady` condition to true, in case no policy applies, and set `message` according to the `default-authorization-mode` (e.g. `Default authorization mode is 'allow-all'`). Check the [Default Authorization Mode section](https://docs.google.com/document/d/1u6vyMviGpD13uyLKSGV7_D_p0ZRBPWg6PlLItRH8UFI/edit?resourcekey=0-gCE8brkIopDqc4jE7c9_tQ#heading=h.c6984bgldrz) in the feature track for additional information.

**Hint:**
* Similar to https://github.com/knative/eventing/issues/7977
* Requires #4069 

**Additional hints for new contributors before starting with this issue:**

1. When the issue has the `Draft` status, the issue is subject to change and thus should not be started to be worked on
2. Make sure you've read and understood the [CONTRIBUTING.md](https://github.com/knative-extensions/eventing-kafka-broker/blob/main/CONTRIBUTING.md) guidelines
[...]

PR #4064, opened 2024-08-16
Provision contract fields for authorization

### LINKED ISSUE #4041: [control-plane] Provision allowed Subjects in the contract
The in #4040 added fields need to be provisioned with the correct values, so that the data-plane has this information too.

**Additional hints for new contributors before starting with this issue:**

1. When the issue has the `Draft` status, the issue is subject to change and thus should not be started to be worked on
2. Make sure you've read and understood the [CONTRIBUTING.md](https://github.com/knative-extensions/eventing-kafka-broker/blob/main/CONTRIBUTING.md) guidelines
3. Make sure you're able to [run Knative Eventing locally](https://github.com/knative/eventing/blob/main/DEVELOPMENT.md#create-a-cluster-and-a-repo) with [Knative Eventing Kafka](https://github.com/knative-extensions/eventing-kafka-broker/blob/main/DEVELOPMENT.md) and [run at least the unit tests](https://github.com/knative-extensions/eventing-kafka-broker/blob/main/DEVELOPMENT.md#run-unit-tests). 
4. Feel free to raise any questions you have either directly here in the issue, in the [#knative-eventing Slack channel](https://cloud-native.slack.com/archives/C04LMU33V1S) or join the [Eventing Workgroup Meeting / Office Hours](https://github.com/knative/community/blob/main/working-groups/WORKING-GROUPS.md#eventing)
5. When you feel comfortable with this issue, feel free to assign it to you (e.g. by commenting `/assign`). Please be aware that we might unassign you, if we don't see any progress from your side to give other contributors also a chance to work on this issue.

creydr:
Fixes #4041

## Proposed Changes

- provision applying EventPolicies from Broker, KafkaChannel and KafkaSink in their contracts
 
 
Hint: currently only the Broker exposes the applying policies in its status. KafkaChannel and KafkaSink are open/WIP #4038 and #4039)


control-plane/pkg/reconciler/broker/broker.go
```diff
@@ -668,6 +662,12 @@ func (r *Reconciler) reconcilerBrokerResource(ctx context.Context, topic string,
 	}
 	resource.EgressConfig = egressConfig
 
+	eventPolicies, err := coreconfig.EventPoliciesFromAppliedEventPoliciesStatus(broker.Status.AppliedEventPoliciesStatus, r.EventPolicyLister, broker.Namespace, features)
```
pierDipi, line 665: it's weird we're relying on the status, so we require a precise ordering for operations, if this method `reconcilerBrokerResource` depends on having the applied event policies, I'd make it explicit with a parameter

pierDipi: Same for other reconcilers

creydr: I see what you mean. The problem is, that we don't have access at this point otherwise to the applying eventpolicies. `auth.UpdateStatusWithEventPolicies()` from eventing-core sets the applying (and ready) policies in the status and marks the conditions.

So I see two alternatives here:
1. use the `auth.GetEventPoliciesForResource()` to get the applying policies, filter on the ready ones and pass them to `reconcilerBrokerResource()` (similar to passing the audience). This has the drawback, that we need to load the eventpolicies anyhow later again (in `auth.UpdateStatusWithEventPolicies()`) as we need to mark the conditions. And this would kind of duplicate the logic from `auth.UpdateStatusWithEventPolicies()` to "filter" which policies apply for a resource
2. Refactor the `auth.UpdateStatusWithEventPolicies()` to take a list of event policies which apply for a resource. Anyhow, this would mean, that we need to duplicate the `auth.GetEventPoliciesForResource()` in each reconciler (in eventing-core too).

Objections or other ways?

creydr: @pierDipi regarding 2: something like https://github.com/knative/eventing/pull/8173. Anyhow this would still mean, we need to duplicate the logic, which event policies apply at the end to a resource (only the ready ones) :/

pierDipi: The difference is that the caller knows and it's orchestrating the order of operations and the expected order can be even be more explicit with a temporary variable

```golang
	appliedEventPolicy := ...
	err = auth.UpdateStatusWithEventPolicies(features, &appliedEventPolicy, &broker.Status, r.EventPolicyLister, eventing.SchemeGroupVersion.WithKind("Broker"), broker.ObjectMeta)
	if err != nil {
		return fmt.Errorf("could not update broker status with EventPolicies: %v", err)
	}

	// Get resource configuration.
	brokerResource, err := r.reconcilerBrokerResource(ctx, topic, broker, secret, topicConfig, audience, appliedEventPolicy)
	if err != nil {
		return statusConditionManager.FailedToResolveConfig(err)
	}

       broker.Status.AppliedEventPoliciesStatus = appliedEventPolicy

Comment (pierDipi) [reply control-plane/pkg/reconciler/broker/broker.go:665]: Also I've noticed this

the function `UpdateStatusWithEventPolicies` resents the status to empty so if there is error on, for example `GetEventPoliciesForResource` we will have no policies applied but I would expect the system to remain configured the the latest known "good" states until it recovers

```
func UpdateStatusWithEventPolicies(featureFlags feature.Flags, status *eventingduckv1.AppliedEventPoliciesStatus, statusMarker EventPolicyStatusMarker, eventPolicyLister listerseventingv1alpha1.EventPolicyLister, gvk schema.GroupVersionKind, objectMeta metav1.ObjectMeta) error {
	status.Policies = nil
//... 
```

creydr: Ah ok. Now I understand better what you mean.

pierDipi (approved):
/lgtm
/approve

PR #4065, opened 2024-08-16
[WIP] Provision contract fields for authz (contract update afterwards)

### LINKED ISSUE #4041: [control-plane] Provision allowed Subjects in the contract

creydr:
Fixes #4041

## Proposed Changes

- provision applying EventPolicies from Broker, KafkaChannel and KafkaSink in their contracts

Same as #4064, except that it refactors the Broker, KafkaChannel and KafkaSink reconcilers to do the contract update at the end of the reconcilation. This leads to having the .status fields which are part of the contract (e.g. OIDC audience or applying EventPolicies) are being updated directly in the contract.
Anyhow, the probing con only be done after the contract update, as otherwise the data-plane is not aware of the resource and this would lead to 404 on probing the resources endpoint. This has the side effect, that the resource will be marked as addressable, before the probe run (anyhow we have the `ProbeSucceeded` condition, which can render the resource as NotReady anyhow)


Cali0707:
> Anyhow, the probing con only be done after the contract update, as otherwise the data-plane is not aware of the resource and this would lead to 404 on probing the resources endpoint. This has the side effect, that the resource will be marked as addressable, before the probe run (anyhow we have the ProbeSucceeded condition, which can render the resource as NotReady anyhow)

@creydr as far as I can tell from the code, we don't use the `.status.address(es)` field when doing the contract update. Maybe we can compute the addresses, but not set them on the resource until after the probing is successful and instead just keep them in variables somewhere? I might also be missing something...

creydr:
We use the audience and later the event policies from the status. So this "safes" us one reconciliation. But I might miss here something too...

creydr:
Closing as discussed with @pierDipi, as sources can start to send events to addressables as soon as they have an address (they don't check the ready condition), so we should set the addressable condition, only when the addressable is ready to accept events.

PR #4070, opened 2024-08-19
Make EventPolicy type in contract more generic

creydr:
Adding matchers for Exact and Prefix, so we can adjust it later to not only match on the "sub" claim.


pierDipi (approved):
/lgtm
/approve

creydr:
Rebased. @pierDipi could you recheck?

PR #4095, opened 2024-09-11
Paired upgrade tests

### LINKED ISSUE #4047: Create paired pre-upgrade and post-upgrade tests which create resources before upgrade and verify them after upgrade
This is currently not the case because the pre-upgrade tests create its own set of resources and verify them, and post-upgrade tests create yet another set of resources and verify them. The resources from pre-upgrades do NOT live through the upgrades.

**Problem**

The only set of tests which create resources before upgrade and verify them later are the "continual" upgrade tests. The problem is that the test suite performs "downgrades" right after upgrades, so potential problems like https://github.com/knative-extensions/eventing-kafka-broker/issues/4034 are not caught because the problem is "fixed" again after downgrades.

The new upgrade tests should be added to the existing ones.
They should follow this scenario:
* Create resources before upgrade
* Perform upgrade
* Verify the existing resources after upgrade
* Delete the resources (either directly or by deleting the test namespace)
* Create a new set of resources
* Perform downgrade
* Verify the existing resources after downgrade
* Delete the resources (either directly or by deleting the test namespace)

Developers/Testers

**Exit Criteria**
Upgrade/downgrade tests working.

**Time Estimate (optional):** 
2-3

mgencur:
Fixes #4047

## Proposed Changes

- The upgrade tests now test every feature in 4 different scenarios:
  - Run the same test post-upgrade and post-downgrade. Create resources and verify them at the same phase.
  - Create resources pre-upgrade and verified/removed them post-upgrade.
  - Create resources pre-upgrade, verify post-upgrade, verify and remove post-downgrade.
  - Create resources post-upgrade, verify and remove post-downgrade.
- Re-use REKT-based tests for upgrade testing
- Sink-specific tests were removed because they're now covered by the KafkaSinkSourceBinaryEventFeature and KafkaSinkSourceStructuredEventFeature. This feature uses both KafkaSource and KafkaSink
- KafkaChannel tests now run only with binary encoding as it uses ContainerSource that is based on the heartbeats image. Previously they were running with both but now it's re-using a test that uses just binary encoding. I don't think testing binary vs. structured encoding in upgrade tests is really necessary.
- Speed up test execution of the whole upgrade test suite. There were a couple of successful runs with this new approach that took 34m24s, 33m1s. Looking at the old approach and successful runs, there were: 38m50s, 40m9s, 38m52s, 43m47s, 42m1s. The new approach is about 6-7 minutes faster on average than the previous approach.
  

**Release Note**

```release-note

```

**Docs**


mgencur:
The upgrade tests show a single failure now: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-extensions_eventing-kafka-broker/4095/upgrade-tests_eventing-kafka-broker_main/1834210401064062976

Trying again to see if this is reproducible.

/test upgrade-tests

pierDipi:
```
E0912 12:47:54.384853   42406 library.go:117] Package knative.dev/eventing/test/upgrade does not have module info. Non go modules projects are no longer supported. For feedback, refer to https://github.com/google/go-licenses/issues/128.
```

mgencur:
This PR requires changes from https://github.com/knative/eventing/pull/8190 . When it's merged I will update dependencies and this failure should go away.
I'm going to provide issue description tomorrow - with details about what's changed. The CI runtime looks better than before, and test coverage is increased!

mgencur:
The `reconciler-tests-namespaced-broker_eventing-kafka-broker_main` fails with same failures as on other PRs. Not related to my PR.

pierDipi (approved):
/lgtm
/approve

PR #4104, opened 2024-09-18
Use v2 specific label to avoid matching previous deployment-based dispatchers

pierDipi:
This ensures that our `consumergroup` controller only match v2-specific pods and won't try to reconcile pods that belong to the previous deployment.

Seen here https://github.com/knative-extensions/eventing-kafka-broker/issues/4091 with
```
failed to schedule consumers: failed to get data plane volume kafka-resources
      in pod knative-eventing/kafka-broker-dispatcher-6b6855d6c5-7nhtw
```
so the consumergroup controller is trying to reconcile a pod that is not of the statefulset but for the "old" deployment

### DIFF
--- a/control-plane/pkg/reconciler/consumergroup/consumergroup.go
+++ b/control-plane/pkg/reconciler/consumergroup/consumergroup.go
@@ -766,7 +766,10 @@ func (r *Reconciler) reconcileSecret(ctx context.Context, expectedSecret *corev1
 }
 
 func (r *Reconciler) ensureContractConfigmapsExist(ctx context.Context, scheduler Scheduler) error {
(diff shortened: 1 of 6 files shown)


pierDipi:
```
executor.go:189: Sep 18 14:46:04.915 install_head [ERR] The StatefulSet "kafka-source-dispatcher" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

Change to labels to STS is not possible

matzew (approved):
/lgtm
/approve

matzew:
Good call, adding this, @pierDipi :+1:

PR #4120, opened 2024-09-25
Set the `name` value for metrics tags to the correct top-level resource

pierDipi:
Previously to v2, we were setting `name` to the Broker or Channel name of the resource.


matzew (commented):
/lgtm
/approve

creydr (commented):
Thanks

/lgtm

PR #4122, opened 2024-09-30
Remove unused editor config Java plugin

pierDipi:
We use spotless to reformat Java code and editor config is not used anymore.


creydr (approved):
/lgtm

PR #4137, opened 2024-10-15
Describe sacura resources when they don't become ready

pierDipi:
Helps with debugging


matzew (approved):
/lgtm
/approve

/retest

PR #4138, opened 2024-10-15
Use and test KafkaSource v1 API

pierDipi:
Replace usages of v1beta1 KafkaSource to v1 KafkaSource

Part of https://github.com/knative-extensions/eventing-kafka-broker/issues/3953


pierDipi:
This can be merged only after 1.16 is out

matzew (commented):
/lgtm
/approve

PR #4155, opened 2024-11-01
Allow kafka-controller to list JobSinks

mgencur:
This is required otherwise KafkaBroker and KafkaSource can't forward events to JobSink. This type of error is thrown on Kafka Broker status:

```
failed to reconcile contract: failed to reconcile egress: failed to resolve subscriber: failed to get lister for sinks.knative.dev/v1alpha1, Resource=jobsinks: jobsinks.sinks.knative.dev is forbidden: User "system:serviceaccount:knative-eventing:kafka-controller" cannot list resource "jobsinks" in API group "sinks.knative.dev" at the cluster scope
```

I have also created a test for KafkaBroker forwarding events to JobSink to verify the fix. But it's probably not necessary to include it. Anyway, it's available here: https://github.com/mgencur/eventing-kafka-broker/commit/57dbb533751c825762d94ca51b10d66e1d8a5439

Fixes #

## Proposed Changes

-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/config/eventing-kafka-broker/200-controller/200-controller-cluster-role.yaml
+++ b/control-plane/config/eventing-kafka-broker/200-controller/200-controller-cluster-role.yaml
@@ -224,6 +224,16 @@ rules:
     verbs:
       - update
 
+  - apiGroups:
+      - "sinks.knative.dev"
+    resources:
+      - "jobsinks"
+      - "jobsinks/status"
+    verbs:
+      - get
+      - list
+      - watch
+
   # resources needed to grant eventtype autocreate rbac to namespaced data plane component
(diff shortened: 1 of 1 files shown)


pierDipi:
@mgencur I think the best fix would be to add jobsink to the addressable resolver aggregation in the core resources: https://github.com/knative/eventing/blob/main/config/core/roles/addressable-resolvers-clusterrole.yaml, that has the benefit that all controllers will get access to it for resolving the address

pierDipi (approved):
/lgtm
/approve

I'll merge this one anyway, but I'd add the permissions there so that other controllers work too

PR #4159, opened 2024-11-06
Thread safe improvements on OIDCDiscoveryConfigListener

matzew:
## Proposed Changes

- using `ConcurrentHashMap`, `AtomicRef` and `AtomicInt` for improving thread safeness

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/auth/OIDCDiscoveryConfigListener.java
+++ b/data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/auth/OIDCDiscoveryConfigListener.java
@@ -20,10 +20,11 @@ import dev.knative.eventing.kafka.broker.core.file.FileWatcher;
 import io.vertx.core.Vertx;
 import java.io.File;
 import java.io.IOException;
-import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/auth/OIDCDiscoveryConfigListener.java
```diff
@@ -62,14 +62,10 @@ public OIDCDiscoveryConfigListener(String featuresConfigPath, Vertx vertx, int t
     }
 
     public OIDCDiscoveryConfig getOidcDiscoveryConfig() {
-        return oidcDiscoveryConfig;
+        return oidcDiscoveryConfig.get();
     }
 
     public int registerCallback(Consumer<OIDCDiscoveryConfig> callback) {
```
pierDipi, line 68:
```suggestion
    public synchronized int registerCallback(Consumer<OIDCDiscoveryConfig> callback) {
```

since executing the 2 `add + size() - 1` can lead to return the wrong index if one thread is executing `add` before `size() - 1` is executed

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/auth/OIDCDiscoveryConfigListener.java
```diff
@@ -36,24 +35,25 @@ public class OIDCDiscoveryConfigListener implements AutoCloseable {
...
-                    if (this.oidcDiscoveryConfig == null) {
+                    if (this.oidcDiscoveryConfig.get() == null) {
                         this.buildFeaturesAndOIDCDiscoveryConfig();
-                        if (this.oidcDiscoveryConfig != null && this.callbacks != null) {
-                            this.callbacks.stream()
-                                    .filter(Objects::nonNull)
-                                    .forEach(c -> c.accept(this.oidcDiscoveryConfig));
+                        OIDCDiscoveryConfig config = this.oidcDiscoveryConfig.get();
+                        if (config != null) {
+                            this.callbacks.forEach(callback -> callback.accept(config));
```
pierDipi, line 56: I see in `deregisterCallback` we set `callback` (the element) to null, don't we need to check that `callback` is not null before calling `accept` ?

data-plane/receiver/src/main/java/dev/knative/eventing/kafka/broker/receiver/impl/auth/OIDCDiscoveryConfigListener.java
```diff
@@ -62,14 +62,10 @@ public OIDCDiscoveryConfigListener(String featuresConfigPath, Vertx vertx, int t
...
-        return oidcDiscoveryConfig;
+        return oidcDiscoveryConfig.get();
     }
 
-    public int registerCallback(Consumer<OIDCDiscoveryConfig> callback) {
-        if (this.callbacks == null) {
-            this.callbacks = new ArrayList<>();
-        }
-
+    public synchronized int registerCallback(Consumer<OIDCDiscoveryConfig> callback) {
```
pierDipi, line 68: One other issue I see with index based registration is that with addition and removal of callbacks the array is ever growing which can lead to OOM, it's better to have a map `id -> string`

matzew: you mean like `ConcurrentMap` ? instead of the `CopyOnWriteArrayList`?

pierDipi: yes

pierDipi (approved):
/lgtm
/approve

PR #4164, opened 2024-11-14
add configurable CE metadata extension prefix

cchayka-campsci:
## Proposed Changes

- allow the CloudEvent extension attributes "partition" and "offset" to have a customizable prefix

As brought up before in this issue https://github.com/knative-extensions/eventing-kafka-broker/issues/3992, a previous version of NodeJS `cloudevents` would throw an error for any extension > 20 chars in length.

We are in a tough position to update this package for a large number of legacy services, so we were hoping you could consider allowing this change.

Basically just adding a new env var "CE_METADATA_EXTENSION_PREFIX" to change the "knativekafka" part of the "knativekafkapartition" and "knativekafkaoffset" CE extensions.

**Release Note**

```release-note
:broom: allow customization of the prefix of the CE extensions
```

**Docs**

_previous pull request to the wrong branch:_ https://github.com/knative-extensions/eventing-kafka-broker/pull/4144

### DIFF
--- a/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/DispatcherEnv.java
+++ b/data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/DispatcherEnv.java
@@ -16,6 +16,7 @@
 package dev.knative.eventing.kafka.broker.dispatcher.main;
 
 import static java.util.Objects.requireNonNull;
+import static java.util.Objects.requireNonNullElse;
 
 import dev.knative.eventing.kafka.broker.core.utils.BaseEnv;
 import java.util.function.Function;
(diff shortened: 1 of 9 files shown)


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/main/DispatcherEnv.java
```diff
@@ -31,12 +32,17 @@ public class DispatcherEnv extends BaseEnv {
     public static final String EGRESSES_INITIAL_CAPACITY = "EGRESSES_INITIAL_CAPACITY";
     private final int egressesInitialCapacity;
 
+    public static final String CE_METADATA_EXTENSION_PREFIX = "CE_METADATA_EXTENSION_PREFIX";
```
matzew, line 35: was there any feedback from the node.js library on why they fail - even the spec says `should` ? 

CC @Cali0707 

Asking here, instead of adding all this code for pluggable prefixes

cchayka-campsci: The node.js library was updated successfully. We are in a tough position to update some of our legacy services so were hoping this change could be considered regardless.

Cali0707: Yeah, we fixed it in the SDK since as you said @matzew  this was a "should" requirement. But, I guess that doesn't fix it if people are using a non-fixed version of the SDK

Cali0707: Circling back here, I don't particularly mind having the extra code for the configuration if it improves interop with other systems. Any opinions @matzew @pierDipi ?

PR #4203, opened 2025-01-08
Upgrade Kafka clients to 3.9.0

pierDipi:
From 3.7.1 to 3.9.0


data-plane/pom.xml
```diff
@@ -57,7 +57,7 @@
     <junit.jupiter.version>5.10.1</junit.jupiter.version>
     <mokito.junit.jupiter.version>5.12.0</mokito.junit.jupiter.version>
     <fabric8.kubernetes.version>6.10.0</fabric8.kubernetes.version>
-    <kafka.version>3.7.1</kafka.version>
+    <kafka.version>3.9.0</kafka.version>
     <debezium.version>2.6.1.Final</debezium.version>
```
matzew, line 61: FWIW, the `vertx` team uses still 3.7 with a slightly newer version of DBZ:
https://github.com/vert-x3/vertx-kafka-client/blob/master/pom.xml#L37-L40

Not sure if @ppatierno and team will eventually update the vertx-kafka-client to 3.9 and/of if than DBZ 2.7.x will work w/ Kafka ?! 


I see however, that DBZ itself is having Kafka 3.9 on its 3.0 branch:
https://github.com/debezium/debezium/blob/3.0/pom.xml

matzew:
/hold
let's see how this goes.

In addition, we could also get to latest Strimzi 0.45 and its Kafka 3.9.0 operand

PR #4207, opened 2025-01-10
feat: add updatescale method to generated clients for consumergroups

Cali0707:
Similar to https://github.com/knative-extensions/eventing-kafka-broker/pull/3976 but for the consumergroup client. This is useful since we currently do not expose the scale subresource for the triggers, so this will enable scaling the consumergroups directly

## Proposed Changes

- Add UpdateScale method to consumergroup generated client

```release-note
the consumergroup client now supports UpdateScale methods
```


pierDipi (approved):
/lgtm
/approve

PR #4216, opened 2025-01-16
Update logback and logstash dependencies

pierDipi:
As per title


pierDipi:
/hold

I'll double check if logging is still visible as we had issues in the past

pierDipi:
Logs are emitted

/unhold

creydr (approved):
/lgtm

PR #4245, opened 2025-02-04
Align w/ Quarkus 3.15 and friends

matzew:
Try with 3.15.2 - since 3.15.3 has a netty bug

Fixes #

## Proposed Changes

- as per title
-
-

**Release Note**

```release-note

```

**Docs**


data-plane/receiver/pom.xml
```diff
@@ -35,6 +35,11 @@
       <version>${project.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-registry-prometheus-simpleclient</artifactId>
```
matzew, line 40: I think we should have inherited this from `core` module... will check

data-plane/core/pom.xml
```diff
@@ -181,7 +181,7 @@
 
     <dependency>
       <groupId>io.micrometer</groupId>
-      <artifactId>micrometer-registry-prometheus</artifactId>
+      <artifactId>micrometer-registry-prometheus-simpleclient</artifactId>
```
matzew, line 184: We use the simpleclient, since that is _backwards_ compatible to the old APIs.

While `vertx-micrometer` 4.5.10 is _still_ on the `1.12` APIs, we do inherit the `1.13` APIs via the Quarkus BOM... 
And hence we really need the simpleclient, we are using that - to avoid nasty breakages

matzew:

the usual source of failures...

pierDipi (approved):
/lgtm
/approve

PR #4266, opened 2025-02-18
client: Allow configurable sarama conf.Version

### LINKED ISSUE #4265: Support configurable Sarama config kafka version
**Problem**
Allow compatibility with kafka brokers that support older versions of the kafka protocol

system operator

**Exit Criteria**
Allow sarama to support kafka broker with protocol version that is not the hard-coded default version

**Time Estimate (optional):** 
0.2 day

Allow to inject environment variable SARAMA_CONFIG_VERSION set to a configured version that sarama will use against a kafka broker

sagigrimberg:
Some kafka servers may not support different kafka protocol versions. Hence, expose sarama kafka protocol Version config knob to allow wider kafka server support.

Fixes #4265


control-plane/pkg/kafka/client.go
```diff
@@ -56,10 +58,24 @@ type NewClientFunc func(addrs []string, config *sarama.Config) (sarama.Client, e
...
+	if exists {
+		for _, v := range sarama.SupportedVersions {
+			if v.String() == version {
+				saramaVersion = v
+			}
+		}
+	}
+	log.Printf("sarama version set to %s", saramaVersion.String())
+	return saramaVersion
+}
```
pierDipi, line 73: Can we call this in the `init` function above so that we do it once at startup and never again?

```golang
var configVersion sarama.KafkaVersion = sarama.DefaultVersion

func init() {
	// Disable Sarama metrics
	metrics.UseNilMetrics = true

	configVersion = getSaramaConfigVersion()
}

// ...

func GetSaramaConfig(configOptions ...ConfigOption) (*sarama.Config, error) {
	config.Version = configVersion
}
```

sagigrimberg: @pierDipi who/where this init() is called?

sagigrimberg: Ah, I see init() exists already - cool, will change...

sagigrimberg: fixed @pierDipi

pierDipi (commented):
/ok-to-test

pierDipi (approved):
/lgtm
/approve

PR #4292, opened 2025-03-10
fix: repeat args from sub-func call

alingse:
## Proposed Changes

I am verifying some lint idea in my project see https://github.com/alingse/sundrylint/pull/10
it report the function call with repeat args from a sub-function call.

we think the sub-function call is heavy, so the repeat args probably a copy-and-paste mistake.

I check the top 4000 github repos, and found this repo has this case.

**Release Note**

```release-note

```

**Docs**


Cali0707:
@alingse could you clarify what the issue you are solving is here? This doesn't seem to be a problem

Cali0707 (approved):
/ok-to-test
/lgtm
/approve

matzew:
Isn't always the same test failing? 

Do we _really_ need this?!?

matzew:
/close

feel free to argue.. but I see no direct value in this, and its constantly failing

PR #4363, opened 2025-05-07
[WIP] Check if OIDC SA is set, before requesting a token for it

creydr:
We have seen a `java.lang.IllegalArgumentException: Name must be provided` in the dispatcher, when the target had an audience set, but the OIDC service account was not reconciled in time.

## Proposed Changes

- :bug: Only request an OIDC token, when OIDC SA is set


data-plane/dispatcher/src/main/java/dev/knative/eventing/kafka/broker/dispatcher/impl/http/WebClientCloudEventSender.java
```diff
@@ -193,7 +196,14 @@ private Future<?> send(final CloudEvent event, final Promise<HttpResponse<Buffer
         if (this.targetOIDCAudience.isEmpty()) {
             requestToken = Future.succeededFuture(null);
         } else {
-            requestToken = this.tokenProvider.getToken(this.oidcServiceAccount, this.targetOIDCAudience);
+            if (this.oidcServiceAccount.name() == null
+                    || this.oidcServiceAccount.name().isEmpty()) {
+                requestToken = Future.failedFuture(
+                        new OIDCTokenRequestException(
```
matzew, line 202: /lgtm

PR #4365, opened 2025-05-08
:broom: Propagate the CaCerts and Audience of the ConsumerGroups Subscribers and DLS in the Triggers status too

creydr:
As per title, just to be consistent and propose this information in the triggers status too as we do for other resources.


control-plane/pkg/reconciler/trigger/v2/triggerv2.go
```diff
@@ -313,9 +313,13 @@ func propagateConsumerGroupStatus(cg *internalscg.ConsumerGroup, trigger *eventi
 		}
 	}
 	trigger.Status.SubscriberURI = cg.Status.SubscriberURI
+	trigger.Status.SubscriberCACerts = cg.Status.SubscriberCACerts
```
matzew, line 316: :heart:

matzew (approved):
/lgtm
/approve

PR #4373, opened 2025-05-14
remove config-tracing ConfigMap

tmvfb:
Fixes https://github.com/knative-extensions/eventing-kafka-broker/issues/4256 and https://github.com/knative-extensions/eventing-kafka-broker/issues/3012

When deploying [eventing-kafka-broker](https://github.com/knative-extensions/eventing-kafka-broker) together with knative-eventing with kustomize, the latter fails to resolve between 2 differing ConfigMaps with differing `_example:` field. This is also an issue when running GitOps with argocd, if deploying knative-eventing and kafka-broker as 2 different apps.

## Proposed Changes

- Delete the ConfigMap, as it is a duplicate of upstream.


metacoma:
Hey @tmvfb, is there any reason to have the `config-tracing` ConfigMap in this repo and include it in the release manifests?

tmvfb:
> Hey @tmvfb, is there any reason to have the `config-tracing` ConfigMap in this repo and include it in the release manifests?

I don't see any reason for that. I took a look at the codebase, there is also "config-features" from upstream reused somewhere, but the manifest is not in this repo. Would probably make sense to delete `config-tracing` for good.

tmvfb:
Both failing tests are related to the webhook service.
1. `reconciler-tests-namespaced-broker_eventing-kafka-broker_main` fails trying to delete a "ConsumerGroup" resource containing a secret that no longer exists (it gets deleted a few steps before). It also has a specific finalizer:
    ```yaml
                "finalizers": [
                  "foregroundDeletion"
                ],
    ```
    I don't think this has anything to do with this PR.
2. Another test seems to execute a function that does not make sense to me (there is no ConfigMap with "kafka-source-dispatcher" name) https://github.com/tmvfb/eventing-kafka-broker/blob/main/test/rekt/features/kafka_source.go#L113-L113. Anyways, related to the same webhook.

I guess, I need some assistance here, not sure if anything needs to be done.

metacoma:
> I guess, I need some assistance here, not sure if anything needs to be done.

I think these "flaky" tests are failing not because of your changes — they’re also failing in other PRs.

matzew:
/retest-required
let's see ;)

PR #4381, opened 2025-05-16
Applying as 'rules: null', since that is how api server is treating i…

matzew:
Instead of empty array (`[]`) we are now applying it as `rules: null`, since that is how _api server_ is treating it.

If applied via operator and the api-server treats it differently _will+ cause the operator to update back to its own `rules: []` version, instead of `null`, as treated by the api-server...

So, you end up with some back and forth, and have a lot of:

```
"msg":"Merging","diff":"{\"rules\":[]}"}
"msg":"Updating","name":"/knative-kafka-controller-post-install","type":"rbac.authorization.k8s.io/v1, Kind=ClusterRole"}
```

## Proposed Changes

- change empty rules field to `null` 

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/control-plane/config/post-install/200-controller-cluster-role.yaml
+++ b/control-plane/config/post-install/200-controller-cluster-role.yaml
@@ -20,4 +20,4 @@ metadata:
   name: knative-kafka-controller-post-install
   labels:
     app.kubernetes.io/version: devel
-rules: []
+rules: null


control-plane/config/post-install/200-controller-cluster-role.yaml
```diff
@@ -20,4 +20,4 @@ metadata:
   name: knative-kafka-controller-post-install
   labels:
     app.kubernetes.io/version: devel
-rules: []
+rules: null
```
creydr, line 23: Wondering, if we can remove the `rules` completely and that way fall back to the api servers defaults :thinking:

creydr (approved):
/lgtm

PR #4401, opened 2025-06-04
Create cluster-issuer before creating kafka certificate for tests

creydr:
In test/kafka_setup.sh, we're [creating a Certificate](https://github.com/knative-extensions/eventing-kafka-broker/blob/f4c76b517f6a41b18a8df6e42651ab1e58de0382/test/kafka/kafka_setup.sh#L30) for Kafka with the cluster-issuer `knative-eventing-ca-issuer` referenced. This requires that the cluster-issuer was created before, otherwise the certificate won't become ready.
Although we create the cluster-issuer when we setup knative-eventing for the tests (via ./hack/run.sh), we should make sure we have the cluster-issuer setup correctly before we apply the certificate in case knative-eventing is not installed via ./hack/run.sh (e.g. like we do in eventing-istio).


creydr:
```
    tracker.go:139: Waiting for eventing-e2e1 to be deleted
--- FAIL: TestSacuraSinkSourceJob (1343.67s
```
/test integration-tests

PR #4413, opened 2025-06-17
Allow the kafka-controller to list EventTransform API

matzew:
Similar like https://github.com/knative-extensions/eventing-kafka-broker/pull/4155

but for `Eventtranform` API

### DIFF
--- a/control-plane/config/eventing-kafka-broker/200-controller/200-controller-cluster-role.yaml
+++ b/control-plane/config/eventing-kafka-broker/200-controller/200-controller-cluster-role.yaml
@@ -225,6 +225,15 @@ rules:
       - update
 
   - apiGroups:
+      - "eventing.knative.dev"
+    resources:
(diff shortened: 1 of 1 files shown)


maschmid:
Are we sure we're doing this the right way?  

both the jobsink and eventtransforms should be addressables  ( `duck.knative.dev/addressable: "true"` ) 

so, instead of adding them explicitly here, the kafka controller should be granted these permissions by `kafka-controller-addressable-resolver` (which it does).

so, both jobsinks and eventtransforms need to be declared in `addressable-resolver` ClusterRole  and have `duck.knative.dev/addressable: "true"` 

In current eventing main they seem to be there https://github.com/knative/eventing/blob/main/config/core/roles/addressable-resolvers-clusterrole.yaml   , so if it doesn't work downstream, it's probably because they're not there in that specific downstream eventing's addressable-resolvers clusterrole ...

PR #4444, opened 2025-07-08
Fix dependency update

Cali0707:
Fixes the errors seen in #4441 

## Proposed Changes

- Update to latest dependencies
- Copy one metrics tag from knative/pkg that was dropped in the OpenCensus -> Otel migration, until we migrate this repo too


evankanderson (commented):
/lgtm
/approve

PR #4454, opened 2025-07-22
Update license files

dsimansk:
Fixes #

## Proposed Changes

- Update license files

**Release Note**

```release-note
Update license files
```

This PR updates GRPC version of java library to latest one and regenerated `.pg.go`. The other pom deps are fixed due to changes inherited from newer lib. 

Let's see how tests will behave.

Update: trying minimal change first to pass release job. 

/cc @gauron99 @Cali0707


Cali0707:
@dsimansk you may need to also change the protoc version here: https://github.com/knative/infra/blob/9947f1951c1071bee7e767bd71c51085f12dc912/images/prow-tests/Dockerfile#L25

But, this will break all the other PRs...

dsimansk:
> @dsimansk you may need to also change the protoc version here: https://github.com/knative/infra/blob/9947f1951c1071bee7e767bd71c51085f12dc912/images/prow-tests/Dockerfile#L25
> 
> But, this will break all the other PRs...

Yep, I was afraid of it. I've noticed the version in prow test image. OK, let me find the minimal change that could pass the release bar. I'll try to match my protoc toolchain to current infra.

PR #4457, opened 2025-07-23
Cleanup invalid symlink

dsimansk:
Fixes #

## Proposed Changes

- Cleanup invalid symlink

Follow-up of original CM removal in https://github.com/knative-extensions/eventing-kafka-broker/pull/4373

/cc @gauron99


gauron99:
nice!
/lgtm
/approve

PR #4465, opened 2025-07-31
Fix Keda Trigger SSL & SASL test to install broker only once

creydr:
Currently the `TriggerSSLScalesToZeroWithKeda` and `TriggerSASLScalesToZeroWithKeda` tests try to install the broker two times. This PR cleans it up and installs it only once.

Also enables to run all tests containing "Keda" in their name from the e2e_new folder as part of the keda-rekt-test suite


Cali0707 (approved):
/lgtm
/approve

PR #4493, opened 2025-08-19
feat: migrate data plane to OTel

Cali0707:
This PR moves the EKB data plane to OTel. We missed this in the initial migration.

## Proposed Changes

- Mount config-observability to the data plane
- Use a ObservabilityConfig class to parse and manage the mounted values
- Use ObservabilityConfig to configure metrics + traces for otel

**Release Note**

```release-note
Breaking Change: The EKB data plane now reports metrics and traces with OTel
```


dprotaso:
> 20:29:53.058 [main] ERROR dev.knative.eventing.kafka.broker.core.reconciler.impl.ResourcesReconcilerImpl - Failed to reconcile [onNewEgress] egress  id=1 consumerGroup=hello destination=
java.lang.RuntimeException: null
	at dev.knative.eventing.kafka.broker.core.reconciler.impl.ResourcesReconcilerImplTest.reconcileFailedOnNewEgress(ResourcesReconcilerImplTest.java:261)

data-plane/config/broker/500-dispatcher.yaml
```diff
@@ -134,18 +130,21 @@ spec:
...
           readinessProbe:
             failureThreshold: 3
-            httpGet:
-              port: 9090
-              path: /metrics
-              scheme: HTTP
+            exec:
+              command:
+                - cat
+                - /tmp/healthy
```
Cali0707, line 147: @matzew @creydr I had to switch the liveness/readiness probes because we can no longer be sure that prometheus is running in the data plane.

This means for the dispatcher there is no running http server for the probes to go to, hence the "exec" probes I have here

dprotaso:
/retest

Github actions flaked out and profiling didn't run

dprotaso:
ok maybe it wasn't a github action failure but when one profiling action fails it fast fails the other ones

Cali0707:
The profiling GH actions pretty much always fail has been my experience -  we should probably revisit them in the future.

/cc @matzew @creydr @aliok 
Tests seem to be passing, this should be good to review now

data-plane/config/broker/500-dispatcher.yaml
creydr: This is something I don't understand. Why can't we be sure the metrics are exposed?

Cali0707: The issue is that the user can choose to use either prometheus or otlp for the metrics. If they use otlp, this is a "push" protocol so there is no web server that starts and listens on any port

creydr (commented):
/lgtm

PR #4558, opened 2025-10-10
KEDA auth using externally managed TriggerAuthentication resource

qswinson:
Follow up for #4516 [comment](https://github.com/knative-extensions/eventing-kafka-broker/pull/4516#discussion_r2364547984)

## Proposed Changes
New Feature
Add optional KEDA autoscaler configuration to specify externally managed TriggerAuthentication or ClusterTriggerAuthentication resource. These are specified through the `config-kafka-autoscaler` ConfigMap or annotations on the scaled resource.

The annotation names:
- `autoscaling.eventing.knative.dev/authentication-name`: The name of the KEDA resource.
- `autoscaling.eventing.knative.dev/authentication-kind`: TriggerAuthentication or ClusterTriggerAuthentication. Only required if the KEDA resource is ClusterTriggerAuthentication.
- `autoscaling.eventing.knative.dev/aws-region`: Required if using AWS IAM authentication for MSK. The KEDA Kafka source requires this property to be defined in the `trigger.metadata`. KEDA does not recognize this property in the authentication resource as it does with properties like `sasl`.

The primary use case for this feature is to support KEDA autoscaling when using AWS IAM authentication for MSK. This allows a different role to be used for the KEDA operator with a different set of permissions from the KNative Eventing services. By allowing the user to supply an externally managed resource, the complexity of configuring this KEDA authentication is removed from the KNative code and gives more flexibility to the user. 

**Release Note**

```release-note
Add optional autoscaler configurations to specify an externally managed KEDA TriggerAuthentication or ClusterTriggerAuthentication resource to be used by the KNative managed KEDA ScaledObject resource.
```

**Docs**

WIP

### DIFF
--- a/control-plane/pkg/autoscaler/config.go
+++ b/control-plane/pkg/autoscaler/config.go
@@ -23,8 +23,9 @@ import (
 
 // AutoscalerConfig contains the defaults defined in the autoscaler ConfigMap.
 type AutoscalerConfig struct {
-	AutoscalerClass    map[string]string
-	AutoscalerDefaults map[string]int32
+	AutoscalerClass          map[string]string
+	AutoscalerDefaults       map[string]int32
+	AutoscalerAuthentication map[string]string
 }
 
 func defaultConfig() *AutoscalerConfig {
@@ -38,6 +39,11 @@ func defaultConfig() *AutoscalerConfig {
 			AutoscalingPollingIntervalAnnotation: DefaultPollingInterval,
 			AutoscalingCooldownPeriodAnnotation:  DefaultCooldownPeriod,
 			AutoscalingLagThreshold:              DefaultLagThreshold,
+		},
+		AutoscalerAuthentication: map[string]string{
+			AutoscalingAuthenticationName: DefaultAuthenticationName,
+			AutoscalingAuthenticationKind: DefaultAuthenticationKind,
+			AutoscalingAwsRegion:          DefaultAwsRegion,
 		},
 	}
 }
@@ -53,9 +59,12 @@ func NewConfigFromMap(data map[string]string) (*AutoscalerConfig, error) {
 	for key, val := range data {
 		key = convertConfigKeyToAutoscalingAnnotation(key)
 
-		if key == AutoscalingClassAnnotation {
+		switch key {
+		case AutoscalingClassAnnotation:
 			ac.AutoscalerClass[AutoscalingClassAnnotation] = val
-		} else {
+		case AutoscalingAuthenticationName, AutoscalingAuthenticationKind, AutoscalingAwsRegion:
+			ac.AutoscalerAuthentication[key] = val
+		default:
 			i, err := strconv.ParseInt(val, 10, 32)
 			if err != nil {
 				return nil, fmt.Errorf("Expected value for autoscaling annotation: "+key+" should be integer but got "+val, err)
--- a/control-plane/pkg/autoscaler/config_test.go
+++ b/control-plane/pkg/autoscaler/config_test.go
@@ -1,0 +1,386 @@
+/*
+ * Copyright 2025 The Knative Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package autoscaler
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestDefaultConfig(t *testing.T) {
+	config := defaultConfig()
+
+	// Test AutoscalerClass defaults
+	expectedClass := map[string]string{
+		AutoscalingClassAnnotation: AutoscalingClassDisabledAnnotationValue,
+	}
+	if !reflect.DeepEqual(config.AutoscalerClass, expectedClass) {
+		t.Errorf("Expected AutoscalerClass %v, got %v", expectedClass, config.AutoscalerClass)
+	}
+
+	// Test AutoscalerDefaults
+	expectedDefaults := map[string]int32{
+		AutoscalingMinScaleAnnotation:        DefaultMinReplicaCount,
+		AutoscalingMaxScaleAnnotation:        DefaultMaxReplicaCount,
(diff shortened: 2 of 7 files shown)


matzew:
CC @twoGiants

Cali0707:
> The primary use case for this feature is to support KEDA autoscaling when using AWS IAM authentication for MSK. This allows a different role to be used for the KEDA operator with a different set of permissions from the KNative Eventing services. By allowing the user to supply an externally managed resource, the complexity of configuring this KEDA authentication is removed from the KNative code and gives more flexibility to the user.

@qswinson could you clarify this some more? Currently, Knative is managing the rest of the KEDA integration for these resources, so I'm a little hesitant to have this one edge case for the AWS bits around the trigger auth, especially since this would then need to be managed per-trigger by the user 

Also re: different set of permissions - doesn't Eventing still need permissions to through AWS IAM to authenticate to these clusters anyway? 

I could be missing something here, so please correct me anywhere you see fit.

Also cc @matzew @twoGiants for their thoughts

control-plane/pkg/autoscaler/config_test.go
twoGiants: I see you added quite a few tests here to cover the new logic you introduced in `config.go`. Thank you for taking care of code coverage :smile_cat: :+1:. Keep adding tests whenever you add new logic.

Now lets dive a bit deeper. Unfortunately the test code does not have the code quality I'd expect. But we will get there, no worries. I can support you, just let me know if something is not clear.

I have a few questions about the tests you wrote:
1. Why did you created a this test file here? There was none before because the config is tested implicitly in a different place.
2. Do you know where config can be tested implicitly or how to figure that out?
3. When you write unit tests for a class or a function always cover those things: conditions (if), loops (for, while), calculations/operations and polymorphism. Now you added a few new default values and extended the existing condition. It would be enough to write a test for each new condition case and one test for the default values. Why didn't you do that?

control-plane/pkg/autoscaler/config_test.go
```diff
@@ -0,0 +1,386 @@
...
+
+package autoscaler
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestDefaultConfig(t *testing.T) {
+	config := defaultConfig()
```
twoGiants: I would avoid package level tests when you can. The rule of thumb is: prefer public api tests over package scoped tests. Add package level tests only if its really needed. Here it isn't.

control-plane/pkg/autoscaler/config.go
```diff
@@ -39,6 +40,11 @@ func defaultConfig() *AutoscalerConfig {
 			AutoscalingCooldownPeriodAnnotation:  DefaultCooldownPeriod,
 			AutoscalingLagThreshold:              DefaultLagThreshold,
 		},
+		AutoscalerAuthentication: map[string]string{
+			AutoscalingAuthenticationName: DefaultAuthenticationName,
+			AutoscalingAuthenticationKind: DefaultAuthenticationKind,
+			AutoscalingAwsRegion:          DefaultAwsRegion,
```
twoGiants: As mentioned above. I would not do this. Probably better a generic way to add trigger metadata key-value pairs.

control-plane/pkg/autoscaler/keda/keda.go
```diff
@@ -80,6 +95,10 @@ func GenerateScaleTriggers(cg *kafkainternals.ConsumerGroup, triggerAuthenticati
 			"allowIdleConsumers": allowIdleConsumers,
 		}
 
+		if *awsRegion != "" {
+			triggerMetadata["awsRegion"] = *awsRegion
```
twoGiants: Same here. Don't think hard coding `awsRegion` is a good idea. A more generic way is better.

control-plane/pkg/autoscaler/keda/resources.go
```diff
@@ -85,3 +85,14 @@ func GetInt32ValueFromMap(dict map[string]string, key string, defaultValue int32
 	i32 := int32(i)
 	return &i32, nil
 }
+
+func GetStringValueFromMap(dict map[string]string, key string, defaultValue string) (*string, error) {
```
twoGiants: This returns either the `defaultValue` or an empty string or nil. Why empty string or nil? Does it really need the `defaultValue` if it's an empty string? 

Can you elaborate? Thx! :+1:

twoGiants (changes requested):
Hi @qswinson!

Great to see you back and thanks again for your new contribution. Keep those contributions coming! :1st_place_medal: 

I have noticed quite big additions in test code and added my comments to it. It would be great if you could address them.

And I think hard coding the `awsRegion` is not the way to go. I would opt for a generic way of adding trigger metadata key-value pairs. Then its the users responsibility to use `awsRegion` for the key and `us-west-1` for the value. Not sure how though :laughing:. Maybe something like that `autoscaling.eventing.knative.dev/trigger-metadata-awsRegion: "us-west-1"`. This part `autoscaling.eventing.knative.dev/trigger-metadata-` is the one that doesn't change and after that the generic one.

And I have one last little request to you. After you finish your re-work I would kindly ask you to fully cleane up the code, proofread it one last time and address all PR review comments before you request the next review. That would be really helpful! Thank you :smiley:  :+1:!

Let me know if I can support in any way. You can also find me on the CNCF slack. Happy coding!

qswinson:
> @qswinson could you clarify this some more? Currently, Knative is managing the rest of the KEDA integration for these resources, so I'm a little hesitant to have this one edge case for the AWS bits around the trigger auth, especially since this would then need to be managed per-trigger by the user

This separates the authentication concerns in a platform agnostic manner. Right now KEDA authentication for Kafka must be the same as KNative authentication for Kafka. That makes it difficult to track what api calls are coming from which service regardless of authentication type or Kafka hosting provider. By giving the end user the option to use a separate authentication mechanism for Keda versus KNative, end users who need to track that kind of usage have the tools to do so.
I do think this is a generic solution and not an edge case for AWS. If the ability to define an externally defined TriggerAuthentication is not included, then this code will need to be updated to support creating TriggerAuthentication resources for AWS MSK.

> 
> Also re: different set of permissions - doesn't Eventing still need permissions to through AWS IAM to authenticate to these clusters anyway?

The permission issue comes down to IAM role permission policies and trust relationship policies. The role used by the Keda service account must be granted permission to assume every role that is referenced in a TriggerAuthentication resource and every role referenced in a TriggerAuthentication resource must have a trust relationship policy that grants the Keda service account permission to assume it. 

If KNative Kafka Controller creates the TriggerAuthentication using the IAM role configured for the broker, then that role has to include the keda service account in its trust relationship. The Keda role must include broker IAM role in its permission policy. This then gives the Keda pod permission to do all IAM permissions granted to the Knative Kafka broker.

The diagram below illustrates how defining an external TriggerAuthentication or ClusterTriggerAuthentication separates the tight coupling of the KNative broker role and the Keda role.
Knative and Keda Auth - Single Account

PR #4571, opened 2025-11-05
Update dependencies

creydr:
Update dependencies and fix unit tests


creydr:
```
Waiting until all pods in namespace cert-manager are up......................................................................................................................................................
```
/test channel-reconciler-tests-ssl

creydr:
We probably need #4572 and the fix in 1.20 for the upgrade tests

Issue #4545, opened 2025-09-29
Installation of Knative Kafka Broker fails.


twoGiants:
**Describe the bug**

Following any of the two docs on how to install a Kafka broker in:
- [Docs -> Eventing -> Concepts -> Brokers -> Broker types -> Apache Kafka -> Installation](https://knative.dev/docs/eventing/brokers/broker-types/kafka-broker/#installation)
- [Docs -> Administration -> Install Knative with YAML -> Install Knative Eventing -> Install Eventing with YAML](https://knative.dev/docs/install/yaml-install/eventing/install-eventing-with-yaml/#optional-install-a-broker-layer)

When executing this cli command:

```sh
kubectl apply --filename https://github.com/knative-extensions/eventing-kafka-broker/releases/download/knative-v1.19.7/eventing-kafka-broker.yaml
```

We get this error:

```sh
The StatefulSet "kafka-broker-dispatcher" is invalid: spec.template.spec.containers[0].volumeMounts[1].name: Not found: "contract-resources" 
```

 **Expected behavior**

Following any of the two docs above I expect being able to deploy a Kafka Broker data plane.

 **To Reproduce**
1. Create Kind cluster.
1. Install Knative Eventing:
   ```sh
   kubectl apply -f https://github.com/knative/eventing/releases/download/knative-v1.19.4/eventing-crds.yaml
   kubectl apply -f https://github.com/knative/eventing/releases/download/knative-v1.19.4/eventing-core.yaml
   ```
1. Install Kafka controller:
   ```sh
   kubectl apply --filename https://github.com/knative-extensions/eventing-kafka-broker/releases/download/knative-v1.19.7/eventing-kafka-controller.yaml
   ```
3. Install Kafka broker:
   ```sh
   kubectl apply --filename https://github.com/knative-extensions/eventing-kafka-broker/releases/download/knative-v1.19.7/eventing-kafka-broker.yaml
   ```

 **Knative release version**

 - Platform (GKE, IKS, AKS, etc.): 
   - `kind v0.30.0 go1.25.0 linux/amd64`
   - `k8 1.34.0`
 - Knative Version:
   - Eventing 1.19.4

 **Additional context**
Add any other context about the problem here such as proposed priority

PR #4572, opened 2025-11-06
Fix race condition in dispatcher pod creation when webhook is not ready

creydr:
When dispatcher pods are created before the kafka-webhook-eventing deployment is ready, the pod defaulting webhook cannot inject the contract-resources ConfigMap volume reference, leaving the volume as emptyDir. This causes ConfigMapNameFromPod to fail when trying to schedule consumers.

This commit adds a fallback in ConfigMapNameFromPod: if the contract-resources volume exists but doesn't have a ConfigMap reference yet, use the pod name as the ConfigMap name. This matches the webhook's behavior and allows the controller to proceed with creating the ConfigMap.

Related to #4545


creydr:
/hold
need to investigate in the upgrade failures again

creydr:
Superseded by #4576

Issue #4421, opened 2025-06-19
Fix ST1016 linter staticcheck issues


dsimansk:
For details check on the parent issue #4417

Cali0707:
Fixing this issue involves re-enabling the ST1016 check in https://github.com/knative-extensions/eventing-kafka-broker/blob/main/.golangci.yaml, and fixing all associated linter errors for the check.

For more info on ST1016, read here: https://staticcheck.dev/docs/checks#ST1016

/triage accepted
/kind good-first-issue

jijo-OO7:
Hi @Cali0707 how do I regenerate `zz_generated.deepcopy.go`? The file has ST1016 errors and I can't find the `//go:generate directive`. What command regenerates deepcopy files in this.

creydr:
> Hi [@Cali0707](https://github.com/Cali0707) how do I regenerate `zz_generated.deepcopy.go`? The file has ST1016 errors and I can't find the `//go:generate directive`. What command regenerates deepcopy files in this.

Hi @jijo-OO7,
you can run `./hack/update-codegen.sh`. This should update all generated files

jijo-OO7:
Hi @creydr Thanks ! I'll run `./hack/update-codegen.sh` and update the PR with the changes. I'll follow up once it's done.

jijo-OO7:
@creydr  I ran `./hack/update-codegen.sh` but encountered a nil pointer dereference error in the code generation tool. The script crashed before completing. 

I've checked `git status` and no files were modified. Could you help me troubleshoot this? Is there an alternative way to regenerate the deepcopy files, or are there specific environment setup requirements for the script?

Thanks!

creydr:
> I ran `./hack/update-codegen.sh` but encountered a nil pointer dereference error in the code generation tool. The script crashed before completing.

This shouldn't happen. Can you show the logs and/or create a PR with your current changes so we can try to reproduce it?

jijo-OO7:
> > I ran `./hack/update-codegen.sh` but encountered a nil pointer dereference error in the code generation tool. The script crashed before completing.
> 
> This shouldn't happen. Can you show the logs and/or create a PR with your current changes so we can try to reproduce it?

## Update 
on  ``./hack/update-codegen.sh`` Issue
I've identified the root cause: the code generation tool had an incompatibility with Windows that was triggering the nil pointer dereference error. This incompatibility prevented the script from completing successfully.
[...]

PR #4599, opened 2025-11-25
Fix/enable st1016 linter

jijo-OO7:
# ST1016: Inconsistent Receiver Names Fixes:[#4421]

## Issue
Methods on the same type must use the same receiver name throughout the codebase.

## Changes
* Standardize receiver names across all methods on the same type
* Rename inconsistent receivers to follow convention (e.g., `ks` for `KafkaSource`)
* Update all related references

## Testing
`golangci-lint run ./...` passes without ST1016 errors


jijo-OO7:
Hi @Leo6Leo  Could you review when you have a moment and add `/ok-to-test`? Please mention if any changes are needed before merge. Thanks!

Cali0707:
Hey @jijo-OO7 - thanks for the PR

> Hi @Leo6Leo Could you review when you have a moment and add /ok-to-test? Please mention if any changes are needed before merge. Thanks!

We are working to review, may take a few days as there are other priorities we have as well. You shouldn't need to ping us individually right away if we are already assigned to the PR by prow, unless we haven't reviewed in a few days

On a different note, can you fix the failing "verify deps and codegen" and "code style" checks?

jijo-OO7:
Hey @Cali0707 thank you for the update. I’ll take care of the failing verify deps and codegen and code style checks. I appreciate your time and guidance.

jijo-OO7:
This PR is ready and waiting for ``/ok-to-test`` so integration tests can run.
Thanks!

data-plane/contract/src/main/java/dev/knative/eventing/kafka/broker/contract/DataPlaneContract.java
creydr: I don't think this file should change on this PR 🤔

.golangci.yaml
```diff
@@ -18,8 +18,6 @@ linters:
     staticcheck:
       checks:
         - all
-        - "-ST1003"
```
creydr: This line was removed in #4603 too which just merged. So you need to rebase

jijo-OO7:
the `reconciler-tests_eventing_kafka-broker_main` tests are failing consistently. Looking at the logs, I see this error:

```
admission webhook "validation.webhook.eventing.knative.dev" denied the request: 
validation failed: oidc-authentication feature not enabled
```

This is happening in both `TestBrokerSupportsAuthZ` and `TestKafkaSinkSupportsAuthZ`. 

From what I can tell, the tests are trying to create EventPolicy resources but the webhook is rejecting them because OIDC isn't enabled. Then the test times out waiting 120 seconds for something that was never created in the first place.

**My understanding:** This doesn't look like a resource issue - the error is pretty clear that OIDC needs to be enabled for these specific tests to work.

**My questions:**
1. Is OIDC supposed to be enabled in the test environment? 
2. Did something change recently that disabled it?
3. Should we be skipping these AuthZ tests if OIDC isn't available, or do we need to fix the test setup to enable it?

What's the right way to handle this?

jijo-OO7:
Hi @creydr I ran `/test all` and everything passes except `reconciler-tests_eventing_kafka-broker_main`. 
The AuthZ tests consistently fail with "oidc-authentication feature not enabled" even when 
run alongside all other tests. 

This suggests it's not a test ordering issue, but rather something specific to how the 
kafka-broker AuthZ tests set up OIDC. Did something change in the test setup or in how 
EventPolicy validation works?

jijo-OO7:
The failure is unrelated to the ST1016 changes. It comes from broker reconciler tests that create `EventPolicy` resources without enabling the required `oidc-authentication` feature flag, which causes admission to reject the resource and the tests to time out.

I’ve opened a separate issue to track this test behavior and propose aligning the test context with the required feature flags: <https://github.com/knative-extensions/eventing-kafka-broker/issues/4630#issue-3829897323>.

creydr (approved):

creydr:
Finally :tada: Thanks a lot @jijo-OO7 for your patience!

jijo-OO7:
@creydr Thanks for the quick turnaround on the fix!

PR #4608, opened 2025-12-05
Fix metrics

Cali0707:
Fixes #4566 

This also resolves an issue discussed in slack at https://cloud-native.slack.com/archives/C04LMU33V1S/p1764165572333779 where metrics stopped showing up with prometheus configured

## Proposed Changes

- Handle OtlpConfig keys correctly (based on work done in #4607 )
- Verify that OtlpConfig actually works
- Handle Prometheus config defaults 
- Use correct prometheus classes with Vertx (the others were silently failing)
- Verify that prometheus config defaults were handled correctly

**Release Note**

```release-note
fix: prometheus and otlp metrics export when configured
```

PR #4609, opened 2025-12-05
[release-1.19] Fix metrics (#4608)

Cali0707:
This is a cherry-pick of #4608 to fix metrics on the 1.19 release


Cali0707:
/hold

Looks like the manual cherry-pick did not actually pull in all of the change 😢

Cali0707:
/unhold

This should be correct now

PR #4618, opened 2025-12-24
feat: add trigger sink failures metric

### LINKED ISSUE #465: Expose Trigger's sink failures count metric
**Problem**

A metric to easily see that a Trigger isn't working, because the trigger's sink is failing (e.g. it always returns a 4xx/5xx or times out, DNS lookup fails, etc.).

Event Consumer

**Exit Criteria**

Expose Trigger's sink failures count metric.

**Time Estimate (optional):** 
1

We can expose the following metric:

- `sink_failures_total{id=<trigger_uid>, url=<subscriber_url>} <num_failures>`

  - The `id` label allows correlating this metric with Kafka consumer metrics like:
    `kafka_consumer_request_rate{client_id="consumer-<trigger_uid>-1",kafka_version="2.6.0",} <value>`
    `kafka_consumer_io_wait_ratio{client_id="consumer-<trigger_uid>-1",kafka_version="2.6.0",} <value>`
    `...`
  - The `url` label allows correlating failures of a specific sink since we can have 2 triggers with the same sink URL.

Sapthagiri777:
Fixes knative-extensions/eventing-kafka-broker#465

## Proposed Changes

- Add `TRIGGER_SINK_FAILURES` counter metric to `Metrics.java`
- Increment the counter in `WebClientCloudEventSender` when all retries are exhausted or on non-retryable failures
- Preserve existing retry behaviour while exposing total sink failures per trigger

**Release Note**

**Docs**

No user-facing docs change included. This metric follows the existing eventing metrics conventions.


Sapthagiri777:
Hey @Leo6Leo @pierDipi  . This is my first PR here - pretty excited about it!
Just added the sink failures metric from #465. Would really appreciate it if someone could run /ok-to-test when you get a chance.

PR #4663, opened 2026-03-09
Fix rekt test race conditions after dependency update

creydr:
## Summary

- Upgrade to latest dependencies
- Fix race conditions in rekt tests by moving readiness checks to Requirement timing

## Details

After upgrading reconciler-test dependency to commit ef80c558 (which made eventshub forwarder installations block while waiting for KnativeService readiness), test race conditions became more frequent. The root cause: within each timing phase (Setup, Requirement, etc.), individual steps execute in parallel and compete for global parallel execution slots. When readiness checks start before installation steps complete, they poll for 120 seconds waiting for resources that don't exist yet.

The fix moves all `IsReady()`, `IsAddressable()`, and `ValidateAddress()` calls from Setup to Requirement timing, ensuring all installations complete before any verification begins.

### DIFF
--- a/test/upgrade-tests.sh
+++ b/test/upgrade-tests.sh
@@ -44,6 +44,7 @@ EVENTING_KAFKA_BROKER_UPGRADE_TESTS_FINISHEDSLEEP="5m" \
 go_test_e2e -v \
   -tags=upgrade \
   -timeout="${TIMEOUT}" \
+  -parallel=16 \
   ./test/upgrade ||
   fail_test
 


creydr:
```
Waiting until all pods in namespace cert-manager are up......................................................................................................................................................
```
/test channel-reconciler-tests-ssl

creydr:
`Pod scheduling timeout.`
/retest

creydr:
`Pod scheduling timeout.`
/retest

creydr:
tests run into a timeout
/retest-required

PR #4673, opened 2026-03-24
add METRICS_DISABLE_EVENT_TYPE_TAG option

### LINKED ISSUE #4672: Optionally disable event_type tag in metrics to limit potential cardinality explosion
**Problem**
In some use cases, the kafkasource may process events with generated event_types, or just have too many event_types. In those cases, the receiver and dispatcher will consume a lot of memory to keep the per event_type metrics, causing readiness/liveness failures and/or restarts due to OOM.

In such scenarios, it might be useful to optionally disable adding the event_type tag to the metrics, or just set it to a static value (e.g. "unknown"), to keep the existing dashboards that expect the tag there to keep working, just showing the event_type as "unknown".

System Operator

**Exit Criteria**
An env flag, or an option in observability config, disabling event-type metrics tag. 

**Time Estimate (optional):** 
1

maschmid:
Fixes #4672

## Proposed Changes

- :gift: Add METRICS_DISABLE_EVENT_TYPE_TAG option

**Release Note**

```release-note
Added METRICS_DISABLE_EVENT_TYPE_TAG option to dispatcher and receiver
```


creydr (approved):
/lgtm

PR #4703, opened 2026-04-22
fix(data-plane): fetch k8s secrets on separate thread from main vertx instance to prevent contention

Cali0707:
This should fix a bug where dispatcher pods sometimes fail to start by timing out while fetching a secret in the egress reconciliation. As far as I can tell, this is happening because we switched to using the default vert.x http client for these requests, so the execute blocking calls + all the fetch requests contend on the same thread.

## Proposed Changes

- Use a separate executor for the auth instead of the default vert.x thread

```release-note
fix(data-plane): resolve issue where dispatcher pods sometimes fail at startup while trying to fetch secrets
```


creydr (approved):
/lgtm

Thanks @Cali0707

PR #4720, opened 2026-05-14
fix(control-plane): kafkasink reconciler cluster admin error

chriscannon:
# Fixes
This fixes an issue where the ready state of a KafkaSink resource is not updated when a reconciliation fails. Specifically this error came about by updating a KafkaSink bootstrapServers field to invalid IPs. When Knative attempts to reconcile the KafkaSink it fails to do so and then requeues the reconciliation. What it should do though is change the ready status to false and then continue to requeue the reconciliation. 

## Proposed Changes
- Sets the KafkaSink object ready condition to false if the reconciler fails to obtain Kafka cluster admin.

### DIFF
--- a/control-plane/pkg/reconciler/base/receiver_condition_set.go
+++ b/control-plane/pkg/reconciler/base/receiver_condition_set.go
@@ -23,14 +23,12 @@ import (
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
+	sources "knative.dev/eventing-kafka-broker/control-plane/pkg/apis/sources/v1"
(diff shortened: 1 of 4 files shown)


creydr (approved):
Good catch. Thanks for fixing @chriscannon 

/lgtm

chriscannon:
@creydr Thanks for approving and merging! I noticed similar issues in consumergroup.go, channel.go, broker.go, and trigger.go. Would you like me to cut PRs for those as well?

creydr:
> @creydr Thanks for approving and merging! I noticed similar issues in consumergroup.go, channel.go, broker.go, and trigger.go. Would you like me to cut PRs for those as well?

That would be awesome if you could provide a fix for those too

PR #4724, opened 2026-05-21
(control-plane): updated broker, channel, consumer group, and trigger errors

chriscannon:
# Fixes #

Updates some of the errors returned in the broker, channel, consumer group, and trigger so that reconciliation properly fails when an error occurs. 

## Proposed Changes

- In channel.go replaces the FailedToCreateTopic error with the more appropriate FailedToGetKafkaClusterAdmin error.
- In broker.go replaces the FailedToResolveConfig with the more appropriate FailedToGetKafkaClusterAdmin error.
- In trigger.go also mark the dependency as failed along with the original error.
- In consumergroup.go mark the condition as false for each specific error.

### DIFF
--- a/control-plane/pkg/reconciler/broker/broker.go
+++ b/control-plane/pkg/reconciler/broker/broker.go
@@ -22,13 +22,6 @@ import (
 	"strings"
 	"time"
 
-	eventingv1alpha1 "knative.dev/eventing/pkg/apis/eventing/v1alpha1"
-
-	"k8s.io/utils/ptr"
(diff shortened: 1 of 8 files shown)


dprotaso:
/retest

different failures - @creydr flake?

creydr:
Thanks a lot @chriscannon for fixing this :muscle: 

/approve

creydr:
This seems like a flake (different test failures across the PR)
/test upgrade-tests

PR #4725, opened 2026-05-26
fix(test): fix flaky upgrade tests

creydr:
The upgrade tests are flaky across multiple PRs (e.g. #4724, #4716) with different tests failing each run.

## Root Cause

Two issues cause the flakes:

1. **Missing readiness checks**: After upgrade/downgrade, the verify phases immediately send events without waiting for resources to become ready again. The `kafka-source-dispatcher` pod restarts during upgrade and the KafkaSource CR temporarily goes NotReady (`BindInProgress`).

2. **Consumer subscription race**: Even after the KafkaSource CR reports Ready (dispatcher pod is running), the Kafka consumer inside the pod may not have joined the consumer group yet. Since no events are produced during pre-upgrade setup, there are no committed offsets. With `auto.offset.reset=latest`, the consumer starts at the end of the topic. The sender sends exactly 1 event — if it is produced before the consumer subscribes, the event is missed permanently and the test times out.

## Proposed Changes

- Add `kafkasink.IsReady()` and `kafkasource.IsReady()` requirements to `KafkaSinkSourceBinaryEventFeature` and `KafkaSinkSourceStructuredEventFeature` verify phases
- Add `broker.IsReady()` and `broker.IsAddressable()` requirements to `BrokerEventTransformationForTrigger` and `NamespacedBrokerEventTransformationForTrigger` verify phases
- Return the KafkaSource name from `KafkaSourceFeatureSetup` so callers can check its readiness
- Add `extraOpts` variadic parameter to `KafkaSourceFeatureAssert` for backward-compatible sender customization
- Send 10 events over ~27s in upgrade verify phases (via `SendMultipleEvents`) so at least some events arrive after the consumer subscribes

### DIFF
--- a/test/rekt/features/kafka_source.go
+++ b/test/rekt/features/kafka_source.go
@@ -348,7 +348,7 @@ func KafkaSourceFeatureSetup(f *feature.Feature,
 	return kafkaSinkCfg.sinkName, kafkaSourceCfg.sourceName, receiver
 }
 
-func KafkaSourceFeatureAssert(f *feature.Feature, kafkaSink, receiver string, customizeFunc CustomizeEventFunc) {
+func KafkaSourceFeatureAssert(f *feature.Feature, kafkaSink, receiver string, customizeFunc CustomizeEventFunc, extraOpts ...eventshub.EventsHubOption) {
 	sender := feature.MakeRandomK8sName("eventshub-sender")
 	options := []eventshub.EventsHubOption{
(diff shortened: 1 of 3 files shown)


creydr:
testing if it passes again:

/test upgrade-tests

creydr:
one more
/test upgrade-tests

creydr:
This is ready for review

/cc @Cali0707

Cali0707 (approved):
/lgtm
/approve

Thanks @creydr !

PR #4752, opened 2026-07-17
Fix OpenTelemetry BOM ordering to ensure declared version takes precedence

creydr:
## Summary

- The `quarkus-bom` was imported before the `opentelemetry-bom` in `data-plane/pom.xml` `<dependencyManagement>`. In Maven, when multiple BOMs manage the same artifact, the first declaration wins. This caused the Quarkus-managed OTel version (`1.44.x` via `opentelemetry-instrumentation-bom-alpha:2.10.0`) to silently override the explicitly declared `opentelemetry.version`, making the OTel version bump from #4633 / #4680 ineffective.
- Moves the `opentelemetry-bom` import to the top of `<dependencyManagement>` so the project's declared version (`1.60.1`) is actually used.

Fixes #4677

/cc @iusztin


Cali0707 (approved):
/lgtm
/approve

PR #4756, opened 2026-07-22
[release-1.21] Fix OpenTelemetry BOM ordering to ensure declared version takes precedence

creydr:
## Summary

- Cherry-pick of #4752 to release-1.21
- The `quarkus-bom` was imported before the `opentelemetry-bom` in `data-plane/pom.xml` `<dependencyManagement>`. In Maven, when multiple BOMs manage the same artifact, the first declaration wins. This caused the Quarkus-managed OTel version (`1.44.x`) to silently override the explicitly declared `opentelemetry.version`, making the OTel version bump from #4633 / #4680 ineffective.
- Moves the `opentelemetry-bom` import to the top of `<dependencyManagement>` so the project's declared version is actually used.

Fixes #4677

### DIFF
--- a/data-plane/pom.xml
+++ b/data-plane/pom.xml
@@ -116,6 +116,17 @@
 
   <dependencyManagement>
     <dependencies>
+      <!-- OpenTelemetry BOM must be imported before quarkus-bom to ensure
+           the explicitly declared opentelemetry.version takes precedence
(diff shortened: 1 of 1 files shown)


creydr:
/cc @Cali0707 

@iusztin fyi

PR #4762, opened 2026-07-29
fix: Fix timer drift in cache impl

dsimansk:
Fixes #

## Proposed Changes

- fix: Fix timer drift in cache impl

The background goroutine in cache.go:108-113 uses time.After(expiration) which creates a new timer each iteration, introducing cumulative drift. A time.NewTicker would fire at more consistent intervals and avoid the drift that causes entries to miss their expiration window under load. This results in unstable unit tests execution that fail intermittently. 

/cc @gauron99

### DIFF
--- a/control-plane/pkg/counter/counter.go
+++ b/control-plane/pkg/counter/counter.go
@@ -18,12 +18,14 @@ package counter
 
 import (
 	"context"
+	"sync"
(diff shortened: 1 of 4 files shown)


gauron99 (approved):
/lgtm

PR #4778, opened 2026-08-02
fix: use configured system namespace for KafkaChannel subscriber SA check

### LINKED ISSUE #4742: KafkaChannel subscriber validation hardcodes knative-eventing
Describe the bug

When Knative Eventing is installed in a custom system namespace (via Knative Operator), Subscription resources that target a KafkaChannel fail with PhysicalChannelSyncFailed.

The Kafka validating webhook (validation.webhook.kafka.eventing.knative.dev) rejects updates to KafkaChannel.spec.subscribers because it only allows changes from:

system:serviceaccount:knative-eventing:eventing-controller

The actual subscription-controller runs as:

system:serviceaccount:<custom-system-namespace>:eventing-controller

Example error:

validation failed: Channel.Spec.Subscribers changed by user system:serviceaccount:ecpk-core-eventing:eventing-controller which was not the system:serviceaccount:knative-eventing:eventing-controller service account: spec.subscribers
Root cause: in control-plane/pkg/apis/messaging/v1/kafka_channel_validation.go the allowed SA is hardcoded:

const eventingControllerSAName = "system:serviceaccount:knative-eventing:eventing-controller"
Setting SYSTEM_NAMESPACE=<custom-namespace> on the kafka-webhook-eventing deployment does not fix the issue, because validation does not use system.Namespace().

Expected behavior

When Eventing control-plane runs in a custom system namespace (e.g. ecpk-core-eventing), subscription-controller should be allowed to update KafkaChannel.spec.subscribers.

The Subscription should become Ready=True, and the channel should contain the subscriber in spec.subscribers.

Validation should use the actual system namespace (e.g. via system.Namespace()), similar to core Knative Eventing InMemoryChannel validation in knative.dev/eventing/pkg/apis/messaging/v1/in_memory_channel_validation.go.

To Reproduce

Install Knative Operator (v1.22.x) and Knative Eventing into a custom namespace (not knative-eventing), e.g. ecpk-core-eventing.
Install eventing-kafka-broker via operator additionalManifests (e.g. knative-v1.22.0/eventing-kafka.yaml).
Create a KafkaChannel in a user namespace.
Create a Subscription referencing the channel and a subscriber.
apiVersion: messaging.knative.dev/v1
kind: Subscription
metadata:
  name: example-sub
  namespace: user-namespace
spec:
  channel:
    apiVersion: messaging.knative.dev/v1
    kind: KafkaChannel
    name: my-channel
  subscriber:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: my-service
Check status:
kubectl describe subscription example-sub -n user-namespace
kubectl get kafkachannel my-channel -n user-namespace -o yaml
Result: Subscription is Ready=False (PhysicalChannelSyncFailed), KafkaChannel.spec.subscribers stays empty, webhook denies the update with the SA mismatch above.

Note: Kafka Broker, Trigger, KafkaSource, KafkaSink work in the same setup. Only Subscription → KafkaChannel sync fails.

Knative release version

Knative Operator: v1.22.2
Knative Eventing: v1.22.0
eventing-kafka-broker: v1.22.0 (eventing-kafka.yaml)
Kubernetes: Client Version: v1.35.5
Kustomize Version: v5.7.1
Server Version: v1.35.0+k3s1
Additional context

Suggested fix — replace hardcoded constant in kafka_channel_validation.go:

var eventingControllerSAName = fmt.Sprintf(
    "system:serviceaccount:%s:eventing-controller",
    system.Namespace(),
)
Workaround: use Broker → Trigger → subscriber instead of KafkaChannel → Subscription → subscriber.

Related:

[Subscriptions docs](https://knative.dev/docs/eventing/channels/subscriptions/)
Webhook manifest sets SYSTEM_NAMESPACE from metadata.namespace (500-webhook.yaml), but validation code ignores it.
Proposed priority: medium — breaks documented Subscription flow for any non-default system namespace install; workaround exists via Trigger.

pujitha24:
Motivation:

KafkaChannel validation only allowed spec.subscribers to be changed by
system:serviceaccount:knative-eventing:eventing-controller, with the
namespace hardcoded. When Knative Eventing is installed into a custom
system namespace, the real subscription-controller runs as
system:serviceaccount:<custom-namespace>:eventing-controller, so the
webhook rejects its updates and Subscriptions targeting a KafkaChannel
never become Ready (PhysicalChannelSyncFailed), leaving
spec.subscribers empty. Setting SYSTEM_NAMESPACE on the webhook
deployment had no effect because the validation code never read it.

Approach:

Replace the hardcoded eventingControllerSAName constant in
control-plane/pkg/apis/messaging/v1/kafka_channel_validation.go and
control-plane/pkg/apis/messaging/v1beta1/kafka_channel_validation.go
with a package-level var built from knative.dev/pkg/system.Namespace(),
which resolves SYSTEM_NAMESPACE at process start. This mirrors the
existing pattern already used by knative/eventing's own
InMemoryChannel validation
(vendor/knative.dev/eventing/pkg/apis/messaging/v1/in_memory_channel_validation.go)
for the same problem. Both the webhook's storage (v1beta1) and served
(v1) KafkaChannel types are updated so the check is consistent across
API versions.

This only changes which service account is treated as the trusted
subscription-controller; behavior for the default "knative-eventing"
namespace is unchanged.

Validation:

Ran `go build ./control-plane/...` (passes) and
`go test ./control-plane/pkg/apis/messaging/...` (passes), which is
the unit-test path documented in DEVELOPMENT.md
(./hack/run.sh unit-tests-control-plane) and does not require a
running cluster. Added kafka_channel_validation_test.go to both
packages exercising CheckSubscribersChangeAllowed against the
configured system-namespace service account, a same-named service
account in a different namespace, and an arbitrary user. Confirmed
these tests reproduce the reported failure: reverting only the two
validation.go changes and rerunning the tests fails with the exact
error from the issue ("... which was not the
system:serviceaccount:knative-eventing:eventing-controller service
account"), and reapplying the fix makes them pass.

Report: https://github.com/knative-extensions/eventing-kafka-broker/issues/4742
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #4742

```release-note
fix: use configured system namespace for KafkaChannel subscriber SA check
```

### DIFF
--- a/control-plane/pkg/apis/messaging/v1/kafka_channel_validation.go
+++ b/control-plane/pkg/apis/messaging/v1/kafka_channel_validation.go
@@ -26,9 +26,10 @@ import (
 	"knative.dev/eventing/pkg/apis/eventing"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/kmp"
+	"knative.dev/pkg/system"
 )
 
-const eventingControllerSAName = "system:serviceaccount:knative-eventing:eventing-controller"
+var eventingControllerSAName = fmt.Sprintf("system:serviceaccount:%s:eventing-controller", system.Namespace())
 
 func (kc *KafkaChannel) Validate(ctx context.Context) *apis.FieldError {
 	errs := kc.Spec.Validate(ctx).ViaField("spec")
--- a/control-plane/pkg/apis/messaging/v1/kafka_channel_validation_test.go
+++ b/control-plane/pkg/apis/messaging/v1/kafka_channel_validation_test.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2026 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	authenticationv1 "k8s.io/api/authentication/v1"
+
+	eventingduck "knative.dev/eventing/pkg/apis/duck/v1"
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/system"
+	_ "knative.dev/pkg/system/testing"
+)
+
+func TestCheckSubscribersChangeAllowed(t *testing.T) {
+	subscriberURI := apis.HTTP("example.com")
+
+	original := &KafkaChannel{
+		Spec: KafkaChannelSpec{},
+	}
+	updated := &KafkaChannel{
+		Spec: KafkaChannelSpec{
+			ChannelableSpec: eventingduck.ChannelableSpec{
+				SubscribableSpec: eventingduck.SubscribableSpec{
+					Subscribers: []eventingduck.SubscriberSpec{{SubscriberURI: subscriberURI}},
+				},
+			},
+		},
+	}
+
+	tests := []struct {
+		name     string
+		username string
+		wantErr  bool
+	}{{
+		name:     "allowed for the eventing-controller in the configured system namespace",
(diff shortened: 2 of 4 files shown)


pujitha24:
/assign @dsimansk

All checks are green and it's been open a little while — happy to make any changes needed if you get a chance to look.

creydr (approved):
Thanks a lot for the fix @pujitha24 

/lgtm

